PR #125, opened 2014-04-09
[WIP] Make RCP always non-null

certik:
This is just to see how fast things get. The Cython interface can't currently work with non-null RCP pointers.

When running benchmarks, I don't see any significant difference in speed.

This is the same as #74 but using my local branch.


certik:
I created an issue #378 for this and I am closing this PR.

PR #361, opened 2015-01-10
Added the method `symbols`

shivamvats:
The method `symbols`lets us declare multiple Symbol
objects in one go


certik:
Thanks for the PR! Couple comments:

* `var` should also be refactored to use the `symbols`, so that functionality is not duplicated.

* Why not to use the SymPy's implementations from https://github.com/sympy/sympy/blob/bcca216ff440202415e27ac37ff3b9ff6bc2c590/sympy/core/symbol.py#L317? There are more features there and I think it should just work.

* Since this is pure Python, it should go into a Python module, not the Cython module.

shivamvats:
@certik 
Do you want me to copy paste Sympy's `symbols`? 

Should I put it in utilities.py?
Btw, will declaring the function as cdef do any good?

certik:
Yes, you can put it to `utilities.py`. I don't think declaring it as `cdef` does any good. 

I would copy and paste it, the only issue might be that SymPy allows you to set old style assumptions to symbols, so you need to remove that part.

shivamvats:
For string_types, we'll need to know the version of python on the system. This is checked in core.compatibility in Sympy. Should I put the check in utilities.py or should there  be a separate file for it?
The code will be roughly:
```
import sys
PY3 =  sys.version_info[0] > 2
if PY3
    string_types = (str, )
else
    string_types = (str, unicode)
```

certik:
Good point. Just copy the compatibility file from SymPy. I think it is self contained and can be reused.

csympy/lib/csympy_wrapper.pyx
```diff
@@ -249,6 +249,31 @@ cdef class Symbol(Basic):
         import sympy
         return sympy.Symbol(str(deref(X).get_name().decode("utf-8")))
 
+
+def symbols(names, *args):
```
thilinarmtb: Normally, one would call `symbols` using named arguments. For example:
```python
x, y, z = symbols('x,y,z', real=True)
```
We are not supporting this at the moment but I think it will be better to use `**args` instead of `*args`.

shivamvats: I did so as we are not using assumptions currently. But on second thought, I agree with you.

shivamvats:
To be able to define a function using `symbols`, we'll need an undefined_function method, which we don't have. I can make `symbols` return a FunctionSymbol, but that's all. Should I add an undefined_function method to this pr or do that later (or is there another way)?

certik:
I would disable the functions for now. We can implement that later.

shivamvats:
@certik Here is a summary of the recent commits:
1) I have copied compatibility.py from Sympy. Besides the method that we currently need(that I've edited for our use), it contains a lot of other methods too(that I haven't edited for our use). Should we keep them for later use or remove them?
2) I've also copied importtools.py from Sympy as compatibility.py has a dependency on it. The Sympy version also contains a sympy_debug function that I've removed.

csympy/compatibility.py
```diff
@@ -0,0 +1,925 @@
...
+    from warnings import warn
+    warn("gmpy library is not installed, switching to 'python' ground types")
+    GROUND_TYPES = 'python'
+
+# SYMPY_INTS is a tuple containing the base types for valid integer types.
+SYMPY_INTS = integer_types
+
+if GROUND_TYPES == 'gmpy':
+    SYMPY_INTS += (type(gmpy.mpz(0)),)
+
```
certik: You can remove everything up to here.

csympy/compatibility.py
```diff
@@ -0,0 +1,925 @@
...
+                from sympy.utilities.iterables import uniq
+                u = list(uniq(d[k]))
+                if len(u) > 1:
+                    raise ValueError(
+                        'not enough keys to break ties: %s' % u)
+        for v in d[k]:
+            yield v
+        d.pop(k)
+
+# If HAS_GMPY is 0, no supported version of gmpy is available. Otherwise,
```
certik: You can remove everything between this line and my comment below.

csympy/importtools.py
```diff
@@ -0,0 +1,156 @@
+"""Tools to assist importing optional external modules."""
```
certik: This file becomes obsolete. I would remove it.

certik:
Otherwise this looks good. Then I would also copy `var` from sympy, which is a simple function that just calls `symbols`.

certik:
Besides what I indicated in the comments, I would leave the rest in `compatibility.py`. It can come handy later.

shivamvats:
Sympy has a `raises` method for tests. Do we need it?

certik:
We might need it. It's useful to test exceptions.

shivamvats:
I'll copy `var` from Sympy in a separate pr. This one has already grown too big.

certik:
That's fine too. The advantage of doing it in this PR is that it would become clear if there is any incompatibility with regards to the new `symbol` function, as we have quite a few tests for `var` already.

pbrady:
I wouldn't copy sympy's pytest.py file.  Running the python tests already depends on pytest: https://github.com/sympy/csympy/blob/master/csympy/__init__.py#L6 which makes the customized pytest.py file obsolete.  I don't think sympy's pytest.py file makes sense unless the entire test runner infrastructure is also copied.  This does introduce a dependency on pytest (pure python package) but it will save thousands of lines of code

certik:
Ah, I didn't realize what you were asking. Yes, we should not copy any test infrastructure, just use py.test or nosetests, as @pbrady said.

shivamvats:
@pbrady @certik Thanks! I didn't see that.

shivamvats:
@certik @pbrady Could you have a look?

certik:
This looks good. The last issue that I have is that I think now it won't work with nosetests, only with pytests. 

I don't know if we should support both, or just stick with py.test. I would support both as long as possible, so we should figure out how to use/import `raises` so that we can use both.

shivamvats:
How about using `assertRaises` from unittest? I think it is supported by both nosetests and py.test

certik:
That works, if it is implemented in both.

shivamvats:
@certik  This works with both nosetests and py.test

certik:
I checked that tests run in both nosetests and py.test, but if you uninstall py.test, then tests fail:
```
$ nosetests
..........................................................................................................................E
======================================================================
ERROR: Failure: ImportError (No module named pytest)
----------------------------------------------------------------------
  File "/local/certik/bld/profile/xd2emat6tjxu/lib/python2.7/site-packages/nose/loader.py", line 414, in loadTestsFromName
    addr.filename, addr.module)
  File "/local/certik/bld/profile/xd2emat6tjxu/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/local/certik/bld/profile/xd2emat6tjxu/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/certik/repos/csympy/csympy/tests/test_var.py", line 4, in <module>
    from pytest import raises
ImportError: No module named pytest

----------------------------------------------------------------------
Ran 123 tests in 0.584s

FAILED (errors=1)
```

Rather than introducing unittests, I think we should keep the tests as we had them, but just introduce raises, that internally would do the right thing. I can help send the necessary patch in the afternoon, I am a bit busy now.

shivamvats:
Oh I'm sorry. I didnt modify 
test_var.py. I'll send a patch to use unittests in test_var.py too. It should work without py.test after that.

shivamvats:
It should work without py.test now.

certik:
That looks good. I still think we should not use unittests, but just the test functions as before, and introduce a "raises" function that internally does the right thing and will work with both nosetests and py.test.

shivamvats:
Alright. Should we reuse Sympy's raises? If not, you can send a patch.

certik:
Yes, just use this code from sympy:
```python
try:
    import py
    from py.test import skip, raises
    USE_PYTEST = getattr(sys, '_running_pytest', False)
except ImportError:
    USE_PYTEST = False

ON_TRAVIS = os.getenv('TRAVIS_BUILD_NUMBER', None)

if not USE_PYTEST:
    def raises(expectedException, code=None):
        """
        Tests that ``code`` raises the exception ``expectedException``.

        ``code`` may be a callable, such as a lambda expression or function
        name.

        If ``code`` is not given or None, ``raises`` will return a context
        manager for use in ``with`` statements; the code to execute then
        comes from the scope of the ``with``.

        ``raises()`` does nothing if the callable raises the expected exception,
        otherwise it raises an AssertionError.

        Examples
        ========

        >>> from sympy.utilities.pytest import raises

        >>> raises(ZeroDivisionError, lambda: 1/0)
        >>> raises(ZeroDivisionError, lambda: 1/2)
        ...
        AssertionError: DID NOT RAISE

        >>> with raises(ZeroDivisionError):
        ...     n = 1/0
        >>> with raises(ZeroDivisionError):
        ...     n = 1/2
        ...
        AssertionError: DID NOT RAISE

        Note that you cannot test multiple statements via
        ``with raises``:

        >>> with raises(ZeroDivisionError):
        ...     n = 1/0    # will execute and raise, aborting the ``with``
        ...     n = 9999/0 # never executed

        This is just what ``with`` is supposed to do: abort the
        contained statement sequence at the first exception and let
        the context manager deal with the exception.

        To test multiple statements, you'll need a separate ``with``
        for each:

        >>> with raises(ZeroDivisionError):
        ...     n = 1/0    # will execute and raise
        >>> with raises(ZeroDivisionError):
        ...     n = 9999/0 # will also execute and raise

        """
        if code is None:
            return RaisesContext(expectedException)
        elif callable(code):
            try:
                code()
            except expectedException:
                return
            raise AssertionError("DID NOT RAISE")
        elif isinstance(code, str):
            raise TypeError(
                '\'raises(xxx, "code")\' has been phased out; '
                'change \'raises(xxx, "expression")\' '
                'to \'raises(xxx, lambda: expression)\', '
                '\'raises(xxx, "statement")\' '
                'to \'with raises(xxx): statement\'')
        else:
            raise TypeError(
                'raises() expects a callable for the 2nd argument.')

    class RaisesContext(object):
        def __init__(self, expectedException):
            self.expectedException = expectedException

        def __enter__(self):
            return None

        def __exit__(self, exc_type, exc_value, traceback):
            if exc_type is None:
                raise AssertionError("DID NOT RAISE")
            return issubclass(exc_type, self.expectedException)
```
My apologies, I should have been more clear above.

shivamvats:
Should I put it in pytest.py? I hope @pbrady has no objections.

certik:
I think that's fine. Just put there the `raises`.

shivamvats:
Should I rebase now?

certik:
This looks good. I talked to @pbrady, he's fine with this too. I don't think you need to rebase. Let's wait until tests pass and we can merge this.

shivamvats:
Nice to know that.
By rebasing I meant sqashing unnecessary commits. I think that will make the commit history more readable.

certik:
Ah yes, if you could create a nice history with commits that make sense for the current PR, that would be awesome.

certik:
Nice patches, I like it, +1. The only issue is that the tests are failing now, but I think that's unrelated to this PR. We need to look into it.

certik:
I see what the problem is. The `pytest` module that you added clashes with the `pytest` library. To fix this, just put the contents of the `pytest.py` into `utilities.py` and update the imports. That will do it.

shivamvats:
I did so, but I'm getting assertion errors in test_var.py:
```
  File "/home/aries/Software/anaconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/aries/csympy/csympy/tests/test_var.py", line 46, in test_var
    raises(NameError, lambda: z2)
  File "/home/aries/csympy/csympy/pytests.py", line 75, in raises
    raise AssertionError("DID NOT RAISE")
AssertionError: DID NOT RAISE
```
This happens only with nosetests. The error however goes away on removing the test function in __init__.py

certik:
What does the test function in `__init__.py` have to do with this?

certik:
Just put some print statements in the `raises` function to figure out the workflow and use `nosetests -s` to see the prints.

certik:
If https://github.com/shivamvats/csympy/pull/1 is merged, then this PR is +1 from me to merge. It should work in all cases (nosetests, py.test, Travis, ...).

certik:
Tests pass, so I am merging this. Thanks @shivamvats!

PR #373, opened 2015-01-19
Allows expand to work with pow instances with negative powers

### LINKED ISSUE #371: (1/(x+y)).expand() should do nothing
Currently it raises this unhelpful exception:
```
In [6]: (1/(x+y)).expand()
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-6-5a1624685c2a> in <module>()
----> 1 (1/(x+y)).expand()

/home/ondrej/repos/csympy/csympy/lib/csympy_wrapper.so in csympy.lib.csympy_wrapper.Basic.expand (/home/ondrej/repos/csympy/csympy/lib/csympy_wrapper.cpp:5842)()

RuntimeError: multinomial_coefficients: n >= 0 must hold.
```

shivamvats:
Added check for negative exponent and also proper handling of such
cases.

Fixes #371

### DIFF
--- a/src/pow.cpp
+++ b/src/pow.cpp
@@ -350,15 +350,15 @@ void multinomial_coefficients_mpz(int m, int n, map_vec_mpz &r)
 RCP<const Basic> pow_expand(const RCP<const Pow> &self)
 {
     RCP<const Basic> _base = expand(self->base_);
-    int negative_pow = 0;
+    bool negative_pow = false;
     if (! is_a<Integer>(*self->exp_) || ! is_a<Add>(*_base))
         return self;
 
     map_vec_mpz r;
     int n = rcp_static_cast<const Integer>(self->exp_)->as_int();
-    if(n < 0){
+    if (n < 0) {
         n = -n;
-        negative_pow++;
+        negative_pow = true;
     }
     RCP<const Add> base = rcp_static_cast<const Add>(_base);
     umap_basic_num base_dict = base->dict_;
@@ -434,8 +434,7 @@ RCP<const Basic> pow_expand(const RCP<const Pow> &self)
         }
     }
     RCP<const Basic> result = Add::from_dict(add_overall_coeff, std::move(rd));
-    RCP<const Number> i = integer(-1);
-    if(negative_pow) result = rcp(new Pow(result, i));
+    if (negative_pow) result = pow(result, minus_one);
     return result;
 }
 
--- a/src/pow.cpp
+++ b/src/pow.cpp
@@ -350,12 +350,16 @@ void multinomial_coefficients_mpz(int m, int n, map_vec_mpz &r)
 RCP<const Basic> pow_expand(const RCP<const Pow> &self)
 {
     RCP<const Basic> _base = expand(self->base_);
+    int negative_pow = 0;
     if (! is_a<Integer>(*self->exp_) || ! is_a<Add>(*_base))
         return self;
 
     map_vec_mpz r;
     int n = rcp_static_cast<const Integer>(self->exp_)->as_int();
-
+    if(n < 0){
+        n = -n;
+        negative_pow++;
+    }
     RCP<const Add> base = rcp_static_cast<const Add>(_base);
     umap_basic_num base_dict = base->dict_;
     if (! (base->coef_->is_zero())) {
@@ -430,6 +434,8 @@ RCP<const Basic> pow_expand(const RCP<const Pow> &self)
         }
     }
     RCP<const Basic> result = Add::from_dict(add_overall_coeff, std::move(rd));
+    RCP<const Number> i = integer(-1);
+    if(negative_pow) result = rcp(new Pow(result, i));
     return result;
 }
 
--- a/src/tests/basic/test_arit.cpp
+++ b/src/tests/basic/test_arit.cpp
@@ -660,7 +660,7 @@ void test_expand2()
     r2 = expand(r1);
     std::cout << *r2 << std::endl;
 
-    assert(eq(r2, r1);
+    assert(eq(r2, r1));
 
     r1 = pow(add(x, y), im2);       // 1/(x+y)^2
     std::cout << *r1 << std::endl;
@@ -668,7 +668,7 @@ void test_expand2()
     r2 = expand(r1);
     std::cout << *r2 <<std::endl;
 
-    assert(eq(r2, pow(add(pow(x, i2), mul(i2, x, y), pow(y, i2)), im1);
+    assert(eq(r2, pow(add(add(pow(x, i2), mul(mul(i2, x), y)), pow(y, i2)), im1)));
     assert(neq(r2, r1));
 
     r1 = mul(im1, add(x, i2));
--- a/src/tests/basic/test_arit.cpp
+++ b/src/tests/basic/test_arit.cpp
@@ -642,7 +642,7 @@ void test_expand2()
     r2 = expand(r1);
     std::cout << *r2 << std::endl;
 
-    assert( eq(r2, add(add(mul(w, x), mul(w, y)), mul(w, z))));
+    assert(eq(r2, add(add(mul(w, x), mul(w, y)), mul(w, z))));
     assert(neq(r2, add(add(mul(w, x), mul(w, w)), mul(w, z))));
 
     r1 = mul(add(x, y), add(z, w)); // (x+y)*(z+w)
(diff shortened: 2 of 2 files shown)


src/pow.cpp
```diff
@@ -355,12 +355,16 @@ void multinomial_coefficients_mpz(int m, int n, map_vec_mpz &r)
 RCP<const Basic> pow_expand(const RCP<const Pow> &self)
 {
     RCP<const Basic> _base = expand(self->base_);
+    int negative_pow = 0;
```
certik: I think this should be a `bool`.

src/pow.cpp
```diff
@@ -428,6 +432,8 @@ RCP<const Basic> pow_expand(const RCP<const Pow> &self)
         }
     }
     RCP<const Basic> result = Add::from_dict(add_overall_coeff, std::move(rd));
+    RCP<const Number> i = integer(-1);
+    if(negative_pow) result = rcp(new Pow(result, i));
```
certik: I think this could cause some assert errors. Consider a case like 1/x^2, so `result = x^2` and then you would get `(x^2)^(-1)`, which will trigger an assert error, as it should rather be `x^(-2)`. You can achieve that by calling `pow(result, i)`.

src/pow.cpp
```diff
@@ -428,6 +432,8 @@ RCP<const Basic> pow_expand(const RCP<const Pow> &self)
         }
     }
     RCP<const Basic> result = Add::from_dict(add_overall_coeff, std::move(rd));
+    RCP<const Number> i = integer(-1);
```
certik: You can use `minus_one` from `constants.h`.

certik:
Thanks for the patch! I left some comments, but I think the overall idea is correct.

src/pow.cpp
shivamvats: This case won't throw an error as the base isn't an instance of Add. expand() will immediately return self in this case. However, I understand what you mean. I'll make the change.

certik: I see, that's right. The other option is to put an assert statement above the `pow` to check that the `result` is `Add`. Then I think it's safe and faster.

certik: Thinking about this more, I don't think it can trigger assert errors, because the result of expand() is always `Add` I think. What about a case like this one: `1/((x+y)^2-2*x*y-y^2)`?

src/pow.cpp
```diff
@@ -355,12 +355,16 @@ void multinomial_coefficients_mpz(int m, int n, map_vec_mpz &r)
 RCP<const Basic> pow_expand(const RCP<const Pow> &self)
 {
     RCP<const Basic> _base = expand(self->base_);
+    bool negative_pow = false;
     if (! is_a<Integer>(*self->exp_) || ! is_a<Add>(*_base))
         return self;
```
certik: These two lines ensure that `_base` is always `Add`, so I think we can speed it up by using your original approach with `rcp(new Pow(...))`. And put an assert statement above it ---- that way we make sure the expression is what we think it is, and if so, pass it directly to Pow.

src/pow.cpp
shivamvats: It doesn't simplify the base:
```
>>> (1/((x+y)**2 - 2*x*y - x**2)).expand()
x + y(-x^2 - 2*x*y + (y + x)^2)^(-1)
```

shivamvats: I'm getting a weird error on trying this
```
>>> (1/((x + y)**2 - x)).expand()
```
https://gist.github.com/shivamvats/87f24b310da644a43bab

certik: SymPy does;
```
In [1]: (1/((x+y)**2 - 2*x*y - x**2)).expand()
Out[1]: 
1 
──
 2
y 
```

certik: That error is an assert error, the dictionary that was passed to Mul was not in canonical form. It needs to be fixed.

src/pow.cpp
isuruf: Last two lines need something like this,
```
if (! is_a<Integer>(*self->exp_) || ! is_a<Add>(*_base)) {
        if (neq(_base, self->base_)) {
            return pow(_base, self->exp_);
        } else {
            return self;
        }
}
```

src/pow.cpp
isuruf: I'm not getting an assertion error with the latest master. Can you rebase with master and check?

src/pow.cpp
isuruf: Although `_base` is a `Add`, that does not guarantee that `result` is not a `Mul` or a `Number` ...
`(1/(sqrt(3)*sqrt(2) + sqrt(6)*I)**4).expand()`

certik: Nice one. I was wondering if there is a counterexample.

isuruf:
@shivamvats Your commits are not getting linked to your profile. Have you verified your email on Github?

shivamvats:
I didnt notice I had a different email-id in my git-config from my primary github id. I've fixed it now.

Thanks @isuruf  for pointing out :)

src/pow.cpp
certik: @isuruf we should also add a test, that fails unless these lines are there:
```c++
if (! is_a<Integer>(*self->exp_) || ! is_a<Add>(*_base)) {
        if (neq(_base, self->base_)) {
            return pow(_base, self->exp_);
        } else {
            return self;
        }
}
```

shivamvats: I get a similar error as I was getting earlier:
https://gist.github.com/shivamvats/87f24b310da644a43bab

Is it because of my commits? I'm unable to test the counter example.

certik: You can see the stacktrace there, showing where the error is. 

Can you post code that you ran to trigger this? We should make sure it doesn't happen in master.

shivamvats: The problem seems to be that the Mul and Pow expressions are not being canonicalised properly. So the is_canonical assertions are failing. I'm getting similar errors in a lot of expressions, like `(1/((x+y)**2 + x)).expand()` , `(1/(sqrt(3)*sqrt(2) + sqrt(6)*I)**4).expand()`, `(1/((x+y)**2 - x**2 - y**2)).expand()`
I'll look into it further after my classes.

isuruf: You should rebase against master. Those errors are fixed in the latest master.

shivamvats: Oh I didn't know! Thanks.

certik:
This looks good to me as of 66da4d463b1624658d7dcdc27ef59bae98d2e9fc. @isuruf did you find a problem with the current PR? You mentioned that these two lines are needed:
```c++
if (! is_a<Integer>(*self->exp_) || ! is_a<Add>(*_base)) {
        if (neq(_base, self->base_)) {
            return pow(_base, self->exp_);
        } else {
            return self;
        }
}
```
If so, we first need to add a test that requires them.

isuruf:
Check out `(1/((x+1)**2-2*x-x**2)).expand()`. The expression stays as it is. This is unrelated to the PR, but we should add a test.

src/pow.cpp
```diff
@@ -350,12 +350,16 @@ void multinomial_coefficients_mpz(int m, int n, map_vec_mpz &r)
...
 
     map_vec_mpz r;
     int n = rcp_static_cast<const Integer>(self->exp_)->as_int();
-
+    if(n < 0){
```
isuruf: space after `if` and before `{`

src/pow.cpp
```diff
@@ -430,6 +434,7 @@ RCP<const Basic> pow_expand(const RCP<const Pow> &self)
         }
     }
     RCP<const Basic> result = Add::from_dict(add_overall_coeff, std::move(rd));
+    if(negative_pow) result = pow(result, minus_one);
```
isuruf: space after `if`

isuruf:
+1 to merge after the spaces are added.

certik:
Does it expand if you do a positive power? If not, then it's indeed
unrelated.

isuruf:
It doesn't expand for positive powers either.

certik:
Cool. We'll fix it in another PR.

certik:
I agree with @isuruf. Just fix the little things and lets merge it.

certik:
+1 to merge as it is if tests pass.

certik:
Tests pass, merging.

shivamvats:
Thanks!

PR #404, opened 2015-02-24
printing: printing of pow changed to match SymPy

### LINKED ISSUE #403: Power should be printed as **, not ^
I think the goal is to get the same results as sympy, only faster. In that case:

```
>>> import sympy
>>> sympy.var("x")
>>> 2**x
2**x
```

Whereas:

```
>>> import csympy
>>> csympy.var("x")
>>> 2**x
2^x
```

See https://github.com/sympy/csympy/blob/master/src/pow.cpp#L108

Sumith1896:
Before:
```
In[1]: 2**x
Out[1]: 2^x
```
After:
```
In[1]: 2**x
Out[1]: 2**x
```
The previous tests were also updated. Fixes #403

### DIFF
--- a/src/pow.cpp
+++ b/src/pow.cpp
@@ -105,7 +105,7 @@ std::string Pow::__str__() const
     }  else {
         o << *base_;
     }
-    o << "^";
+    o << "**";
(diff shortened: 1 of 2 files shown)


certik:
This looks good to me. Thanks!

PR #406, opened 2015-03-03
[WIP] Implementation of the basic structure of Polynomial class

shivamvats:


shivamvats:
@certik

src/polynomial.h
```diff
@@ -0,0 +1,67 @@
...
+#include "integer.h"
+
+namespace CSymPy {
+
+    class Polynomial : public Basic{
+    public:
+        int degree;
+        std::string var_;
+        //umap_vec_mpz dict;
+        map_integer_uint dict_;
```
certik: I assume this is packing all the exponents into `uint`?

shivamvats: Currently it is the other way round (exponents in `integer`). I used `map_integer_uint` just to test the code. I'll update it to use `uint` for exponent and `rational` (or `integer`) for coefficients.

Ultimately, however coefficients should depend on the base ring it is being defined on.

shivamvats:
@certik I see that `Matrix` avoids the  use  of `RCP`. Is there a reason for doing so? Do we want to use `RCP` for polynomials?

src/polynomial.cpp
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    RCP<const Polynomial> add_poly(const RCP<const Polynomial> &a, const RCP<const Polynomial> &b) {
+        map_uint_integer dict;
+        for(auto &it : a->dict_)
+            dict[it.first] = integer(0);
+        for(auto &it : b->dict_)
+            dict[it.first] = integer(0);
+        for(auto &it : a->dict_)
+            dict[it.first] = rcp_static_cast<const Integer>(add(dict[it.first],(it.second)));
```
certik: You should always check that the input is `Integer` before using `rcp_static_cast`.

certik:
Good point about `RCP`. If Polynomial is to be used in sympy expressions, then it needs to be able to work with `RCP`, which it will if it is subclassed from `Basic`. The function `add_poly` should not use `RCP` --- the rule of thumb is that you should just use a regular reference, not `RCP`, unless you have to. You would only use `RCP` if you are allocating memory. So you would need to use it for the polynomial that you return, but you don't need to use it for the input arguments.

The Travis error:
```
/home/travis/build/sympy/csympy/src/eval_arb.cpp: In function ‘void CSymPy::eval_arb(arb_struct*, const CSymPy::Basic&, long int)’:
/home/travis/build/sympy/csympy/src/eval_arb.cpp:276:20: error: cannot declare variable ‘v’ to be of abstract type ‘CSymPy::EvalArbVisitor’
/home/travis/build/sympy/csympy/src/eval_arb.cpp:18:7: note: because the following virtual functions are pure within ‘CSymPy::EvalArbVisitor’:
/home/travis/build/sympy/csympy/src/visitor.h:17:18: note: virtual void CSymPy::Visitor::visit(const CSymPy::Polynomial&)
```
suggests you also need to update the Arb visitor with Polynomial.



More generally, your implementation is going to be very slow, because it's using the `Integer` class with RCP and so on. I think the Polynomial class in csympy (that you started in this PR) should be a wrapper around a very fast lower level implementation.

The low level implementation should be class that does not subclass Basic, and it is just a tiny wrapper around the hashtable. The hashtable should accept mpz directly (no overhead of `RCP<Integer>`). More importantly, as we talked about, it should also be able to eventually accept our own fast integer implementation that switches between machine integer and mpz.

Sumith1896:
I have taken @shivamvats  commits from here to https://github.com/sympy/symengine/pull/454. This PR is now redundant.
Let me know if there is any other information that I need to carry over before this can be closed.

aktech:
@Sumith1896 

BTW, Another way could have been sending a Pull Request to @shivamvats  in his `poly_initial` branch.
In that case this PR would have sustained & no need for the new PR itself.

Sumith1896:
@aktech I had tried what you said here https://github.com/shivamvats/csympy/pull/2 but it didn't seem clean as we needed it to update with master. I don't know if I did it wrong it https://github.com/shivamvats/csympy/pull/2 but sending in a new PR seemed easiest and cleanest way forward.

aktech:
I am not sure what went wrong. But this should have worked:
```
1. (On shivam's branch)
git fetch
git rebase origin/master

2.
Then Add your commits.
```
(Though, you may have done it right, but right now I don't have time to
figure out what exactly happened.)

On Sun, May 24, 2015 at 8:31 PM, Sumith <notifications@github.com> wrote:

> @aktech <https://github.com/aktech> I had tried what you said here
> shivamvats#2 <https://github.com/shivamvats/csympy/pull/2> but it didn't
> seem clean as we needed it to update with master. I don't know if I did it
> wrong it shivamvats#2 <https://github.com/shivamvats/csympy/pull/2> but
> sending in a new PR seemed easiest and cleanest way forward.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sympy/symengine/pull/406#issuecomment-105022854>.
>



-- 



*Amit Kumar3rd Year, Mathematics & Computing EngineeringDelhi Technological
University (Formerly DCE)Website : www.iamit.in <http://www.iamit.in>*

shivamvats:
Closed in favour of #454

PR #419, opened 2015-04-02
[WIP] Make zero, one RCP<const Basic>

certik:
This needs to be carefully benchmarked, but it avoids frequent `refcount_` increases and decreases in here:

``` c++
    // Copy constructor
    template<class T2> RCP(const RCP<T2>& r_ptr) : ptr_(r_ptr.get()) {
        if (!is_null()) (ptr_->refcount_)++;
    }
```

which was triggered by expressions like `eq(b, zero)` in here:

``` c++
RCP<const Basic> pow(const RCP<const Basic> &a, const RCP<const Basic> &b)
{
    if (eq(b, zero)) return one;
    if (eq(b, one)) return a;
...
```

The `b` was passed as is (not incrementing the refcount), but both `zero` and `one` was temporary incremented and then decremented. The only difference was that `zero` and `one` was defined as `RCP<const Integer>` and was passed into `RCP<const Basic>` and so the copy constructor above (with the `T2` template) was called and somehow the compiler didn't optimize out the refcount increase. By applying the changes in this PR, the `zero` and `one` are now defined as `RCP<const Basic>` and thus the compiler doesn't need to create a new `RCP` instance with a different template (thus incrementing the refcount), but rather just passes a reference to the same `RCP` (avoiding incrementing the refcount).

See also the section "5.7.2 Conversions between different memory management types" in [TeuchosMemoryManagementSAND](http://web.ornl.gov/~8vt/TeuchosMemoryManagementSAND.pdf).

### DIFF
--- a/src/add.cpp
+++ b/src/add.cpp
@@ -236,7 +236,7 @@ RCP<const Basic> Add::from_dict(const RCP<const Number> &coef, umap_basic_num &&
         } else {
             insert(m, p->first, one);
             insert(m, p->second, one);
-            return rcp(new Mul(one, std::move(m)));
+            return rcp(new Mul(rcp_static_cast<const Integer>(one), std::move(m)));
         }
     } else {
         return rcp(new Add(coef, std::move(d)));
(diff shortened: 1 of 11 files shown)


certik:
Here are the benchmarks:

master: https://gist.github.com/certik/567321a493ef5de5404e
this PR: https://gist.github.com/certik/cb85d7c6e34bc9754377

Things indeed got a bit faster.

shivamvats:
@certik Looks good!
So, is it that the compiler misses all the implicit type conversions and hence makes superfluous refcount increments and decrements in all those cases? We should try to optimize them, too.

src/add.cpp
```diff
@@ -236,7 +236,7 @@ RCP<const Basic> Add::from_dict(const RCP<const Number> &coef, umap_basic_num &&
         } else {
             insert(m, p->first, one);
             insert(m, p->second, one);
-            return rcp(new Mul(one, std::move(m)));
+            return rcp(new Mul(rcp_static_cast<const Integer>(one), std::move(m)));
```
isuruf: This should be `rcp_static_cast<const Number>` right? Otherwise it is first cast to an `Integer` and then to a `Number`

certik:
It looks like it. We should also test other compilers like clang and icc on
this.

src/add.cpp
certik: Good catch.

certik:
I think the right fix is #508.

PR #427, opened 2015-04-25
Implemented Inverse hyperbolic secant: asech

aktech:
@certik  @sushant-hiray  @isuruf 

**Reference:**
* http://mathworld.wolfram.com/InverseHyperbolicSecant.html


src/functions.cpp
```diff
@@ -2364,7 +2364,58 @@ RCP<const Basic> acoth(const RCP<const Basic> &arg)
...
+RCP<const Basic> asech(const RCP<const Basic> &arg)
+{
+    // TODO: Lookup into a cst table once complex is implemented
+    if (eq(arg, one)) return zero;
+    return rcp(new ASech(arg));
+}
+
+RCP<const Basic> ASech::diff(const RCP<const Symbol> &x) const
+{
+    return mul(div(minus_one, mul(sqrt(sub(one, pow(x, i2))), x)), get_arg()->diff(x));
```
isuruf: Here `mul(sqrt(sub(one, pow(x, i2))), x)` should be `mul(sqrt(sub(one, pow(get_arg(), i2))), get_arg())` right?

src/functions.cpp
```diff
@@ -2364,7 +2364,58 @@ RCP<const Basic> acoth(const RCP<const Basic> &arg)
 
 RCP<const Basic> ACoth::diff(const RCP<const Symbol> &x) const
 {
-    return mul(div(one, sub(one, pow(x, i2))), get_arg()->diff(x));
+    return mul(div(one, sub(one, pow(get_arg(), i2))), get_arg()->diff(x));
```
isuruf: Thanks for catching this, is there a test for it?

isuruf: It seems this bug is there in all inverse hyperbolic functions

src/functions.cpp
sushant-hiray: Yes, `x` should be `get_arg()`.

src/functions.cpp
sushant-hiray: @aktech Thanks for catching this. Can you send a new PR and fix this issue for all the other inverse hyperbolic functions too?

src/functions.cpp
aktech: Yes, Indeed.

src/functions.cpp
aktech: @sushant-hiray 
Yes, sure. 
I would open a new PR for that after this gets Merged.

@isuruf 
Yes, the test is [here](https://github.com/sympy/csympy/blob/master/src/tests/basic/test_functions.cpp#L1529).

src/functions.h
```diff
@@ -935,6 +935,33 @@ class ACoth: public HyperbolicFunction {
...
+    ASech(const RCP<const Basic> &arg);
+    /*! Equality comparator
+     * \param o - Object to be compared with
+     * \return whether the 2 objects are equal
+     * */
+    virtual bool __eq__(const Basic &o) const;
+    virtual int compare(const Basic &o) const;
+    //! \return stringify version
+    virtual std::string __str__() const;
+    //! \return `true` if canonical
```
certik: Now when #423 was merged, the `__str__()` method should be removed and rather implemented in the `StrPrinter`.

src/functions.cpp
isuruf: Ah, we should add a new test that is not `acoth(x)` or `acoth(-x)` in a new PR when other inverse hyperbolic functions are fixed

aktech: Indeed.

aktech: @isuruf 
Do you have any other other comments to be addressed in this PR?

certik:
This looks good to me. @isuruf feel free to merge this if you are +1 as well.

isuruf:
Merged. Thanks for the PR @aktech .

aktech:
Thanks @isuruf  @sushant-hiray  @certik !

certik:
Thanks for the patch.

PR #434, opened 2015-05-07
Rename CSymPy to SymEngine

certik:
This only renames things in text files. No code is touched yet.


pbrady:
Why?

certik:
Because CSymPy is a terrible name for a pure C++ library, and even worse name for a Julia library (with no Python in it). SymEngine is the name of the C++ library now, and CSymPy are just the Python wrappers. We do not have a name for the Julia wrappers yet.

pbrady:
Makes a lot of sense.  +1

certik:
The first test passed on Travis (https://travis-ci.org/sympy/symengine/builds/61645391), the second commit only affects README, so I am merging.

PR #442, opened 2015-05-12
Changed ^ to ** in comments

Sumith1896:
Changed ^ to ** in comments of all source as well as test files, 
so as to maintain uniformity throughout.
The printing of power was previously changed from ^ to ** in https://github.com/sympy/symengine/pull/404.

### DIFF
--- a/src/mul.cpp
+++ b/src/mul.cpp
@@ -25,35 +25,35 @@ bool Mul::is_canonical(const RCP<const Number> &coef,
         return false;
     if (dict.size() == 0) return false;
     if (dict.size() == 1) {
-        // e.g. 1*x, 1*x^2
+        // e.g. 1*x, 1*x**2
         if (coef->is_one()) return false;
     }
     // Check that each term in 'dict' is in canonical form
     for (auto &p: dict) {
         if (p.first == null) return false;
         if (p.second == null) return false;
-        // e.g. 2^3, (2/3)^4
+        // e.g. 2**3, (2/3)**4
         // However for Complex no simplification is done
         if ((is_a<Integer>(*p.first) || is_a<Rational>(*p.first))
             && is_a<Integer>(*p.second))
             return false;
-        // e.g. 0^x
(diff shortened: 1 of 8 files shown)


src/mul.cpp
```diff
@@ -128,7 +128,7 @@ RCP<const SymEngine::Basic> Mul::from_dict(const RCP<const Number> &coef, map_ba
     }
 }
 
-// Mul (t^exp) to the dict "d"
+// Mul (t**exp) to the dict "d"
```
Sumith1896: Is this a valid change? I am not sure about this.

src/ntheory.cpp
```diff
@@ -643,7 +643,7 @@ bool crt(const Ptr<RCP<const Integer>> &R, const std::vector<RCP<const Integer>>
         t = rem[i]->as_mpz() - r;
         if (!mpz_divisible_p (t.get_mpz_t(), g.get_mpz_t()))
             return false;
-        r += m * s * (t / g);           // r += m * (m^-1 mod[i]/g)* (rem[i] - r) / g
+        r += m * s * (t / g);           // r += m * (m**-1 mod[i]/g)* (rem[i] - r) / g
```
Sumith1896: Have a look at this too! I hope this is fine.

src/mul.cpp
certik: Yes.

src/ntheory.cpp
certik: Yes, I think that's fine.

certik:
Thanks. I don't think this should change any test results, but let's wait for tests anyway.

Sumith1896:
Agreed. Just felt like doing this clean up while reading the code.

certik:
Much appreciated. If you find more things that need cleanup, feel free to do that.

Sumith1896:
Ready to merge @certik

certik:
Thanks!

PR #444, opened 2015-05-14
[ci skip] An effort to document: Entry guide to contributors

Sumith1896:
To improve the documentation of SymEngine, an entry guide to future contributors also to increase understanding of SymEngine as part of GSoC community bonding.

### DIFF
--- a/doc/symengine.md
+++ b/doc/symengine.md
@@ -1,0 +1,32 @@
+#SymEngine - How Things Work
+
+SymEngine is a fast symbolic manipulation library, written in C++ (SymEngine was previously called CSymPy)
+
+Website for SymPy : http://sympy.org
+
+This document contains information about the decisions made when SymEngine evolved.<br/>
+The plan is this along with [C++ Style Guide](https://github.com/sympy/symengine/blob/master/doc/style_guide.md) and [Design](https://github.com/sympy/symengine/blob/master/doc/design.md) doc should be exhaustive and contain all you need to know to contribute for SymEngine.<br/>
+
+##RCP
+Info about RCP
+
+##Dict
+Info about the dicts used everywhere
+
+##Classes
+Info about basic, other subclasses - about hash, eq<br/>
+
+##Modules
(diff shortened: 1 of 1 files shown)


debugger22:
+1. Nice initiative!

abinashmeher999:
@Sumith1896 Nice initiative! +1

aktech:
+1
Probably, it's one the most important thing SymEngine need to broaden it's community.

Sumith1896:
@certik Why don't we have a wiki option in the SymEngine repo?

abinashmeher999:
@Sumith1896 Also add a section on libraries that we use, and why we use it.

Sumith1896:
@abinashmeher999 I was wondering about using the Wiki option in SymEngine. That'll be open for all and much easier to maintain.
But SymEngine doesn't have a Wiki option, maybe because SymPy does.
Any ideas?

abinashmeher999:
@Sumith1896 I am part of an org. There is a repo whose wiki is empty, and only I can see the option to navigate to that section. And it doesn't appear to one of my friends who is not a part of that org. I think a wiki is visible only to the owners when it is empty. We can't see it unless a page is added to it. After a page is added then we can view the option to navigate there. After that we should automatically see a wiki in our fork of the repo. I am not sure how to send pull requests to a wiki. We should try after @certik has added a homepage to the wiki. Anyway, we can always send patches if we can't figure a way to send pull requests.

OrkoHunter:
> I am not sure how to send pull requests to a wiki.

Wiki pages are open to be *edited* by people. After the one wiki page is created by owners (or maybe collaborators), anyone would be able to edit or create new pages.

certik:
I created the wiki pages here: https://github.com/sympy/symengine/wiki, sorry for the delay.

Sumith1896:
I'm closing this. Maintaining wiki section is a lot easier and cleaner than committing a markdown file.
Thanks for the info.
Cheers

PR #451, opened 2015-05-21
Changed rcp(new(symbol("*") to symbol("*") in tests

Sumith1896:
Following
```
RCP<const Basic> x = rcp(new Symbol("x"));
RCP<const Basic> x = rcp(new Integer(2));
```
were converted to
```
RCP<const Basic> x = symbol("x");
RCP<const Basic> x = integer(2);
```
cc @certik @isuruf

### DIFF
--- a/benchmarks/add1.cpp
+++ b/benchmarks/add1.cpp
@@ -16,6 +16,7 @@ using SymEngine::Add;
 using SymEngine::Mul;
 using SymEngine::Pow;
 using SymEngine::Symbol;
+using SymEngine::symbol;
 using SymEngine::map_vec_int;
 using SymEngine::integer;
 using SymEngine::multinomial_coefficients;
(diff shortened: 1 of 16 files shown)


abinashmeher999:
@Sumith1896 Could you please tell me why was this done?

isuruf:
+1, Can you rebase this on top of master? Also there's some benchmarks and source files that needs to be changed as well.
@abinashmeher999, we should use the `symbol` method and avoid using `new` keyword everywhere, so that it's easier to check whether it is correct.

certik:
Abinash, see also the gitter conversation.

isuruf:
I'm +1 to merge

abinashmeher999:
Seems to good to me too. +1

certik:
Thanks @Sumith1896 for the thorough patch, +1 from me as well. Thanks @isuruf and @abinashmeher999 for the review. merging.

PR #454, opened 2015-05-24
Adding UnivariatePolynomial class

Sumith1896:
@shivamvats Initial commits from https://github.com/sympy/symengine/pull/406.
Merge conflicts resolved.
Will contain the first implementation of Polynomial class in SymEngine.
Tasks at hand before merging this:
- [x] Implement the basic structure with testing env. 
- [x] Implement printer(`__str__()`) with tests.
- [x] Implement `__hash__`, `__eq__`, `compare()`.
- [x] Implement `from_dict()`, `get_args()`, etc.
- [x] Implement `eval()`, `diff()`.
- [x] Implement `neg_poly()`, `sub_poly()`, `mul_poly()`.


src/basic.h
```diff
@@ -29,7 +29,7 @@ namespace SymEngine {
 
 class Visitor;
 class Symbol;
-
+//class Polynomial;
```
sushant-hiray: Any reasons for adding the Polynomial here in the first place?

certik:
Thanks. Let me know when you want me to review it.

src/basic.h
Sumith1896: @shivamvats was the author here, he'd be the better person to answer.

shivamvats: @sushant-hiray It is not required. It seems to be left-over from when I was debugging.

Sumith1896: I'll see to it that it is removed in the next commit. Any other clean-up you find necessary here, please let me know.
cc @sushant-hiray @shivamvats

Sumith1896:
@certik You can suggest any changes now itself, as the PR progresses.
Currently I am assuming that the Travis failure is because Polynomial is not integrated with Visitor pattern.

src/polynomial.cpp
```diff
@@ -0,0 +1,35 @@
+#include "polynomial.h"
+#include "add.h"
+
+namespace SymEngine {
+
+    Polynomial::Polynomial(const std::string& var, map_uint_mpz&& dict) :
```
Sumith1896: Any specific reason using map?

Sumith1896:
Had a first attempt with integrating `Polynomial` with visitor pattern, let me know if this is how you want it to be.

src/polynomial.cpp
shivamvats: Having numbers sorted usually speeds up computations. You might want to look at [this](http://stackoverflow.com/questions/2196995/is-there-any-advantage-of-using-map-over-unordered-map-in-case-of-trivial-keys)

certik: Did you benchmark it? If I remember well, unordered_map was a lot faster.
Make sure we can switch these two, so that we can benchmark it later and
switch if needed.

Sumith1896: Since we are not doing much insertion/deletions I think unordered_map is the way to go. But still let's keep this for now, once we have benchmarks we can try both, choose and adjust accordingly.

shivamvats: @certik In that case, it should not be  too difficult to switch to unordered map.

shivamvats:
@Sumith1896 I do not think this will work. var_ merely holds the variable of the polynomial. To print it properly, you'll need to use the dict_. Did you test the printer?

src/polynomial.cpp
```diff
@@ -0,0 +1,35 @@
...
+    RCP<const Polynomial> add_poly(const Polynomial &a, const Polynomial &b) {
+        map_uint_mpz dict;
+        for(auto &it : a.dict_)
+            dict[it.first] = 0;
+        for(auto &it : b.dict_)
+            dict[it.first] = 0;
+        for(auto &it : a.dict_)
+            dict[it.first] = dict[it.first] + it.second;
+        for(auto &it : b.dict_)
+            dict[it.first] = dict[it.first] + it.second;
```
certik: This looks inefficient. There is probably a better way.

Sumith1896:
@isuruf The last commit has the map_uint_mpz_eq and map_uint_mpz_compare methods we discussed. Could you review it?

Sumith1896:
Can anyone suggest me on how to write the `__hash__()`?

certik:
You need to create a hash of the polynomial. You can adapt the `Add::__hash__()`:
```c++
std::size_t Add::__hash__() const
{
    std::size_t seed = ADD;
    hash_combine<Basic>(seed, *coef_);
    map_basic_num ordered(dict_.begin(), dict_.end());
    for (auto &p: ordered) {
        hash_combine<Basic>(seed, *(p.first));
        hash_combine<Basic>(seed, *(p.second));
    }
    return seed;
}
```
I.e. you convert to using `map_basic_num` (or a similar `map` container) which is ordered, and then compute the hash term by term.

bluescarni:
@certik How about adding the values of the hashes instead? They are unsigned integers, so the order of the operation will not matter. This is how it is done, e.g., in Java:

http://docs.oracle.com/javase/7/docs/api/java/util/Set.html#hashCode()

If the hashes coming out from the individual elements of the original dict are good, using the addition for mixing should not be a problem.

certik:
@bluescarni that's a very good point, that should speed it up. We should do that in `Add` as well. I guess you can do any commutative operation, like add, mul.

bluescarni:
@certik mul can be a problem, if you iterate the multiplication over enough elements the hash will always end up being 0. add does not have this problem.

src/polynomial.h
```diff
@@ -0,0 +1,69 @@
...
+            //return hash_fn(this->var_);
+        //}
+
+        bool __eq__(const Basic &o) const
+        {
+            if ((var_ == static_cast<const Polynomial &>(o).var_) &&
+                map_uint_mpz_eq(dict_, static_cast<const Polynomial &>(o).dict_))
+                return true;
+
+            return false;
```
certik: Move these methods to .cpp. The lest code is in headers, the faster compilation and less clutter.

src/polynomial.h
```diff
@@ -0,0 +1,69 @@
...
+            const Polynomial &s = static_cast<const Polynomial &>(o);
+
+            if (dict_.size() != s.dict_.size())
+                return (dict_.size() < s.dict_.size()) ? -1 : 1;
+
+            int cmp = (var_ < s.var_) ? -1 : 1;
+            if (cmp != 0)
+                return cmp;
+
+            return map_uint_mpz_compare(dict_, s.dict_);
```
certik: Move this to .cpp

src/printer.h
```diff
@@ -27,6 +27,34 @@ class Precedence : public BaseVisitor<Precedence> {
...
+                if (it->first == 0 || it->first == 1) {
+                    precedence = PrecedenceEnum::Atom;
+                }
+                else {
+                    precedence = PrecedenceEnum::Pow;
+                }
+            }
+            else {
+                if (it->first == 0) {
+                    precedence = PrecedenceEnum::Atom;
```
isuruf: `polynomial("x", {{0, -1}}` comes here right? For negative `it->first` it would be `Precedence::Mul`

certik:
@Sumith1896 you were asking what to finish here --- the hashing should be finished. Then mul_poly.

Then we should start benchmarking against Piranha, for example we should allow to use Piranha's integer. Essentially make sure that we get close with the basic operations. Piranha has a fast hash table, it would be nice to use it somehow as well.

src/printer.h
Sumith1896: `it->first` is `uint`, we are handling only `uint` degrees now.
Are you trying to say that `-1` doesn't come under `Atom`, I'm not getting you I think.

isuruf: Yes, `-1` doesn't come under `Atom`. See https://github.com/Sumith1896/csympy/blob/poly_init/src/printer.h#L76.

isuruf: Ah you are right. Degrees are `uint`, so this is okay.

Sumith1896: Thanks @isuruf. I'll update.

src/dict.h
```diff
@@ -103,6 +108,7 @@ std::ostream& operator<<(std::ostream& out, const SymEngine::map_basic_num& d);
 std::ostream& operator<<(std::ostream& out, const SymEngine::map_basic_basic& d);
 std::ostream& operator<<(std::ostream& out, const SymEngine::umap_basic_basic& d);
 std::ostream& operator<<(std::ostream& out, const SymEngine::vec_basic& d);
+std::ostream& operator<<(std::ostream& out, const SymEngine::map_uint_mpz& d);
```
isuruf: This is unimplemented right? Maybe you don't need this.

Sumith1896:
@isuruf Both the changes suggested have been taken care of.

Sumith1896:
The most recent build results can be found here: https://app.shippable.com/builds/556ffaaf7f74c80d007ebbf6

Sumith1896:
I have implemented all the methods I have planned for the first PR.
This is ready to go in now, so review would be needed.
cc @certik @sushant-hiray @isuruf @shivamvats

isuruf:
You'll need to rebase against master

src/printer.cpp
```diff
@@ -171,6 +171,96 @@ void StrPrinter::bvisit(const Pow &x) {
...
+                }
+                else {
+                    s << x.var_->get_name();
+                }
+            }          
+            else {
+                if (first && it->second == -1) {
+                    s << "-" << x.var_->get_name() << "**"  << it->first;
+                }
+                else {
```
isuruf: Formatting `if` and `else` is documented here, https://github.com/sympy/symengine/blob/master/doc/style_guide.md
You have to fix it everywhere.

src/tests/basic/test_polynomial.cpp
```diff
@@ -0,0 +1,169 @@
...
+    RCP<const Polynomial> P = polynomial(x, {{0, 1}, {1, 2}, {2, 1}});
+    assert(P->__str__() == "x**2 + 2*x + 1");
+
+    RCP<const Polynomial> Q = rcp(new Polynomial(x, {1, 0, 2, 1}));
+    assert(Q->__str__() == "x**3 + 2*x**2 + 1");
+}
+
+void test_add_poly()
+{
+	RCP<const Symbol> x  = symbol("x");
```
isuruf: 4 spaces

src/polynomial.cpp
```diff
@@ -0,0 +1,310 @@
...
+        RCP<const Polynomial> c = polynomial(a.var_, std::move(dict));
+        return c;
+    }
+
+    //Calculates bit length of number, used in mul_poly() only
+    template <typename T>
+    uint bit_length(T t){
+        uint count = 0;
+        while (t > 0) {
+        count++;
```
isuruf: 4 spaces before count

src/polynomial.cpp
```diff
@@ -0,0 +1,310 @@
...
+    bool Polynomial::is_pow() const {
+        if (dict_.size() == 1 && dict_.begin()->second == 1 &&
+                dict_.begin()->first != 1 && dict_.begin()->first != 0)
+            return true;
+        return false;
+    }
+
+    RCP<const Polynomial> add_poly(const Polynomial &a, const Polynomial &b) {
+        map_uint_mpz dict;
+        for(auto &it : a.dict_)
```
isuruf: space after `for`

src/polynomial.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+    Polynomial::Polynomial(const RCP<const Symbol> &var, map_uint_mpz&& dict) :
+         var_{var}, dict_{std::move(dict)} {
+
+        map_uint_mpz::iterator it = dict_.begin();
+        uint largest = it->first;
+        uint cur;
+        while(it!=dict_.end()){
+            cur =it->first;
+            if(cur > largest)
```
isuruf: space after `if`

src/polynomial.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+namespace SymEngine {
+
+    Polynomial::Polynomial(const RCP<const Symbol> &var, map_uint_mpz&& dict) :
+         var_{var}, dict_{std::move(dict)} {
+
+        map_uint_mpz::iterator it = dict_.begin();
+        uint largest = it->first;
+        uint cur;
+        while(it!=dict_.end()){
```
isuruf: Should be `while (it != dict_.end()) {`

src/polynomial.cpp
```diff
@@ -0,0 +1,310 @@
+#include "polynomial.h"
+#include "add.h"
+#include "mul.h"
+#include "pow.h"
+#include "constants.h"
+
+namespace SymEngine {
+
+    Polynomial::Polynomial(const RCP<const Symbol> &var, map_uint_mpz&& dict) :
```
isuruf: Rest of `SymEngine` code does not use four spaces like this before `Polynomial`. @certik, should this be changed?

Sumith1896: Yes, compared to the rest of the classes, most of the functions here should be shifted by 4 spaces. Shall I do that?

Sumith1896: I am going to shift it back by 4 spaces to maintain uniformity across SymEngine. @isuruf

certik: Yeah, I would shift it left. Everything is in a namespace, so it would be a
waste of 4 spaces and vertical space is we indent it.

Sumith1896: Done

certik:
So this is a polynomial of only one variable? How will multivariate
polynomials be handled? With the same class? If so, then I think that
should be implemented as well, before we merge, because that will change
the interface.

Sumith1896:
This the univariate class.
I was planning of different classes for univariate and multivariate, at least till multivariate is completely implemented. Most of this will be re-written for multivariate, I don't think changes can be done on top of this PR.

certik:
@Sumith1896 in that case let's rename this class to something like UnivariatePolynomial.

Sumith1896:
I had that in mind, what about `Sup` for Sparse Univariate Polynomial?

src/polynomial.cpp
```diff
@@ -0,0 +1,307 @@
...
+{
+    std::hash<std::string> hash_string;
+    std::hash<uint> hash_uint;
+    std::hash<long long int> hash_si;
+    std::size_t seed = POLYNOMIAL;
+
+    seed += hash_string(this->var_->get_name());
+    for (auto &it : this->dict_)
+    {
+        seed += hash_uint(it.first) + hash_si(it.second.get_si());
```
certik: You should combine the hashes for `it.first` and `it.second`, not just with a `+`. The only place you use `+` is for combining hashes for different `it`, as you do. Use the `hash_combine` function.

src/polynomial.h
```diff
@@ -0,0 +1,97 @@
...
+
+namespace SymEngine {
+//! Polynomial Class
+class Polynomial : public Basic{
+public:
+    //! `degree` : Degree of Polynomial
+    //! `var_` : Variable of the uni-variate Polynomial
+    //! `dict_` : holds the Polynomial
+    // Polynomial x**2 + 2*x + 1 has dict_ = {{0, 1}, {1, 2}, {2, 1}} with var_ = "x" 
+    uint degree;
```
certik: Use a trailing `_` for internal variables. So rename `degree` to `degree_`.

src/polynomial.cpp
```diff
@@ -0,0 +1,307 @@
...
+    map_uint_mpz::iterator it = dict_.begin();
+    uint largest = it->first;
+    uint cur;
+    while (it != dict_.end()) {
+        cur =it->first;
+        if (cur > largest)
+            largest = cur;
+        it++;
+    }
+    degree = largest;
```
certik: This is calculating the degree of the polynomial, right? I think this should be passed in as a parameter. Many time the user (or the caller) knows the degree, so you just want to pass it in. Check in `is_canonical` that the degree is equal to the calculated one. That way the constructor is very fast in Release mode.

certik:
I would just call it `UnivariatePolynomial`. And the function `univariate_polynomial`. 

It is sparse, but there are many choices to implement even a sparse polynomial, and we just need to have a good enough class for univariate and multivariate polynomials. So I would just call it `UnivariatePolynomial` and `Polynomial` (for multivariate polynomials). These two classes are high level, because they can take part in SymPy expressions.

I can imagine later implementing lower level classes with various data structures, as well as using Piranha. These lower level classes do not use RCP at all, thus they could be faster for some applications.

The two high level classes `UnivariatePolynomial` and `Polynomial` are meant to be used by the user to get a decent speed for polynomials, and we should choose some fast representation for those so that it works fast for most cases, and I think it needs to be sparse, and we should benchmark it to see whether to use std::map or std::unordered_map or other specialized data structures. There are many many ways how these could be implemented, and for SymEngine I just want us to make some very good default choices and just call the classes `UnivariatePolynomial` and `Polynomial`. The user could then call specialized classes if needed for a given application (if we implement any).

certik:
P.S. You need to resolve the conflicts with master.

src/polynomial.cpp
Sumith1896: It can be passed as a parameter, but in many internal calls then we have to calculate degree from dict, then pass. Won't this be an overhead?

Sumith1896:
All the suggestions have been taken care of.
Let me know if there are any other changes.

src/polynomial.cpp
certik: Can you show me an internal call that needs to calculate it?

certik:
This looks good to me. One thing that I didn't find was a test that tests the `UnivariatePolynomial` inside an expression, i.e. something like `(x^3+x^2+x)^3` or more complicated. To make sure that things work properly.

src/polynomial.cpp
```diff
@@ -0,0 +1,297 @@
...
+            if (d.begin()->second == 0)
+                return zero;
+            else if (d.begin()->second == 1) 
+                return pow(var, integer(d.begin()->first));
+            else 
+                return Mul::from_dict(integer(d.begin()->second), 
+                    {{var, integer(d.begin()->first)}});
+        }
+    } else {
+        return rcp(new UnivariatePolynomial(var, (--(d.end()))->first, std::move(d)));
```
Sumith1896: @certik Here

src/polynomial.cpp
```diff
@@ -0,0 +1,297 @@
...
+}
+
+RCP<const Basic> UnivariatePolynomial::diff(const RCP<const Symbol> &x) const
+{
+    if (var_->__eq__(*x)) {
+        map_uint_mpz d;
+        for (auto &p : dict_) {
+            d[p.first - 1] = p.second * p.first;
+        }
+        return rcp(new UnivariatePolynomial(var_, (--(d.end()))->first, std::move(d)));
```
Sumith1896: @certik Here too, notice that since it's a map things are easier.

src/polynomial.cpp
certik: So this `from_dict` is used when you do not know the degree, isn't it? So there must be a code that calculates it, in this case. The way you did it is one way --- does it use the fact that the std::map is ordered properly?

It's just that the code to calculate the degree should happen before calling the `UnivariatePolynomial` constructor.

Sumith1896: Yes, `from_dict` is used to construct `Polynomial` when we do not know the degree.
It uses the fact the std::map is ordered.

src/tests/basic/test_polynomial.cpp
```diff
@@ -0,0 +1,185 @@
...
+void test_expand()
+{
+    RCP<const Symbol> x  = symbol("x");
+    RCP<const UnivariatePolynomial> a = univariate_polynomial(x, 3, {{1, 1}, {2, 1}, {3, 1}});
+    RCP<const Basic> b = rcp(new Pow(a, integer(3)));
+    RCP<const Basic> c = expand(b);
+
+    assert(b->__str__() == "(x**3 + x**2 + x)**3");
+    //std::cout<<b->__str__()<<std::endl;
+    assert(c->__str__() == "x**9 + 3*x**8 + 6*x**7 + 7*x**6 + 6*x**5 + 3*x**4 + x**3");
```
Sumith1896: @certik Do these kind of tests suffice or should I add something more?

Sumith1896:
Tests pass. All that I had planned for this PR is implemented, it's ready to merge unless there are any more changes suggested.

src/tests/basic/test_polynomial.cpp
certik: I think that's good enough.

certik:
I think this looks good, merging. Thanks!

Next step is to tackle multivariate?

Sumith1896:
Yes. We could meet to discuss the further tasks at hand.

PR #457, opened 2015-05-27
Openmp support

thilinarmtb:
Add OpenMP support so that we can use it for parallel algorithms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,20 +87,6 @@ if (WITH_ARB)
     set(HAVE_SYMENGINE_ARB yes)
 endif()
 
-# OpenMp
-set(WITH_OPENMP no
-    CACHE BOOL "Build with OpenMP")
-
-if (WITH_OPENMP)
-    find_package(OpenMP REQUIRED)
-    if(OPENMP_FOUND)
-        set(CMAKE_CXX_FLAGS_DEBUG
-            "${CMAKE_CXX_FLAGS_DEBUG} ${OpenMP_CXX_FLAGS}")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -87,6 +87,20 @@ if (WITH_ARB)
     set(HAVE_SYMENGINE_ARB yes)
 endif()
 
+# OpenMp
+set(WITH_OPENMP no
+    CACHE BOOL "Build with OpenMP")
+
+if (WITH_OPENMP)
+    find_package(OpenMP REQUIRED)
```
certik: Is this `OpenMP` module distributed with cmake?

certik:
This is good. But I think when openmp is enabled, we have to use the thread safe version of symengine.

CMakeLists.txt
thilinarmtb: Yes.

thilinarmtb:
I'll enable it.

thilinarmtb:
Code block for finding OpenMP was placed after the line:

`set(WITH_SYMENGINE_THREAD_SAFE no CACHE BOOL "Enable SYMENGINE_THREAD_SAFE support")`

to avoid `SYMENGINE_THREAD_SAFE` being set to `no` for inconsistent user input
like the following:

`cmake -DWITH_OPENMP=yes WITH_SYMENGINE_THREAD_SAFE=no .`

certik:
Very good, thanks! +1 to merge if tests pass.

thilinarmtb:
Thanks for merging !!

PR #460, opened 2015-06-03
Use initializer list for Integer::Integer()

certik:
This provides a nice few percent speedup in e.g. the expand2 benchmark.


thilinarmtb:
+1

PR #464, opened 2015-06-06
Add support for Piranha in CMake

Sumith1896:
@certik @isuruf @sushant-hiray


CMakeLists.txt
```diff
@@ -96,6 +96,17 @@ if (WITH_MPFR)
...
 endif()
 
+# Piranha
+set(WITH_PIRANHA no
+    CACHE BOOL "Build with Piranha")
+
+if (WITH_PIRANHA)
+    find_package(Piranha REQUIRED)
+    include_directories(${PIRANHA_INCLUDE_DIRS})
+    set(LIBS ${LIBS} pthread)
```
isuruf: `${LIBS}` needs to be given the path of the library `pthread`.

Sumith1896: Should I replace `pthread` with `${PIRANHA_LIBRARIES}`?

certik: `pthread` is a library that contains threads. It is simply a dependency of PIranha. Another dependency is boost.

So we should simply add the following:
```
    find_package(BOOST REQUIRED)
    include_directories(${BOOST_INCLUDE_DIRS})
    set(LIBS ${LIBS} ${BOOST_LIBRARIES})
```
And the same for `PTHREAD`.

Then we need to add the two packages (`PTHREAD` and `BOOST`) as well. 

The `PTHREAD` package internally can use this: http://stackoverflow.com/questions/15193785/how-to-get-cmake-to-recognize-pthread-on-ubuntu, the `BOOST` package should use our usual library finding code.

certik: @isuruf, perhaps instead of adding the dependencies like this, we should rather just
`` `
set(WITH_BOOST yes)
set(WITH_PTHREAD yes)
```
and then later have the usual `if (WITH_BOOST) ...`. The reason is that then the user can add `BOOST` but not `PIRANHA`. The same for `ARB` and its dependencies `MPFR` and `FLINT`. Now when we have `eval_mpfr`, you want to be able to sometimes link with `MPFR` but not with `ARB`.

Comment (isuruf) [inline CMakeLists.txt]: @certik, yes, that's what's happening now for `ARB` and `MPFR`
```CMake
if (WITH_ARB)
    find_package(FLINT REQUIRED)
    include_directories(${FLINT_INCLUDE_DIRS})
    set(LIBS ${LIBS} ${FLINT_LIBRARIES})

    # Workaround for https://github.com/fredrik-johansson/arb/issues/24
    include_directories(${FLINT_INCLUDE_DIRS}/flint)

    set(WITH_MPFR yes)

    find_package(ARB REQUIRED)
    include_directories(${ARB_INCLUDE_DIRS})
    set(LIBS ${LIBS} ${ARB_LIBRARIES})
    set(HAVE_SYMENGINE_ARB yes)
endif()
```
We could do the same for `FLINT` as well.

isuruf: @certik, CMake has `FindBoost.cmake` as well. (It's there in v2.8)

certik: It will not support our COMMON_DIR though.

isuruf: Yes. I forgot about that one.

CMakeLists.txt
```diff
@@ -96,6 +96,41 @@ if (WITH_MPFR)
...
+
+if (WITH_PTHREAD)
+    find_package(PTHREAD REQUIRED)
+    include_directories(${PTHREAD_INCLUDE_DIRS})
+    set(LIBS ${LIBS} ${PTHREAD_LIBRARIES})
+    set(HAVE_SYMENGINE_PTHREAD yes)
+endif()
+
+if (WITH_BOOST)
+    find_package(Boost REQUIRED COMPONENTS date_time filesystem system)
```
Sumith1896: Is the expected version?

Sumith1896:
FindBoost.cmake is a file off the internet, I have not written it.
I still have to address @isuruf's comments.

isuruf:
`FindBoost.cmake` is in the `CMake` installation. So you don't need to include it.

Sumith1896:
Just realized it, removed.

CMakeLists.txt
certik: I think that's fine. It won't find our COMMON_DIR, but we can improve this later.

certik:
That looks good. You need to resolve conflicts with master and make sure Travis tests pass.

CMakeLists.txt
Sumith1896: I don't have much idea of finding COMMON_DIR, I'll open an issue so that we can keep track of this.

Sumith1896:
@certik Any changes needed?

certik:
I think this is good enough. We can improve once it is merged if needed. Thanks!

Sumith1896:
Thanks for review @isuruf @certik.
I'll open an issue, we can keep track of quality of this.

certik:
No need. We'll open an issue if there is a bug.

Sumith1896:
Okay

PR #469, opened 2015-06-10
Rename `csympy` --> `symengine`

thilinarmtb:


symengine/lib/symengine_wrapper.pyx
```diff
@@ -9,50 +9,50 @@ from cpython cimport PyObject, Py_XINCREF, Py_XDECREF, \
...
         r = FunctionWrapper.__new__(FunctionWrapper)
-    elif (csympy.is_a_RealDouble(deref(o))):
+    elif (symengine.is_a_RealDouble(deref(o))):
         r = RealDouble.__new__(RealDouble)
-    elif (csympy.is_a_ComplexDouble(deref(o))):
+    elif (symengine.is_a_ComplexDouble(deref(o))):
         r = ComplexDouble.__new__(ComplexDouble)
     else:
-        raise Exception("Unsupported CSymPy class.")
+        raise Exception("Unsupported symengine class.")
```
thilinarmtb: This should be `SymEngine`.

symengine/lib/symengine_wrapper.pyx
```diff
@@ -9,50 +9,50 @@ from cpython cimport PyObject, Py_XINCREF, Py_XDECREF, \
...
     r.thisptr = o
     return r
 
-def sympy2csympy(a, raise_error=False):
+def sympy2symengine(a, raise_error=False):
     """
-    Converts 'a' from SymPy to CSymPy.
+    Converts 'a' from SymPy to symengine.
```
thilinarmtb: `SymEngine`

certik:
I think we initially wanted to use csympy for the Python wrappers, but I am fine either way.

symengine/tests/test_sympy_conv.py
```diff
@@ -1,14 +1,14 @@
-from csympy import (Symbol, Integer, sympify, SympifyError, sin, cos,
+from symengine import (Symbol, Integer, sympify, SympifyError, sin, cos,
         function_symbol, I, E, pi, exp)
-from csympy.lib.csympy_wrapper import densematrix, Subs, Derivative
+from symengine.lib.symengine_wrapper import densematrix, Subs, Derivative
 import sympy
 
-# Note: We test _sympy_() for CSymPy -> SymPy conversion, as those are methods
+# Note: We test _sympy_() for symengine -> SymPy conversion, as those are methods
```
thilinarmtb: `SymEngine`

symengine/tests/test_sympy_conv.py
```diff
@@ -1,14 +1,14 @@
...
 # that are implemented in this library. Users can simply use sympy.sympify(...)
 # to do this conversion, as this function will call our _sympy_() methods under
 # the hood.
 #
-# For SymPy -> CSymPy, we test csympy.sympify(...) which does the conversion.
+# For SymPy -> symengine, we test symengine.sympify(...) which does the conversion.
```
thilinarmtb: `SymEngine`

thilinarmtb:
Ah, I didn't know that. That makes sense.
@isuruf : WDYT?

certik:
I am fine with naming it symengine. It will be consistent with Julia and Ruby wrappers.

Sumith1896:
Isn't there possibility of confusion with plain `SymEngine` for wrappers?
Since it resides in the same repo.

thilinarmtb:
@certik : Great !! I'll fix the errors.

@Sumith1896 : It was called `csympy` when the repo was `csympy`?

Sumith1896:
Oh It was completely out of mind that previously we called it `csympy`. I think it is fine then.

certik:
There are also test failures like:
```
+mkdir -p empty
+cd empty
+cat
+python
ImportError: No module named csympy
```
So you need to update the Travis tests.

I am fine with naming the python module symengine. I think it's a better name and less confusion.

Sumith1896:
+1

certik:
This is a good change, I think it's the way to go. I am merging it.

thilinarmtb:
Thanks for merging !!

PR #471, opened 2015-06-13
Adding non-implemented functions in Complex

mafiya69:
some non-implemented overloads for `rdivcomp` added.

### DIFF
--- a/src/complex.h
+++ b/src/complex.h
@@ -273,6 +273,8 @@ public:
     virtual RCP<const Number> rdiv(const Number &other) const {
         if (is_a<Integer>(other)) {
             return rdivcomp(static_cast<const Integer&>(other));
+        } else if (is_a<Rational>(other)) {
+            return rdivcomp(static_cast<const Rational&>(other));
         } else {
             throw std::runtime_error("Not implemented.");
         }
@@ -281,8 +283,6 @@ public:
     virtual RCP<const Number> pow(const Number &other) const {
         if (is_a<Integer>(other)) {
             return powcomp(static_cast<const Integer&>(other));
-        } else if (is_a<Rational>(other)) {
-            return rdivcomp(static_cast<const Rational&>(other));
         } else {
             return other.rpow(*this);
         }
--- a/src/complex.h
+++ b/src/complex.h
@@ -194,6 +194,18 @@ public:
             return from_mpq((this->real_ * other.i) / conjugate, (this->imaginary_ * (-other.i)) / conjugate);
         }
     }
+    /*! Divide other by the Complex
+    * \param other of type Rational
+    * */
+    inline RCP<const Number> rdivcomp(const Rational &other) const {
+        mpq_class conjugate = this->real_*this->real_ + this->imaginary_*this->imaginary_;
+        if (conjugate.get_num() == 0) {
+            throw std::runtime_error("Divide by zero.");
+        } else {
+            return from_mpq((this->real_ * other.i) / conjugate, (this->imaginary_ * (-other.i)) / conjugate);
+        }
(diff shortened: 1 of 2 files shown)


src/complex.h
```diff
@@ -194,6 +194,34 @@ class Complex : public Number {
...
+            throw std::runtime_error("Divide by zero.");
+        } else {
+            return from_mpq((this->real_ * other.i) / conjugate, (this->imaginary_ * (-other.i)) / conjugate);
+        }
+    }
+
+    /*! Divide other by the Complex
+     * \param other of type Complex
+     * */
+     inline RCP<const Number> rdivcomp(const Complex &other) const {
```
isuruf: We don't really need this method because of `divcomp(const Complex &other) const {` method of `Complex`

mafiya69: @isuruf I will remove that, is everything else ok?

src/complex.h
```diff
@@ -229,6 +242,8 @@ class Complex : public Number {
             return rsubcomp(static_cast<const Rational&>(other));
         } else if (is_a<Integer>(other)) {
             return rsubcomp(static_cast<const Integer&>(other));
+        } else if (is_a<Complex>(other)){
+            return rsubcomp(static_cast<const Complex&>(other));
```
isuruf: These two lines are not needed as well

src/complex.h
```diff
@@ -261,6 +276,10 @@ class Complex : public Number {
     virtual RCP<const Number> rdiv(const Number &other) const {
         if (is_a<Integer>(other)) {
             return rdivcomp(static_cast<const Integer&>(other));
+        } else if (is_a<Rational>(other)) {
+            return rdivcomp(static_cast<const Rational&>(other));
+        } else if (is_a<Complex>(other)) {
+            return (static_cast<const Complex&>(other).divcomp(*this));
```
isuruf: Since we are not calling `rdiv` directly, lines 281-282 are not needed as well.

mafiya69:
@isuruf

certik:
This looks good, thanks. Can you add a test as well? That fails in master
and that your patch fixes.

mafiya69:
@certik what does shippable basically do? I am asking because `ctest` is passing on my system.

certik:
I think you can see for yourself in Shippable --- you just need to login using your github account (for example). I looked in there, looks like all tests pass except the last 3, which fail due to some Shippable bug -- it doesn't say why. So I wouldn't worry about it.

mafiya69:
@certik will surely try to add tests for `Complex` as there are none

mafiya69:
@certik I have added some tests, Have a look and let me know if I am doing something wrong.

isuruf:
@mafiya69, 
Tests are here, https://github.com/sympy/symengine/blob/master/src/tests/basic/test_basic.cpp#L485

mafiya69:
@isuruf as there are `divnum`, `subnum` etc global functions for `div`, `sub` etc. ([Here](https://github.com/sympy/symengine/blob/master/src/number.h#L57)), shouldn't there be `rdivnum`, `rsubnum` etc for `rdiv`, `rsub` etc. ?

isuruf:
You could always use `divnum(y, x)` instead of `rdivnum(x, y)`. That's why they are not implemented.

isuruf:
If you could add some tests that use, `Complex::rdivcomp(const Rational &other)` then we can merge this.

mafiya69:
@isuruf

isuruf:
+1

Sumith1896:
@mafiya69 rebase needed, look into the updated test files in master to modify your tests.

PR #480, opened 2015-06-22
Support for Symbol in ruby wrappers

abinashmeher999:
I guess the files `cwrapper.h` had sufficient functions to create a `Symbol` class in module `SymEngine` and use the `expand` method. These just contain the Ruby wrappers but they don't build now.
**TODO:**
- [x] Add `#free_symbols` method to basic
- [x] Add `#args` method to basic
- [x] Write tests for both


abinashmeher999:
@certik This is now ready for review :sparkles:

certik:
@abinashmeher999 can you rebase this on top of the latest master please?

abinashmeher999:
@certik I have rebased it.

src/ruby/ext/symengine/ruby_basic.c
```diff
@@ -75,3 +75,21 @@ VALUE cbasic_pow(VALUE self, VALUE operand2){
...
 }
+
+VALUE cbasic_to_str(VALUE self){
+    basic_struct *this;
+    char *str_ptr;
+    VALUE result;
+
+    Data_Get_Struct(self, basic_struct, this);
+
+    str_ptr = basic_str(this);
```
isuruf: You have to free the data allocated here.

isuruf: My mistake

certik:
Seems like Travis didn't pick up the latest commit, I am not sure why.

src/ruby/spec/symbol_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+            context 'with nothing as argument' do
+                it 'returns the expanded form' do
+                    x = SymEngine::Symbol.new("x")
+                    y = SymEngine::Symbol.new("y")
+                    z = SymEngine::Symbol.new("z")
+                    e = (x+y+z)*(x+y+z)
+                    f = e.expand()
+                    expect(e.to_s).to eql("(x + y + z)**2")
+                    expect(f).to be_an_instance_of SymEngine::Symbol
+                    expect(f.to_s).to eql("2*x*y + 2*x*z + 2*y*z + x**2 + y**2 + z**2")
```
certik: In general we should not compare to a string like this, as the order of terms might be platform dependent, though I think we did some work to make it print the same order on all platforms.

You should implement a comparison operator (using `eq` from SymEngine) and use that instead.

certik:
I think this looks good. If tests pass, we can merge it.

abinashmeher999:
The commit that has the yellow dot is the latest one. I don't know how github orders commits after they are rebased, but that is the latest one when I run `git log`

certik:
I see. The tests passed in https://travis-ci.org/sympy/symengine/builds/68134316.

src/ruby/spec/symbol_spec.rb
abinashmeher999: Function result platform dependent! Could you please give any hints how?

Even if we implement the something like the `eq` method, how do we standardise the output, something to say is correct and test against. Because testing `expand` in that case will also require a similar invocation.

certik:
#481 was merged, would you mind rebasing this on top? You have to move it from src to symengine.

src/ruby/spec/symbol_spec.rb
certik: We use `std::unordered_map` to store the individual terms in `Add`. When you loop over `std::unordered_map`, there are no guarantees on which specific element order is considered. That's why it is called "unordered".

abinashmeher999:
@certik Done. There are two builds now pending for the same PR. The older one can be cancelled.

src/ruby/spec/symbol_spec.rb
abinashmeher999: I see. Thanks! :blush:. How do we test it then?

certik:
Ok, I canceled the old one.

certik:
You need to move the files from src/ruby to symengine/ruby.

src/ruby/spec/symbol_spec.rb
certik: As I said, you test by using the `eq` function and you construct the expression `2*x*y + 2*x*z + 2*y*z + x**2 + y**2 + z**2`. That's how you should be comparing expressions, not using their string representation (that can also change if we change from `**` to `^` or vice versa as we did in the past).

isuruf: Printing order should be the same across all platforms. If not it's a bug.
You could test it by reconstructing the expanded term and then using an `eq` function.

symengine/ruby/spec/basic_spec.rb
```diff
@@ -16,56 +16,5 @@
...
-
-        describe 'binary operations' do
-            before :each do
-                @a = SymEngine::Basic.new
-                @b = SymEngine::Basic.new
-            end
-            describe '#+' do
-                context 'with another Basic object as argument' do
-                    xit 'returns a Basic object that is result of
-                        self added to the argument' do
```
isuruf: I'm not getting why this was removed.

src/ruby/spec/symbol_spec.rb
certik: Yes, I think you fixed it @isuruf. Still, one should construct the RHS directly to compare. That's much more robust.

abinashmeher999: @certik I get it now. You are suggesting to compare the expressions, not the strings. Took some time to get to that, Sorry. I will change it once I support that function.

symengine/ruby/spec/basic_spec.rb
abinashmeher999: These didn't test anything. The `xit` keyword marks them as pending. As you said, to test these methods, the `Basic` must be initialised. If we are initialising it with a `Symbol`, I thought it would be better to club it with other tests of `Symbol`.

isuruf: All the operators here are common to all `Basic` classes. So, I think it makes sense to keep them here. You could have `symbols`, that's okay. I would define only `Symbol` specific methods in the other file.

certik: Right. You can follow the Python tests: https://github.com/sympy/symengine/blob/062a20759808ff4f657e879bcb03818c51a75094/symengine/python/symengine/tests/test_arit.py

symengine/ruby/ext/symengine/symengine.c
```diff
@@ -22,4 +24,11 @@ void Init_symengine() {
     rb_define_method(c_basic, "/", cbasic_div, 1);
     rb_define_method(c_basic, "**", cbasic_pow, 1);
     rb_define_method(c_basic, "-@", cbasic_neg, 0);
+    rb_define_method(c_basic, "to_s", cbasic_to_str, 0);
+    rb_define_method(c_basic, "expand", cbasic_expand, 0);
+
+    //Symbol class
+    c_symbol = rb_define_class_under(m_symengine, "Symbol", c_basic);
+    rb_define_alloc_func(c_symbol, csymbol_alloc);
```
isuruf: Is it possible to use `cbasic_alloc` here?

symengine/ruby/spec/basic_spec.rb
abinashmeher999: I didn't get the `symbols` part. Could you please elaborate? 
Yes, it does make sense. Testing it once in `Basic` will cover it, but what should I initialise `Basic` object with?

isuruf: What I meant was since `Symbol` is a `Basic`, think of it as a `Basic`. Instead of `@a = SymEngine::Basic.new` you could use  `a = SymEngine::Symbol.new("a")`

symengine/ruby/ext/symengine/symengine.c
abinashmeher999: Yes, it is. Both are the same, as you noticed. This was just for clarity in code. We will have the freedom to modify what `csymbol_alloc` does, without having to change anything in this file in the future.

symengine/ruby/spec/basic_spec.rb
abinashmeher999: Ok. I can do that now, since `Symbol` has been added.

abinashmeher999:
@certik Shall we put this on hold, till I reproduce the python tests?

certik:
Address the changes, most importantly move the files to the symengine directory. Make sure tests pass.

In the meantime, you can implement the Integer and Rational. Or you can finish this PR. It's up to you, whatever is the most efficient for you.

src/ruby/ext/symengine/ruby_symbol.h
```diff
@@ -0,0 +1,9 @@
+#include "ruby.h"
+#include "ruby_basic.h"
+#include "cwrapper.h"
+
+void csymbol_free(void *ptr);
```
isuruf: This is not used anywhere, right?

abinashmeher999: No, it isn't. But it might be required in future, if I use `Data_Make_Struct` to make a new object of `Symbol` in ruby wrapper code. I can remove it for the time being.

certik: How are things deallocated in the wrappers?

abinashmeher999: When we make an object of a certain class of Ruby in the Ruby wrapper code, we use a method `Data_Make_Struct` as used [here](https://github.com/sympy/symengine/pull/480/files#diff-4c1a05fa0d0faea4de4f1b04b88cce71R47). It allocates the space for the type, wraps the c pointer for it with the Ruby class we define. It also takes the function to free the wrapped c pointer, to be used when Garbage Collector is called on the formed object.
More can be found [here](http://clalance.blogspot.in/2013/11/writing-ruby-extensions-in-c-part-13.html)

certik: Cool, thanks. Did you try to run the test suite in Valgrind? Just to make sure it doesn't leak.

abinashmeher999: No, I haven't. But I plan to do that. I was going through a few blog posts trying to find a more Ruby-like way to find memory leaks. Something like rspec. I will give them a try first. Like, [this answer](http://stackoverflow.com/questions/20956401/how-do-i-track-down-a-memory-leak-in-my-ruby-code#20956962) on stackoverflow.

symengine/ruby/ext/symengine/symengine.c
isuruf: I just looked at how inheritance works and when you create a new `Symbol` instance both `cbasic_alloc` and `csymbol_alloc` gets called since `Symbol` is a child of `Basic`

abinashmeher999: But I don't think that's how it works with the Ruby C API. The function is just given to allocate such an instance and return the object. It's just the C API counterpart of ".allocate" method in ruby. I decided to check it out and added a `printf` statement in `cbasic_alloc`.
``` c
VALUE cbasic_alloc(VALUE klass){
    VALUE obj;
    basic_struct *struct_ptr;

    obj = Data_Make_Struct(klass, basic_struct, NULL, cbasic_free, struct_ptr);

    struct_ptr->data = NULL;
    printf("basic object allocated\n");

    return obj;
}
```
This is the output of the test files
``` bash
SymEngine
  SymEngine::Basic
    .new
      with no arguments
basic object allocated
        returns a Basic object

SymEngine
  SymEngine::Symbol
    .new
      with an integer as an argument
basic object allocated
        returns a Symbol object
    binary operations
      #+
        with another Symbol object as argument
basic object allocated
basic object allocated
          returns a Symbol object that is result of
                        self added to the argument
      #-
        with another Symbol object as argument
basic object allocated
basic object allocated
          returns a Symbol object that is result of
                        argument subtracted from self
      #*
        with another Symbol object as argument
basic object allocated
basic object allocated
          returns a Symbol object that is result of
                        self multiplied by the argument
      #/
        with another Symbol object as argument
basic object allocated
basic object allocated
          returns a Symbol object that is result of
                        self divided by the argument
      #**
        with another Symbol object as argument
basic object allocated
basic object allocated
          returns a Symbol object that is result of
                        self raised to the power of argument
    unary operations
      #-@
        doesn't take any argument
basic object allocated
          returns the negation of self
    #expand
      with nothing as argument
basic object allocated
basic object allocated
basic object allocated
        returns the expanded form

Finished in 0.00452 seconds (files took 0.09732 seconds to load)
9 examples, 0 failures
```
This shows they are not being called twice.

abinashmeher999: The way you mentioned happens in C++, but not true for C API.

src/ruby/spec/symbol_spec.rb
abinashmeher999: @isuruf Could you please point me to the function `eq` by which I can compare two expressions? I tried `_eq_` in `Basic` but that doesn't quite work as expected.

isuruf: `__eq__` is an internal function. You should use the function 
```
bool eq(const RCP<const Basic> &a, const RCP<const Basic> &b)
```
which calls `__eq__` inside.

abinashmeher999: It works now. If I compare `e` and `f`, in this case, should it give `true` or `false`?

isuruf: `False` because, `eq` checks structural equality not mathematical equality

abinashmeher999: Is there a method to check mathematical equality too? That will be nice to have in this minimal set of functions.

symengine/ruby/ext/symengine/symengine.c
isuruf: Okay. So, this means that each instance of `Basic` needs to have the same struct right? (because `+` is defined for `Basic` and that function expects a the sBasic)

abinashmeher999: That should be how it happens. Still I can't say for sure, because Ruby variables don't have a type, the data has the type and every Ruby object in the C API is a `VALUE` which is void pointer to the struct. All `VALUE`s have a `Type` which tells us the type of the data. There's a possibility that the `Basic` object has a pointer to the actual struct(that we allocated) that contains the information which it passes around when we want to call a function, and that function will work since it gets the pointer to the data that the `Basic` object wraps, the struct it was defined for, the actual struct that contains the information.

src/ruby/spec/symbol_spec.rb
certik: In SymPy that's called simplify (f-g) == 0. We do not have simplify yet in
SymEngine until we implement more advanced polynomial algorithms.

certik:
@abinashmeher999 Looks like you figured it out. Let me know if you have any further questions regarding this PR.

abinashmeher999:
@certik could you point me to the SymEngine counterpart of `free_symbols` used [here](https://github.com/sympy/symengine/blob/062a20759808ff4f657e879bcb03818c51a75094/symengine/python/symengine/tests/test_arit.py#L126).
And for `test_arit`, I will add the corresponding test file in rspec, in a different PR. It needs the `Integer` class right from the first test case. Also till then I can search for a less verbose alternative to assert which is `expect(<expression for assertion>).to be true`

certik:
@abinashmeher999 sure, see the end of symengine/visitor.h for the free_symbols function. What about test_arit?

abinashmeher999:
@certik thanks! I am trying to reproduce all the test cases in `test_arit.py` in rspec, and that requires `Integer` class also it has `assert`s whose counterpart in rspec(what I have found till now) is very verbose, I am trying to search for a shorter one. I will include the `arit_spec`(the reproduced test cases) with the `Integer` and `Rational` class PR.

certik:
Ok, that works.

symengine/tests/cwrapper/test_cwrapper.c
```diff
@@ -2,6 +2,7 @@
 
 void test_cwrapper() {
     char* s;
+    basic_struct *basic_array;
```
certik: This line doesn't seem to be needed here.

symengine/ruby/ext/symengine/ruby_basic.c
```diff
@@ -72,6 +76,90 @@ VALUE cbasic_pow(VALUE self, VALUE operand2){
...
+
+    size = vecbasic_size(args);
+    VALUE ruby_array = rb_ary_new2(size);
+    int i = 0;
+    VALUE temp = NULL;
+    for(i = 0; i < size; i++) {
+        basic_init(iterator_basic);
+        vecbasic_get(args, i, iterator_basic);
+        temp = Data_Make_Struct(rb_obj_class(self), basic_struct, NULL,
+                                                    cbasic_free, iterator_basic);
```
certik: So you are assigning `iterator_basic`, without copying it to a new variable, is that right?

symengine/ruby/ext/symengine/ruby_basic.c
```diff
@@ -72,6 +76,90 @@ VALUE cbasic_pow(VALUE self, VALUE operand2){
...
+        rb_ary_push(ruby_array, temp);
+        basic_free(iterator_basic);
```
certik: So in there you make `iterator_basic` invalid, so the pointer in there becomes dangling.

certik: And then it fails later.

symengine/tests/cwrapper/test_cwrapper.c
abinashmeher999: Oh, it isn't. I am removing it. Thanks!

symengine/ruby/ext/symengine/ruby_basic.c
certik: You need to use `basic_assign` if you want to copy the variable to a new one. In this case, you just want to remove the `basic_free(iterator_basic)`, that should fix the problem. My understanding is that Ruby will free the expression when needed using `cbasic_free`, is that right? So there should be no leaks.

symengine/ruby/ext/symengine/ruby_basic.c
abinashmeher999: Yes, I am wrapping the C pointer and making a new object in ruby.

symengine/ruby/ext/symengine/ruby_basic.c
abinashmeher999: Ah, yes. I shouldn't have done this, freeing the pointer. Ruby will indeed call `cbasic_free` to free the object.

abinashmeher999: The new loop should work something like this right?
```c
    basic_init(iterator_basic);
    for(i = 0; i < size; i++) {
        basic_struct *temp_basic;
        basic_init(temp_basic);

        vecbasic_get(args, i, iterator_basic);
        basic_assign(temp_basic, iterator_basic);
        temp = Data_Make_Struct(rb_obj_class(self), basic_struct, NULL,
                                                    cbasic_free, temp_basic);
        rb_ary_push(ruby_array, temp);
    }
```

certik: Much better. Then there is another problem. `basic_init` initializes *inplace*. So this will cause a segfault. We need to extend our C interface to allow both static and dynamic allocation.

abinashmeher999: In every loop there is a new pointer being statically allocated, would they still cause segfault. When exactly can `basic_init` cause a segfault?
I tried this approach. I do not get a segfault but the same error that happened persists.

certik: I think you are overwriting unallocated memory, so the behavior is undefined (it only segfaults in the better case, in the worse case it silently fails and overrides some important data and the program will misbehave).

You need to use #514 to implement this.

certik: So this code looks valid to me:
```c++
    for(i = 0; i < size; i++) {
        basic_struct *temp_basic = basic_new_heap();
        vecbasic_get(args, i, temp_basic);
        temp = Data_Make_Struct(rb_obj_class(self), basic_struct, NULL,
                                                    cbasic_free_heap, temp_basic);
        rb_ary_push(ruby_array, temp);
    }
```
and you need to create a new function `cbasic_free_heap` that just calls `basic_free_heap()`.

abinashmeher999: Thanks a lot! Let me try this out.

isuruf:
@abinashmeher999, were you able to fix this error?

abinashmeher999:
@isuruf No I wasn't. I had shifted to Integer and Rational, since I know what to do there. Sorry for keeping this stagnant. Any ideas?

isuruf:
Yes, I took a look today and found the problem.
```
        basic_struct *temp_basic = basic_new_heap();
        vecbasic_get(args, i, temp_basic);
        temp = Data_Make_Struct(rb_obj_class(self), basic_struct, NULL,
                                                    cbasic_free_heap, temp_basic);
```

The problem here is that `Data_Make_Struct` allocates memory using Ruby's `ALLOC`. So what happens is you allocate memory in the first line, fill it in second line and then in the 3rd line allocated memory is overwritten.

Since we are allocating memory ourselves, use `Data_Wrap_Struct`. which does not do that.

Another problem I found in `cbasic_binary_op`
```
    result = Data_Make_Struct(rb_obj_class(self), basic_struct, NULL , cbasic_free, cresult);
    basic_init(cresult);
    cwfunc_ptr(cresult, this, cbasic_operand2);
```

In this way, first memory is allocated on the heap using ruby' `ALLOC` and then placement new is called (which is not needed). I suggest you use `basic_new_heap` here as well. Something like,

```
    basic_struct *cresult = basic_new_heap();
    cwfunc_ptr(cresult, this, cbasic_operand2);
    result = Data_Wrap_Struct(rb_obj_class(self), NULL , cbasic_free, cresult);
```
And this would keep everything uniform and we use only one allocation function `basic_new_heap` and deallocation function `basic_free_heap` throughout ruby wrappers.

abinashmeher999:
@isuruf Thanks a lot for looking into it :smiley: ! This makes a lot of things clear for me. I am trying this.

However, why were we allocating basic on the stack? Performance?
>```
    basic_struct *cresult = basic_new_heap();
    cwfunc_ptr(cresult, this, cbasic_operand2);
    result = Data_Wrap_Struct(rb_obj_class(self), NULL , cbasic_free, cresult);
```
And this would keep everything uniform and we use only one allocation function basic_new_heap and deallocation function basic_free_heap throughout ruby wrappers.

If performance is an issue, and the way you mentioned this works, replacing the first line with the placement new should still work.

**Comment by certik** [comment]
Yes, for performance we want to allow people to allocate on stack, just like GMP allows that. But for Ruby (and Python) applications, it makes sense to allocate on heap. That's why we implemented #514.

**Comment by certik** [comment]
Thanks @isuruf for helping out with this!

**Comment by abinashmeher999** [comment]
@isuruf Thanks again for making this work! :smile: @certik This is now ready for review.

File: symengine/ruby/ext/symengine/ruby_basic.c [hunk 9]
```diff
@@ -5,23 +5,25 @@ void cbasic_free(void *ptr){
...
+}
+
+VALUE alloc_func(VALUE klass, void(*free_func_ptr)(void *)) {
     VALUE obj;
     basic_struct *struct_ptr;
 
-    obj = Data_Make_Struct(klass, basic_struct, NULL, cbasic_free, struct_ptr);
-
-    struct_ptr->data = NULL;
+    obj = Data_Make_Struct(klass, basic_struct, NULL, free_func_ptr, struct_ptr);
```
isuruf: Use `basic_new_heap` to allocate and wrap using `Data_Wrap_Struct` as well.

symengine/ruby/ext/symengine/ruby_basic.c
```diff
@@ -5,23 +5,25 @@ void cbasic_free(void *ptr){
     basic_free(basic_ptr);
 }
 
-VALUE cbasic_alloc(VALUE klass){
+void cbasic_free_heap(void *ptr) {
+    basic_struct *basic_ptr = ptr;
+    basic_free_heap(basic_ptr);
+}
+
+VALUE alloc_func(VALUE klass, void(*free_func_ptr)(void *)) {
```
isuruf: We have only one method for allocation right? So, this can just be `cbasic_alloc`

symengine/ruby/ext/symengine/ruby_basic.c
abinashmeher999: Ok. I am doing it.

symengine/ruby/lib/symengine/basic.rb
```diff
@@ -0,0 +1,13 @@
+require 'set'
+
+module SymEngine
+  class Basic
+    def free_symbols
+      pr_free_symbols.to_set
+    end
+
+    def hash
+      [self.to_s].hash
```
isuruf: You should use the hash function of the Basic objects here

isuruf: We can do this another PR as well

symengine/ruby/ext/symengine/symengine.c
```diff
@@ -1,25 +1,39 @@
...
+    rb_define_method(c_basic, "!=", cbasic_neq, 1);
     rb_define_method(c_basic, "-@", cbasic_neg, 0);
+    rb_define_method(c_basic, "to_s", cbasic_to_str, 0);
+    rb_define_method(c_basic, "expand", cbasic_expand, 0);
+    rb_define_method(c_basic, "args", cbasic_get_args, 0);
+    rb_define_protected_method(c_basic, "pr_free_symbols", cbasic_free_symbols, 0);
+
+    //Symbol class
+    c_symbol = rb_define_class_under(m_symengine, "Symbol", c_basic);
+    rb_define_alloc_func(c_symbol, csymbol_alloc);
```
isuruf: You can use `cbasic_alloc` here instead of `csymbol_alloc` since both are the same. When all 50 classes in SymEngine have wrappers in Ruby, there are going to be 50 functions doing the same thing with this approach.

isuruf:
Left a few comments. Other than that +1 from me

symengine/ruby/lib/symengine/basic.rb
abinashmeher999: Ok. Any ideas for casting `size_t` to `long`? Implementation of `size_t` varies and `#hash` returns a `FIXNUM`, whose size is about the size of `long`, and also there is a macro for converting `long` to `FIXNUM`.

symengine/ruby/ext/symengine/symengine.c
abinashmeher999: I see. I will revert back to `cbasic_alloc` then.

abinashmeher999: Done

symengine/tests/cwrapper/test_cwrapper.c
certik: Did you forget to remove it?

symengine/ruby/lib/symengine/basic.rb
certik: Just write a C hash function that returns FIXNUM, and that checks that `sizeof(size_t) == sizeof(FIXNUM)`.

symengine/ruby/ext/symengine/ruby_basic.c
```diff
@@ -5,23 +5,17 @@ void cbasic_free(void *ptr){
...
 }
 
 VALUE cbasic_init(VALUE self){
-    basic_struct *this;
-
-    Data_Get_Struct(self, basic_struct, this);
-    basic_init(this);
-
     return self;
 }
```
certik: This function is empty now --- is that intended? Does it still need to be declared? It's only used in
```
     rb_define_method(c_basic, "initialize", cbasic_init, 0);
```

symengine/ruby/spec/basic_spec.rb
```diff
@@ -1,71 +1,171 @@
...
-                    end
-                end
-            end
+    describe '#expand' do
+      context 'with nothing as argument' do
+        it 'returns the expanded form' do
+          x = SymEngine::Symbol.new('x')
+          y = SymEngine::Symbol.new('y')
+          z = SymEngine::Symbol.new('z')
+          e = (x + y + z) * (x + y + z)
```
certik: Can you now write this as `(x + y + z) ** 2`?

symengine/ruby/spec/basic_spec.rb
```diff
@@ -1,71 +1,171 @@
...
+          f = e.expand
+          expect(e.to_s).to eql('(x + y + z)**2')
+          expect(f).to be_an_instance_of SymEngine::Symbol
+          expect(f.to_s).to eql('2*x*y + 2*x*z + 2*y*z + x**2 + y**2 + z**2')
```
certik: As we talked about, can you now compare this directly to `2*x*y + 2*x*z + 2*y*z + x**2 + y**2 + z**2`, constructed as such (i.e. no string)?

certik:
It starts to look pretty good. Left some more comments.

symengine/ruby/ext/symengine/ruby_basic.c
abinashmeher999: That was for the `new` method for instantiating a Basic object. I don't think it uses a default one, if we don't declare it. I will give it a try.

certik: Which ruby code is using this? I don't understand how the variable can get initialized (i.e. allocated on heap) if this function does nothing.

abinashmeher999: I was wrong. It does get a default `initialize` method. Allocation on the heap is done by the method `alloc`. In ruby, the creation of objects happen in two stages, first it gets allocated(`alloc`) and the initialized(`initialize`). We already have a alloc_func for basic, so this can be removed. Thanks for pointing it out.

symengine/ruby/spec/basic_spec.rb
abinashmeher999: Sorry, not yet. I will send a PR for it.

symengine/ruby/spec/basic_spec.rb
abinashmeher999: Same here, the integers is the only thing that is stopping this from happening. I am adding this as soon as I can.

abinashmeher999:
I have made the suggested changes. I haven't rebased the changes to clean up the PR now(like squashing and all), since review was going on. Do let me know if it is needed.

symengine/ruby/lib/symengine/basic.rb
isuruf: `ULL2NUM(unsigned long long) - take an unsigned long long int and convert it to a FIXNUM object if it will fit; otherwise, convert to a Bignum object`

isuruf: Or better yet, use `SIZET2NUM`

certik:
I am ok with this. @isuruf any more comments? Otherwise we can merge it.

isuruf:
+1 to merge

isuruf:
Travis passes. Shippable failure is due to an error in Shippable

PR #487, opened 2015-06-26
Support for Integer and Rational classes in Ruby wrappers

abinashmeher999:
TODO:
- [x] Create instance of `SymEngine::Integer`
- [x] Create instance of `SymEngine::Rational`
- [x] Identify Ruby Integer as SymEngine Integer when used with a Basic
- [x] Identify Ruby Rational as SymEngine Rational when used with a Basic


certik:
The Shippable builds fail due to #488.

certik:
I think this looks good.

certik:
2 of the 4 ruby tests fail with:
```
............../home/travis/.rvm/gems/ruby-1.9.3-p551/bin/rspec: symbol lookup error: /home/travis/build/sympy/symengine/symengine/ruby/lib/symengine/symengine.so: undefined symbol: rb_funcallv_public
```

abinashmeher999:
I had been following the guide for Ruby 2.1. I will have to look for other functions that can do the same. I am sure there is one for what I wanted to achieve with this function. This just calls the method if it is public, there is no need to mention that.

certik:
Or we can just support Ruby 2.1 and up.

abinashmeher999:
I found the function and it works. It is not a big issue. We can still support the older versions.

abinashmeher999:
@certik I have pushed some new changes that allow operations like `2*x` and `x*2`. I need to do the same for `Rational` and maybe then we can ask people to try it out.

certik:
Cool, good job!

symengine/ruby/ext/symengine/ruby_basic.c
```diff
@@ -16,15 +16,53 @@ VALUE cbasic_alloc(VALUE klass){
...
+            GET_SYMINTFROMVAL(den, den_basic);
+
+            Klass = rb_path2class("SymEngine::Rational");
+            new_operand2 = cbasic_alloc(Klass);
+            Data_Get_Struct(new_operand2, basic_struct, cbasic_operand2);
+            rational_set(cbasic_operand2, num_basic, den_basic);
+
+            basic_free_heap(num_basic);
+            basic_free_heap(den_basic);
+
```
isuruf: `break;` is needed

symengine/ruby/ext/symengine/symengine_macros.c
```diff
@@ -0,0 +1,14 @@
...
+VALUE Klass_of_Basic(const basic_struct *basic_ptr) {
+    switch(basic_get_type(basic_ptr)) {
+        case SYMENGINE_ADD:
+        case SYMENGINE_MUL:
+        case SYMENGINE_POW:
+            return rb_path2class("SymEngine::Symbol");
+        case SYMENGINE_INTEGER:
+            return rb_path2class("SymEngine::Integer");
+        case SYMENGINE_RATIONAL:
+            return rb_path2class("SymEngine::Rational");
```
isuruf: You could use `c_rational` here right?

symengine/ruby/ext/symengine/symengine_macros.h
```diff
@@ -0,0 +1,22 @@
...
+
+VALUE Klass_of_Basic(const basic_struct *basic_ptr);
+
+#define GET_SYMINTFROMVAL(num_value, this) \
+    switch(TYPE(num_value)) { \
+        case T_FIXNUM: \
+            integer_set_si(this, FIX2LONG(num_value)); \
+            break; \
+        case T_BIGNUM: \
+            integer_set_ui(this, NUM2ULONG(num_value)); \
```
isuruf: BigNum is bigger than a long most of the time. You should get the string representation and then construct a `SymEngine::Integer`

symengine/ruby/lib/symengine/integer.rb
```diff
@@ -0,0 +1,93 @@
+class Fixnum
```
isuruf: You shouldn't modify built-in classes of a programming language even if it is allowed.
Take a look at class coercion in Ruby.

symengine/ruby/ext/symengine/ruby_basic.c
```diff
@@ -16,15 +16,53 @@ VALUE cbasic_alloc(VALUE klass){
...
 VALUE cbasic_binary_op(VALUE self, VALUE operand2, void (*cwfunc_ptr)(basic_struct*, const basic_struct*, const basic_struct*)){
-    basic_struct *this, *cbasic_operand2, *cresult;
+    basic_struct *this, *cbasic_operand2, *cresult, *num_basic, *den_basic;
     VALUE result;
 
     Data_Get_Struct(self, basic_struct, this);
-    Data_Get_Struct(operand2, basic_struct, cbasic_operand2);
+    VALUE Klass, new_operand2, num, den;
+
+    switch(TYPE(operand2)) {
```
isuruf: You can take out the functionality of this block to a new method like `void sympify (basic new_operand2, VALUE operand2)`

symengine/ruby/ext/symengine/symengine_macros.c
abinashmeher999: I have never tried it. But `c_rational` was declared in `symengine.c`. Will it be able to link ?

symengine/ruby/ext/symengine/symengine_macros.h
abinashmeher999: OK. I will do that.

symengine/ruby/ext/symengine/ruby_basic.c
abinashmeher999: Yes. Thanks!

symengine/ruby/lib/symengine/integer.rb
abinashmeher999: I was looking for the exact same thing! Thanks! Now I think this part of code won't be needed.

symengine/ruby/ext/symengine/symengine_macros.c
isuruf: It doesn't matter to the compiler.
You can move the declarations to a `symengine.h`

abinashmeher999:
@certik @isuruf Sorry for the delay. I have made the suggested changes. I will add the tests soon. One more feature that can be added is that ability for the SymEngine classes to be coerced into other classes. I will better make a new PR for that after this is over.

symengine/ruby/lib/symengine/basic.rb
```diff
@@ -7,3 +7,13 @@ def free_symbols
     end
   end
 end
+
+class SymEngine::Basic
+  def coerce(other)
```
isuruf: You can define and implement this in C API and use the `sympify` function.

certik:
@abinashmeher999 what else needs to be done here? Tests seem to pass.

symengine/ruby/lib/symengine/basic.rb
abinashmeher999: Yes that can be done too.

abinashmeher999:
If you remember, I was also trying to make the rspec counterpart for `test_arit.py`. The tests have helped a lot. I found out that the `==` operator still needs to be integrated with Ruby classes. That won't take long. 
This and the test file `arit_spec.rb`. That will be sufficient, I guess, to suggest this to others for feedback.

certik:
That sounds good. Can you ping me once you have something for review?

abinashmeher999:
Sure, I will.

symengine/ruby/ext/symengine/ruby_basic.c
```diff
@@ -66,7 +66,7 @@ VALUE cbasic_eq(VALUE self, VALUE operand2) {
     basic_struct *this, *cbasic_operand2;
 
     Data_Get_Struct(self, basic_struct, this);
-    Data_Get_Struct(operand2, basic_struct, cbasic_operand2);
+    sympify(operand2, &cbasic_operand2);
```
abinashmeher999: I have been getting [this](https://gist.github.com/abinashmeher999/7f4935459b334b7f8353) error, after the introduction of this method. I don't understand what has changed, except that I have passed the pointer to `cbasic_operand2`. The same thing(`Data_Get_Struct`) that happened earlier, happens in [`sympify`](https://github.com/sympy/symengine/pull/487/files#diff-5f89ca9c732a55713f6682f4f7a61d5eR4)

abinashmeher999:
Two tests still fail
```ruby
......FF..............................

Failures:

  1) Arithmetic test cases test_arit7
     Failure/Error: assert 2 * x - x == x
     TypeError:
       wrong argument type Fixnum (expected Class)
     # ./spec/arit_spec.rb:80:in `-'
     # ./spec/arit_spec.rb:80:in `block (2 levels) in <top (required)>'

  2) Arithmetic test cases test_arit8
     Failure/Error: assert x**2 / x == x
     TypeError:
       wrong argument type Fixnum (expected Class)
     # ./spec/arit_spec.rb:94:in `/'
     # ./spec/arit_spec.rb:94:in `block (2 levels) in <top (required)>'

Finished in 0.01956 seconds (files took 0.10097 seconds to load)
38 examples, 2 failures

Failed examples:

rspec ./spec/arit_spec.rb:75 # Arithmetic test cases test_arit7
rspec ./spec/arit_spec.rb:86 # Arithmetic test cases test_arit8
```
The error messages show that the argument type to `-` and `/` are `Fixnum`. And the error messages don't even make sense. It can be clearly seen that the operand to `-` and `/` are `SymEngine::Symbol`s. I had even tried printing them out, just to be double sure. I am not able to figure out what might be the reason.
@isuruf Any ideas?

symengine/ruby/ext/symengine/ruby_basic.c
abinashmeher999: After the commit [`ae56d65`](https://github.com/abinashmeher999/symengine/commit/ae56d65e3238339c1c9153c67989435670eb1d49), the error message has changed. Now it says that something is being freed twice, while earlier it said that I was not freeing some.
```
*** Error in `ruby /home/abinashmeher999/.rvm/gems/ruby-2.2.0/bin/rspec spec': double free or corruption (fasttop): 0x00000000026ef3e0 ***
Aborted (core dumped)
```
What am I missing here?

symengine/ruby/ext/symengine/symengine_macros.c
```diff
@@ -0,0 +1,51 @@
...
+            num = rb_funcall(operand2, rb_intern("numerator"), 0, NULL);
+            den = rb_funcall(operand2, rb_intern("denominator"), 0, NULL);
+
+            num_basic = basic_new_heap();
+            den_basic = basic_new_heap();
+
+            GET_SYMINTFROMVAL(num, num_basic);
+            GET_SYMINTFROMVAL(den, den_basic);
+
+            new_operand2 = cbasic_alloc(c_rational);
```
certik: This leaks a memory, doesn't it? You created a new VALUE using `cbasic_alloc`, but you return it as our own data structure using `cbasic_operand2`, but I don't know if that is supported in Ruby C/API. I would expect that you have to return it as `VALUE`.

certik: So Ruby will never free `new_operand2`, and so it won't call our `cbasic_free_heap`, and thus it leaks.

certik: In all the other cases, Ruby is responsible to free the `VALUE operand2`, in other words, the caller should not free `cbasic_operand2`, is that your understanding? 

So in that case, the `T_RATIONAL` Ruby type cannot be handled like that, since you are creating a new type.

So what this function instead needs to do is to create a copy of the SymEngine object, so that the caller is always responsible for freeing `cbasic_operand2`. That way, you can create a new SymEngine object Rational and just return it.

Also, why not to use the low level SymEngine API, and just define it as:
```c++
void sympify(VALUE operand2, basic basic_operand2) {
...
    
            num = rb_funcall(operand2, rb_intern("numerator"), 0, NULL);
            den = rb_funcall(operand2, rb_intern("denominator"), 0, NULL);

            num_basic = basic_new_heap();
            den_basic = basic_new_heap();

            GET_SYMINTFROMVAL(num, num_basic);
            GET_SYMINTFROMVAL(den, den_basic); 

            rational_set(basic_operand2, num_basic, den_basic);

            basic_free_heap(num_basic);
            basic_free_heap(den_basic); 
}
```

You can actually simplify this and allocate `num_basic` and `den_basic` on stack, but that's minor.

When you call this, the caller is responsible to call either `basic_init` or `basic_new_heap` on the `basic_operand2` argument, and then it is responsible to free it properly.

Let me know what you think.

abinashmeher999: > So Ruby will never free new_operand2, and so it won't call our cbasic_free_heap, and thus it leaks.

What you said makes a lot of sense. I had kind of overlooked that fact. I was under the assumption that the objects created on the heap in C, were automatically put in the object space of Ruby. This still needs confirmation though.

abinashmeher999: Ruby is dynamically-typed and the binary functions need to do different things based on the type of the operand they receive. This led to the need for a function like `sympify`. So, that whatever the type of the operand it gets, it extracts the proper pointer from it, either by *unwrapping* or by creating a new Ruby-SymEngine object from the operand and then *unwrapping* it.

So, `sympify` should not need to return the object to ruby. However, we still need an appropriate C struct made from the information that we got from the second operand. And making that ourselves will add an extra overhead to delete that. And it can be tricky many times(like in case of `SymEngine::Integer` in ruby) when we need to make a new object of it and use that for operations. So, there is indeed a need to pass that newly made object to ruby, so that it is properly handled.

What you said about first allocating and then passing it to the function and then again deleting it after getting it back can be done too. Only that we will have to change that for all the functions, that uses sympify.

abinashmeher999: I just recalled a function `rb_gc_mark()`. That might be the one to tell the Ruby GC to remember that an object has been created. I will look into it. That will solve this problem easily.

abinashmeher999:
`#coerce` method is used to extend already existing classes to achieve a specific behaviour with user defined types, without having to modify the existing class. In this case, if the `Fixnum` sees it can't call `#+` with `SymEngine::Basic`, it will call `#coerce` method on the passed object(operand) so that now the operand decides how it should behave with the class.

A much better explaination is [here](https://www.mutuallyhuman.com/blog/2011/01/25/class-coercion-in-ruby/)

symengine/ruby/ext/symengine/symengine_macros.c
abinashmeher999: Thanks @certik !

abinashmeher999: Also I hadn't used `basic` instead of `basic_struct *` is, initially when I started writing the wrappers, I was using `Data_Make_Struct`which itself allocated space, and hence using `basic` with it caused errors. But now since we are not using `Data_Make_Struct` anymore, `basic` can be used again.

certik: If sympify creates a new object, then the caller is responsible to delete it, unless you can somehow tell Ruby to delete it.

abinashmeher999: @certik @isuruf I found [`rb_class_new_instance`](http://rxr.whitequark.org/mri/source/object.c#1799). This should work, because I saw some lines for GC when I browsed through the source. I will let you know.

isuruf: I agree with @certik.

symengine/ruby/ext/symengine/symengine_macros.c
```diff
@@ -0,0 +1,51 @@
...
+VALUE Klass_of_Basic(const basic_struct *basic_ptr) {
+    switch(basic_get_type(basic_ptr)) {
+        case SYMENGINE_ADD:
+        case SYMENGINE_MUL:
+        case SYMENGINE_POW:
+            return c_symbol;
+        case SYMENGINE_INTEGER:
+            return c_integer;
+        case SYMENGINE_RATIONAL:
+            return c_rational;
```
isuruf: Two tests fail because of this.
```C++
        case SYMENGINE_SYMBOL:
            return c_symbol;
        case SYMENGINE_INTEGER:
            return c_integer;
        case SYMENGINE_RATIONAL:
            return c_rational;
        default:
            return c_basic;
```
works for me

symengine/ruby/ext/symengine/symengine_macros.c
certik: Is sympify meant to return an object to SymEngine? If so, then you should just let the caller allocate and deallocate. Then you can always pass it back to Ruby, just like you pass SymEngine objects to Ruby now.

abinashmeher999: Ah! I get it now. Yes, `sympify` is meant to return a SymEngine object. I wonder why this didn't come to me earlier. We can always extract the information from Ruby objects and return a SymEngine object. And after `sympify` has returned, we can wrap it, if needed. Thanks @certik!

symengine/ruby/ext/symengine/symengine_macros.c
abinashmeher999: Thanks!! :smile: I am trying it.

symengine/ruby/ext/symengine/symengine_macros.c
certik: Exactly. The only thing you need to make sure is to do a copy of the RCP
(which increments refcount) from Ruby. That way when the Ruby object goes
out of scope, the underlying SymEngine object doesn't get deallocated if
you still use it elsewhere.

abinashmeher999:
@certik I just rebased the branch. Should I open a new PR?

abinashmeher999:
@certik @isuruf except rebase, this is ready for review.

certik:
Yes, either rebase or merge with master, so that tests can be run.

symengine/ruby/ext/symengine/ruby_basic.c
```diff
@@ -19,12 +19,15 @@ VALUE cbasic_binary_op(VALUE self, VALUE operand2, void (*cwfunc_ptr)(basic_stru
...
+
     Data_Get_Struct(self, basic_struct, this);
-    Data_Get_Struct(operand2, basic_struct, cbasic_operand2);
+    sympify(operand2, cbasic_operand2);
 
     cresult = basic_new_heap();
     cwfunc_ptr(cresult, this, cbasic_operand2);
-    result = Data_Wrap_Struct(rb_obj_class(self), NULL , cbasic_free_heap, cresult);
+    result = Data_Wrap_Struct(Klass_of_Basic(cresult), NULL , cbasic_free_heap, cresult);
+    basic_free_heap(cbasic_operand2);
```
certik: I am worried that you are double freeing it here, aren't you? You free the `cbasic_operand2` using `basic_free_heap()` here, and then Ruby will double free it using `cbasic_free_heap`. Or am I misunderstanding it?

isuruf: I don't think `cbasic_operand2` is freed by Ruby. It is only a temporary used here. Only `cresult` is freed by Ruby.

certik: But isn't `cresult` a dangling pointer after you free `cbasic_operand2`?

isuruf: Nope. Since `cbasic_operand2` has refcount 2 or more, freeing `cbasic_operand2` would only decrease the refcount by 1

certik: I see, it's not --- since `cresult` is also allocated, and then the `cwfunc_ptr` is just calling e.g. `basic_add`. So I think all is good here.

certik:
I think the code looks ok to me, thanks @isuruf for clarifying the issue above.

@abinashmeher999 can you merge with the latest master? Let's see if tests run.

abinashmeher999:
@certik Tests pass on travis.

certik:
AppVeyor passes too. That one failure is caused by #577.

symengine/ruby/callgrind_test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+valgrind --tool=callgrind --dump-instr=yes --simulate-cache=yes --collect-jumps=yes ruby -S rspec spec
```
certik: This file should not be here. You can put it into `bin` directory (in the root).

symengine/ruby/valgrind_test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+valgrind --tool=memcheck --num-callers=15 --partial-loads-ok=yes --undef-value-errors=no ruby -S rspec spec
```
certik: This file should not be here. You can put it into bin directory (in the root).

symengine/ruby/ext/symengine/symengine_macros.c
```diff
@@ -0,0 +1,51 @@
...
+    }
+}
+
+VALUE Klass_of_Basic(const basic_struct *basic_ptr) {
+    switch(basic_get_type(basic_ptr)) {
+        case SYMENGINE_ADD:
+        case SYMENGINE_MUL:
+        case SYMENGINE_POW:
+        case SYMENGINE_SYMBOL:
+            return c_symbol;
```
certik: My understanding of the switch statement is that `SYMENGINE_ADD` (as well as Mul, and Pow) will return `c_symbol`, is that right? That seems wrong.

certik:
It starts to look pretty good. I left few more comments.

symengine/ruby/valgrind_test.sh
certik: Also document how it should be used and what the difference is from the other valgrind script.

symengine/ruby/callgrind_test.sh
abinashmeher999: This was a substitute to automate tasks by `rake`. Shifting it to `bin` at the root will require writing a path name everytime I want to execute it. How about we shift it to the root of the gem dir. I am trying to make it run from other place, a `LoadError` shows up everytime. Or we can bring back rake, I know how to do it with `rake`.

symengine/ruby/ext/symengine/symengine_macros.c
abinashmeher999: This was a workaround that seemed to make the wrappers work till `SYMENGINE_ADD` is wrapped. I should better search for a test case that fails with this, and then try to fix it. Any suggestions?

symengine/ruby/valgrind_test.sh
abinashmeher999: OK

symengine/ruby/ext/symengine/symengine_macros.c
certik: Just remove the first three cases for now. It will return the default case.
Then if needed, more cases can be implemented. How is this function
supposed to be used? Can you add some documentation  (into the header file)?

symengine/ruby/callgrind_test.sh
certik: Can you document how it is used? I am sure there is a solution.

symengine/ruby/valgrind_test.sh
certik: Cannot you run this as:
```
../../bin/valgrind_test.sh
```
?

certik: The same comment about the other script.

Also, you still need to document how this script is meant to be used.

certik:
See my comments above.

symengine/ruby/valgrind_test.sh
abinashmeher999: Oh that worked! So for making this work, I need to be in the root dir of the ruby wrappers. Since, this script will only be used in the ruby wrappers, shall I move this to the `bin` dir in the root dir of the ruby wrappers rather than the root dir of the symengine project?

certik: Ok, you can do that. I think that's ok.

abinashmeher999: @certik Done.

certik:
Thanks. That looks good to me. +1 to merge as it is.

Next time, don't use the `[ci skip]` thing, as even moving a file can make some tests fail, since we install things, and it might break. In this case hopefully it still works.

abinashmeher999:
I see. I will keep that in mind. @isuruf any comments?

isuruf:
I'm +1 to merge.

In another PR, you could,
1. Provide a ruby interface to the sympify function
2. Use `basic_new_stack` for temporaries

certik:
Ah yes, I forgot to suggest that to use `basic_new_stack` whenever you can, since those functions completely avoid the heap (`new`/`delete`) so they are much faster.

PR #489, opened 2015-06-27
Added a beginner IRuby notebook

abinashmeher999:
- Added an IRuby notebook for beginners
- Added a README redirecting to instructions on setting up IRuby
- Added `iruby` as a development dependency

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,4 @@ symengine/ruby/pkg/
 symengine/ruby/ext/symengine/mkmf.log
 symengine/ruby/bin/*
 symengine/ruby/ext/symengine/extconf.h
+symengine/ruby/notebooks/.ipynb_checkpoints/
--- a/symengine/ruby/lib/symengine.rb
+++ b/symengine/ruby/lib/symengine.rb
@@ -1,4 +1,5 @@
 require 'symengine/symengine'
+require 'symengine/iruby'
 
 module SymEngine
   def SymEngine.ascii_art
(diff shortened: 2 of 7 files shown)


abinashmeher999:
@dilcom Could you please review this? I would like to have your suggestions. :blush:

dilcom:
Sure =)

It doesn't seem good to call ```#to_s``` each time to output value to iRuby. You can define ```#to_iruby``` method as alias of ```#to_s``` and iRuby should call it. You may also take a look how iRuby supported in [GnuplotRB](https://github.com/dilcom/gnuplotrb/blob/master/lib/gnuplotrb/mixins/iruby.rb) and Daniel's [advices](https://github.com/SciRuby/iruby/pull/37).

I'm also not sure if you should add iRuby as dependency.

Apart of this I think notebook is ok since only Symbols work.

abinashmeher999:
Thanks for the suggestions! :smile:  I looked at your repo. It's a great idea to have a separate `Gemfile` for the notebooks. It's to isolate the dependencies of notebooks from the dependencies of the gem, right? 

Also, `iruby` requires ruby version 2.1 or higher. So, the build breaks for versions lower than that. It should not be included as a dependency in a sense that it is not essential for development. Better to adopt your method. I will try out the `#to_iruby` approach, and let you know how that goes.

dilcom:
Since you don't test your gem with iruby, there is no need to add it to dependencies.

Additional gemfile is needed because otherwise ruby will use main gemfile. There is no iruby in it so ruby will be unable to find it and notebook will not work.

abinashmeher999:
@dilcom Thanks for the review and the suggestion! :smile: . @certik The PR is ready. :sparkles:

abinashmeher999:
@certik This PR can be merged too.

certik:
I think that works. Thanks!

PR #495, opened 2015-06-30
Wrap vec_basic and get_args() into C

certik:
This uses a clean approach to wrap C++ classes into C. No casting is done in this new code (except for the already implemented `basic` type itself, which we need to refactor to use this new approach later), so both the C and C++ compilers know all they need to know about the pointers:

* From the C perspective, each type is an opaque structure (but it's not a void pointer, nor any other casts are necessary), so the C compiler can check the type
* From the C++ perspective, the opaque type from the .h file is implemented in the .cpp file as a simple C structure which has the C++ data member `m` in it.

This should have no runtime overhead. You access the `m` member directly.

The ugly function that @abinashmeher999 and I were designing here: https://github.com/abinashmeher999/symengine/commit/b32926674800476176d23200439c7e66c9324b26#commitcomment-11936123:
```c++
void basic_get_args(const basic a, basic_struct **array, int *size)
{
    SymEngine::vec_basic args = (*RCP_const_cast(a))->get_args();
    *size = args.size();
    *array = (basic_struct*)malloc((*size)*sizeof(basic_struct));
    int index = 0;
    for (auto &a: args) {
        *RCP_cast(&((*array)[index++])) = a;
    }
}
```
now becomes a simple two liner:
```c++
void basic_get_args(const basic self, CVecBasic *args)
{
    for (auto &a: (*RCP_const_cast(self))->get_args())
        args->m.push_back(a);
}
```
The only ugly cast comes from the `basic` type. After we upgrade the rest of the C wrappers, this function will look like this:
```c++
void basic_get_args(const CRCPBasic *self, CVecBasic *args)
{
    for (auto &a: (self->m->get_args())
        args->m.push_back(a);
}
```
Where `CRCPBasic` is defined as:
```c++
struct CVecBasic {
    SymEngine::RCP<const SymEngine::Basic> m;
};
```

So everything becomes clean, no casts and as safe as it can be in C. Writing the C wrappers then just becomes writing a wrapper C structure with the C++ `m` member in it, and accessing and manipulating `m` directly using C++. The only potentially unsafe operation is dereferencing the function arguments (e.g. `const CRCPBasic *self`) to access the `m` member. Manipulating the `m` member is then done in C++ and thus is 100% safe in Debug mode. The only way this can fail is if the user doesn't allocate the memory (e.g. for `self`), but there is no way around that in C.

This PR also shows how to allocate directly on stack using placement new (see `test_CVectorInt2`):
```c++
    char data3[50]; // Aligned properly and enough size to fit std::vector<int>
    CVectorInt *vec = (CVectorInt*)data3;
    assert(vectorint_placement_new(vec, sizeof(data3)) == 0);
```
And that's how we need to allocate the `CRCPBasic` type. The approach of `vectorint_placement_new` is very general, since the C wrappers do not know its size, but for `CRCPBasic` we know its size, so we'll use the same trick that we already implemented in `basic_init`.

/cc @isuruf, @thilinarmtb


certik:
Actually, I just simplified the `basic_get_args` function even further:
```c++
void basic_get_args(const basic self, CVecBasic *args)
{
    args->m = (*RCP_const_cast(self))->get_args();
}
```
Which will eventually become:
```c++
void basic_get_args(const CRCPBasic *self, CVecBasic *args)
{
    args->m = self->m->get_args();
}
```
So I think this is the way to go. Very simple.

isuruf:
>The approach of vectorint_placement_new is very general, since the C wrappers do not know its size, but for CRCPBasic we know its size, so we'll use the same trick that we already implemented in basic_init.

How about using CMake to discover the size?

abinashmeher999:
Wow! :flushed: http://yosefk.com/c++fqa/fqa.html#fqa-32.8 this turned out to be really elegant. This seems as clean as it can get. We should refactor the `basic_struct` too as soon as possible. It will accelerate writing the C wrappers. Shall I send a PR to this branch?
And here
> Where `CRCPBasic` is defined as: 
```cpp
struct CVecBasic {
    SymEngine::RCP<const SymEngine::Basic> m;
};
```

Maybe you meant `CRCPBasic` in the code.

symengine/cwrapper.cpp
```diff
@@ -200,4 +210,78 @@ int is_a_Symbol(const basic c)
...
+    if (!SymEngine::is_aligned(self)) return 2;
+    if (size < sizeof(CVectorInt)) return 1;
+    new(self) CVectorInt;
+    return 0;
+}
+
+void vectorint_free(CVectorInt *self)
+{
+    delete self;
+}
```
abinashmeher999: Let's say we have more memory than `sizeof(CVectorInt)` allocated on the stack, and we use `vector_placement_new` to initialize `CVectorInt` at the given location. So, when we call delete on the pointer to that location, will the whole memory(that is more than the size of CVectorInt) be freed? If I am guessing correctly, does `SymEngine::is_aligned` has anything to do to ensure that?

certik:
Yes, that was a typo. Yes, please send a PR against this branch. Once we
merge this, you can then easily rebase against master.

symengine/cwrapper.cpp
certik: You should not call delete on it. The caller provided the memory, the
caller is responsible to free it. However, you are right that we need to
call the destructor ourselves, i.e. provide a function to do that. Perhaps
call it vectorint_placement_free or something like that, and it would just
call the destructor.

symengine/cwrapper.h
```diff
@@ -100,6 +101,37 @@ int is_a_Rational(const basic s);
...
+//! Wrapper for std::vector<int>
+
+typedef struct CVectorInt CVectorInt;
+
+CVectorInt* vectorint_new();
+
+// 'self' must point to allocated memory of size 'size'. The function returns 0
+// if std::vector<int> can be initialized using placement new into 'self',
+// otherwise 1 if 'size' is too small or 2 if 'self' is not properly aligned.
+// No memory is leaked either way.
```
abinashmeher999: Can we `enum` our error_types, like it is mentioned [here](http://stackoverflow.com/questions/1102542/how-to-define-an-enumerated-type-enum-in-c#1102556)? However, there is this issue that I got to know from one of the comments.
> @InnaKamoze isn't enum defined in C89? So unless you're not using ANSI C, I think that enum is supported.

certik:
For basic? Yeah, we might need to do that actually. For other things I
don't know yet. Maybe. It should be pretty simple, just printing sizeof and
capturing by cmake and provide these constants in the C header file for
users to use. Then asserting at compile time in the C++ code that the size
is correct. You still have to check alignment at runtime.

Alternatively, why not to provide a C function that returns the correct
size? That can still be used in the user C declarations I think.

symengine/cwrapper.h
certik: Yes, definitely. We can use C macros, those are portable.

isuruf:
Yes, we should define a C function that returns the correct size.
To discover the size of other things, there is a function we can use in CMake, http://www.cmake.org/cmake/help/v2.8.6/cmake.html#module:CheckTypeSize

certik:
Actually, let's use cmake to find the size and then declare the right sized
and aligned C structs and the array [1] trick for allocating on stack in
our C headers. Check size and alignment of these structs at compile time
(since the alignment won't change at runtime, as long as the same compiler
and platform is used, but we can add checks at runtime too if the user is
using a different compiler). The user can the use sizeof() on these structs
to determine the right size if he wants to allocate memory himself. But
normally the user would just use our predefined structs.

What do you think?

isuruf:
Yes, we should do that.
We should also provide a function to get the size without knowing about the struct as well. (In Julia, we can use this function to get the size, define the struct using that and call it)

certik:
Sure, or you can just call sizeof() on the struct, or provide a function
that does that.

abinashmeher999:
Should I make a new struct for `const Symengine::RCP<const Basic>` like `Cc_RCPBasic` or make the pointer point to a `const CRCPBasic`, to substitute the need to pass `const basic s` and use `RCP_const_cast`?

abinashmeher999:
>```cpp
void basic_get_args(const CRCPBasic *self, CVecBasic *args)
{
    args->m = self->m->get_args();
}
```

I found the answer in this example.

**Comment by certik** [comment]
Yes, I think I even posted how the struct should look like.

**Comment by certik** [comment]
I merged with master. See if it fixes the tests failures. They pass for me locally.

**Comment by certik** [comment]
Tests still fail. It looks like Debug mode works, and Release mode segfaults. I need to see what's going on, it might take me a day or two.

**Comment by certik** [comment]
Here is the stacktrace in release mode of the segfault:
```
#0  0x000000000043e2e7 in construct<int, int const&> (
    __p=0x4c8a0d <__libc_csu_init+77>, this=0x7fffffffde80)
    at /usr/include/c++/4.8/ext/new_allocator.h:120
No locals.
#1  _S_construct<int, int const&> (__p=0x4c8a0d <__libc_csu_init+77>, __a=...)
    at /usr/include/c++/4.8/bits/alloc_traits.h:254
No locals.
#2  construct<int, int const&> (__p=0x4c8a0d <__libc_csu_init+77>, __a=...)
    at /usr/include/c++/4.8/bits/alloc_traits.h:393
No locals.
#3  push_back (__x=@0x7fffffffde6c: 5, this=0x7fffffffde80)
    at /usr/include/c++/4.8/bits/stl_vector.h:905
No locals.
#4  vectorint_push_back (self=0x7fffffffde80, value=5)
    at /home/ondrej/repos/symengine/symengine/cwrapper.cpp:236
No locals.
#5  0x0000000000437c96 in main ()
No symbol table info available.
```

**Comment by certik** [comment]
Found it. It's the NDEBUG thing biting us again not executing the assert statements in the C test file. Solution is to turn it off or (better) write our own macros for assert in C.

**Comment by Sumith1896** [comment]
@certik Is `DNDEBUG` passed anywhere? 
Where is the `NDEBUG` macro declared, that's giving the problem?

File: symengine/tests/cwrapper/test_cwrapper.c [hunk 3]
```diff
@@ -63,8 +64,96 @@ void test_cwrapper() {
...
+
+    char data1[1];  // Not aligned properly
+    vec = (CVectorInt*)data1;
+    assert(vectorint_placement_new(vec, sizeof(data1)) == 2);
+
+    struct X data2[1];  // Aligned properly but small
+    vec = (CVectorInt*)data2;
+    assert(vectorint_placement_new(vec, sizeof(data2)) == 1);
+
+    char data3[50]; // Aligned properly and enough size to fit std::vector<int>
```
isuruf: Is this aligned properly?

isuruf:
@Sumith1896, In CMake, we haven't changed the default C flags, which is `-O3 -DNDEBUG` when built in release mode

Sumith1896:
Ohh cool, the same construct should do then, if we want a custom assert.

symengine/tests/cwrapper/test_cwrapper.c
certik: We check it at runtime. It seems it is aligned properly in Debug mode, but
not in Release mode (after I enabled the assert statement). I need to look
into it more.

certik:
Except that we need to use printf instead of std::cout.

symengine/tests/cwrapper/test_cwrapper.c
isuruf: An array is only required to have the alignment of a member, which is 1 byte in this case (alignment of char). For example,
```
    char d[50];
    std::cout << alignof(decltype(d)) << std::endl;
```
gives me 1

certik: That is correct. It's weird that it passes in Debug mode though...

certik:
I think I fixed it. I pushed in some unpolished patches, I will tidy it up later. Let's see if tests pass now.

I'll have to rework how we test for misaligned pointers, because I was testing for them to be misaligned, but sometimes (in Release mode) they happened to be aligned properly.

certik:
@abinashmeher999 I just pushed in patches that make it work for all of the C wrappers. I need to polish the commits and create nice few patches, but the changes are pretty much done, except some minor things.

I think the C interface pretty much didn't change, it's just implemented a lot more robustly.

symengine/cwrapper.cpp
certik: vectorint_placement_free() is now implemented. I checked with Valgrind that there are no memory leaks.

certik:
@isuruf do you think the way we use `basic` in this PR is ok?

Unfortunately I still have to use ifdef to have a different `struct` in C and C++, which is considered "evil" in one of the FAQs. There is a way to do it without it, similarly as I do it for the `vectorint_placement_new`, but then it is not as easy to use for the user. So it seems to me we'll use this "evil" approach for basic, but I would just use the heap based non-evil approach for the rest like std::vector and other types if needed.

certik:
Everything passes, except the Ruby wrappers. Closing in favor of #506.

PR #498, opened 2015-07-01
Make SYMENGINE_ASSERT use custom assert instead of assert()

Sumith1896:
En route to fixing https://github.com/sympy/symengine/issues/482
/cc @certik @isuruf

### DIFF
--- a/symengine/symengine_assert.h
+++ b/symengine/symengine_assert.h
@@ -1,10 +1,21 @@
 #ifndef SYMENGINE_ASSERT_H
 #define SYMENGINE_ASSERT_H
 
-
+// SYMENGINE_ASSERT uses internal functions to perform as assert
+// so that there is no effect with NDEBUG
 #if !defined(SYMENGINE_ASSERT)
 #if defined(WITH_SYMENGINE_ASSERT)
-#define SYMENGINE_ASSERT(cond) assert(cond);
+#define stringize(s) #s
+#define XSTR(s) stringize(s)
+#define SYMENGINE_ASSERT(cond) \
+{ \
+if (0 == (cond)) { \
+std::cerr << "SYMENGINE_ASSERT failed: " << __FILE__ << \
+	"\nfunction " << __func__ <<  "(), line number " << __LINE__ << \
+	" at \n" << XSTR(cond) << "\n"; \
+abort(); \
+} \
+}
 #else
 #define SYMENGINE_ASSERT(cond)
 #endif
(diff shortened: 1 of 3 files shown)


certik:
Why do you use do/while? Wouldn't a simple block suffice?

Sumith1896:
@certik 
If I remove the `do-while`, just `if ` remains
Say I do
```
if(X)
SYMENGINE_ASSERT(Y)
else
something_else;
```
The `else` is matched to the `if` of `SYMENGINE_ASSERT`.
This practice I noted in places, here  bytes.com/topic/c/answers/215019-use-do-while-0-assert-macro not something I knew of beforehand.
Let me know what you think.

certik:
Cannot you use the {} block? Then the else cannot be match to an if inside
the block.

Sumith1896:
Oh I just noticed that we don't put semi-colon after usage of macros like `SYMENGINE_ASSERT` etc.
By standard practice, people usually put `;` after, which means
```
#define FOO(x) do { foo(x); bar(x); } while (0)
```
serves the syntactic purpose while just braces `{}` don't.
I think for us braces would do the job.
Changing now.

Sumith1896:
@certik Done.

certik:
Ah I see what you mean. Either is fine with me. Just using {} seems simpler
to me.

Sumith1896:
Travis passes.

symengine/symengine_assert.h
```diff
@@ -1,10 +1,21 @@
...
+// SYMENGINE_ASSERT uses internal functions to perform as assert
+// so that there is no effect with NDEBUG
 #if !defined(SYMENGINE_ASSERT)
 #if defined(WITH_SYMENGINE_ASSERT)
-#define SYMENGINE_ASSERT(cond) assert(cond);
+#define stringize(s) #s
+#define XSTR(s) stringize(s)
+#define SYMENGINE_ASSERT(cond) \
+{ \
+if (0 == (cond)) { \
```
certik: Is there a reason why this cannot be written as `if (!(cond)) { \` ? Not a big deal, just asking.

Sumith1896: The standard `assert()` takes in int: http://en.cppreference.com/w/cpp/error/assert
ans hence this practice is adopted and used in the internal `assert()` too. I am not finding the exact place where I read this.

certik: I see. That's fine.

certik:
The shippable bugs seem internal to Shippable. Merging.

Sumith1896:
Thanks for merging!

PR #500, opened 2015-07-01
Expression wrapper.

Sumith1896:
Starting off from @bluescarni 's works.


symengine/expression.h
```diff
@@ -0,0 +1,108 @@
...
+#include <symengine/symengine_rcp.h>
+#include <symengine/integer.h>
+#include <symengine/mul.h>
+#include <symengine/pow.h>
+#include <symengine/symbol.h>
+
+namespace SymEngine
+{
+
+class expression
```
abinashmeher999: Shouldn't class name start with a capital letter?

Sumith1896: Just started form Francesco's work. Will take care of this.

symengine/tests/expression/test_expression.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+TEST_CASE("Arithmetic of expression", "[expression]")
+{
+	expression x = new Symbol("x"), y = new Symbol("y");
+	auto z = x + y;
+	std::cout << z << '\n';
+	z += y;
+	std::cout << z << '\n';
+	assert(z == x + y + y);
+	assert(z == x + 2 * y);
```
certik: you need to use REQUIRED, not assert.

certik: That should also fix the test failures.

symengine/expression.h
```diff
@@ -0,0 +1,108 @@
...
+			}
+			return *this;
+		}
+		~expression() noexcept {}
+		friend std::ostream &operator<<(std::ostream &os, const expression &expr)
+		{
+			os << (*expr.m_basic.get());
+			return os;
+		}
+		friend expression operator+(const expression &a, const expression &b)
```
Sumith1896: @certik We never overload operators, do we?
Or is it allowed for this class?

symengine/tests/expression/test_expression.cpp
Sumith1896: Oh hadn't noticed, will update.

symengine/expression.h
certik: It's allowed for this class, since it is a value class (behaving like a
value).

symengine/tests/expression/test_expression.cpp
```diff
@@ -0,0 +1,40 @@
...
+using namespace SymEngine;
+
+TEST_CASE("Constructors of expression", "[expression]")
+{
+	expression e0 = new Symbol("x");
+}
+
+TEST_CASE("Printing of expression", "[expression]")
+{
+	expression e0 = new Symbol("x");
```
Sumith1896: Is using `new` a problem?

certik: Yes. We need to probably make the expression accept rcp<basic>, as usual.

certik: Just pass it a symbol("x"), it looks like it will just work.

certik:
I fixed it myself in https://github.com/Sumith1896/csympy/pull/1.

Sumith1896:
@certik I thought it would be better if I could move some things to cpp but there most of them cannot be moved.
This is ready for review
/cc @certik @bluescarni @abinashmeher999

symengine/expression.cpp
```diff
@@ -0,0 +1,8 @@
+#include <symengine/expression.h>
+
+namespace SymEngine
+{
+
+
+
+} // SymEngine
```
certik: Get rid of this cpp file, it's not needed.

symengine/expression.h
```diff
@@ -0,0 +1,128 @@
...
+#include <symengine/pow.h>
+#include <symengine/symbol.h>
+
+namespace SymEngine
+{
+
+class Expression
+{
+private:
+	RCP<const Basic> m_basic;
```
certik: You should use 4 spaces, not 8 spaces, this applies to the rest of the file.

certik:
Otherwise I think this looks good.

symengine/expression.h
Sumith1896: Will do. Tab on my editor always comes here as 8 spaces :smile:

Sumith1896:
Done

symengine/tests/expression/test_expression.cpp
```diff
@@ -0,0 +1,40 @@
...
+TEST_CASE("Arithmetic of Expression", "[Expression]")
+{
+	Expression x = symbol("x"), y = symbol("y");
+	auto z = x + y;
+	std::cout << z << '\n';
+	z += y;
+	std::cout << z << '\n';
+	REQUIRE(z == x + y + y);
+	REQUIRE(z == x + 2 * y);
+	std::cout << pow(z,z) << '\n';
```
isuruf: Space after `,`

symengine/tests/expression/test_expression.cpp
```diff
@@ -0,0 +1,40 @@
...
+{
+	Expression e0 = symbol("x");
+	std::cout << e0 << '\n';
+}
+
+TEST_CASE("Arithmetic of Expression", "[Expression]")
+{
+	Expression x = symbol("x"), y = symbol("y");
+	auto z = x + y;
+	std::cout << z << '\n';
```
isuruf: `std::endl` instead of `\n`

symengine/expression.h
```diff
@@ -0,0 +1,128 @@
...
+class Expression
+{
+private:
+    RCP<const Basic> m_basic;
+
+public:
+    //! Plain constructor of Expression
+    Expression():m_basic(integer(0)) {}
+    //! Construct Expression from `int`
+    Expression(int n):m_basic(integer(n)) {}
```
isuruf: Space before and after `:`

symengine/tests/expression/test_expression.cpp
```diff
@@ -0,0 +1,40 @@
+#include "catch.hpp"
+#include <chrono>
+#include <iostream>
+
+#include <symengine/symengine_rcp.h>
+#include <symengine/expression.h>
+#include <symengine/pow.h>
+#include <symengine/symbol.h>
+
+using namespace SymEngine;
```
isuruf: Use explicit imports like `using SymEngine::Expression`

symengine/expression.h
```diff
@@ -0,0 +1,128 @@
...
+    {
+        m_basic = sub(m_basic,other.m_basic);
+        return *this;
+    }
+    //! Overload multiplication
+    friend Expression operator*(const Expression &a, const Expression &b)
+    {
+        return Expression(mul(a.m_basic,b.m_basic));
+    }
+    //! Overload multiplication and assignment (-=)
```
isuruf: `*=` instead of `-=`

symengine/expression.h
certik: If you use Vim, you should configure it to do 4 spaces (or 2 spaces in some
projects). Here is my config:

https://github.com/certik/dotfiles/blob/bf5d3f006782000441d46d818dd24ada3e4094b1/.vimrc

and you can drop in customized .project.vim into projects that need
different amount of space than 4.

Sumith1896:
Thanks @isuruf . Things are a lot cleaner now.

symengine/expression.h
```diff
@@ -0,0 +1,128 @@
...
+    //! Overload multiplication and assignment (*=)
+    Expression &operator*=(const Expression &other)
+    {
+        m_basic = mul(m_basic,other.m_basic);
+        return *this;
+    }
+    //! Overload check equality (==)
+    bool operator==(const Expression &other) const
+    {
+        return eq(m_basic,other.m_basic);
```
isuruf: You missed some `,` with no space after in this file. Otherwise +1.

Sumith1896:
@certik Travis passes.
Ready for final review.

certik:
Shippable errors are internal (it fails to download some stuff using apt-get).

I think this looks good, merging.

Sumith1896:
@certik Sorry for the delays.
Thanks for merging.

PR #502, opened 2015-07-03
Use mpzaddmul for poly_mul

Sumith1896:
Nice speedup obtained using FMA in `poly_mul`.
`expand2b` average falls from `95ms` to `60ms` on my machine.
As suggested by @bluescarni

PS: There may be other areas throughout SymEngine where we can use this, something to keep in mind.


certik:
Cool!

Sumith1896:
Thanks @certik

PR #506, opened 2015-07-05
C Wrappers Refactoring

certik:
These are polished patches from #495. Ready for review.

### DIFF
--- a/symengine/cwrapper.h
+++ b/symengine/cwrapper.h
@@ -7,38 +7,47 @@
 extern "C" {
 #endif
 
-// The size of 'basic_struct' must be the same as RCP<const Basic> *and* they
-// must have the same alignment (because we allocate RCP<const Basic> into the
-// memory occupied by this struct in cwrapper.cpp). We cannot use C++ in this
-// file, so we need to use C tools to arrive at the correct size and alignment.
-// The size of the RCP object on most platforms (with WITH_SYMENGINE_RCP on)
-// should be just the size of the 'T *ptr_' pointer that it contains (as there
-// is no virtual function table) and the alignment should also be of a pointer.
-// So we just put 'void *data' as the only member of the struct, that should
-// have the correct size and alignment. With WITH_SYMENGINE_RCP off (i.e. using
-// Teuchos::RCP), we have to add additional members into the structure.
+
+// The size of 'CRCPBasic_C' must be the same as CRCPBasic (which contains a
+// single RCP<const Basic> member) *and* they must have the same alignment
+// (because we allocate CRCPBasic into the memory occupied by this struct in
+// cwrapper.cpp). We cannot use C++ in this file, so we need to use C tools to
+// arrive at the correct size and alignment.  The size of the RCP object on
+// most platforms (with WITH_SYMENGINE_RCP on) should be just the size of the
+// 'T *ptr_' pointer that it contains (as there is no virtual function table)
+// and the alignment should also be of a pointer.  So we just put 'void *data'
+// as the only member of the struct, that should have the correct size and
+// alignment. With WITH_SYMENGINE_RCP off (i.e. using Teuchos::RCP), we have to
+// add additional members into the structure.
 //
 // However, this is checked at compile time in cwrapper.cpp, so if the size or
 // alignment is different on some platform, the compilation will fail --- in
 // that case one needs to modify the contents of this struct to adjust its size
 // and/or alignment.
-typedef struct
+struct CRCPBasic_C
 {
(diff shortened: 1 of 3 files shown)


symengine/cwrapper.h
```diff
@@ -1,44 +1,64 @@
...
+//  Assignment should be done only by using basic_assign(). Before the variable
+//  goes out of scope, basic_free() must be called.
+//
+//  For C, define a dummy struct with the right size, so that it can be
+//  allocated on the stack. For C++, the CRCPBasic is declared in cwrapper.cpp.
+#ifdef __cplusplus
+typedef struct CRCPBasic basic[1];
+#else
+typedef struct CRCPBasic_C basic[1];
+#endif
```
isuruf: We could have something like, 
`typedef struct CRCPBasic basic_struct` and `typedef struct CRCPBasic_C basic_struct`
and then define `typedef basic_struct basic[1]`
so that Ruby and Julia wrappers would work without any problem.

certik: Good point. I'll do it.

symengine/cwrapper.h
```diff
@@ -107,6 +127,42 @@ int is_a_Rational(const basic s);
...
+int vectorint_get(CVectorInt *self, int n);
+
+//! Wrapper for vec_basic
+
+typedef struct CVecBasic CVecBasic;
+
+CVecBasic* vecbasic_new();
+void vecbasic_free(CVecBasic *self);
+void vecbasic_push_back(CVecBasic *self, const basic value);
+void vecbasic_get(CVecBasic *self, int n, basic result);
```
isuruf: Since result is an output, it should come first.
`void vecbasic_get(basic result, CVecBasic *self, int n)`

certik: Well, the other argument is that the `self` variable should always be first, since these are just methods in the given object. Is there some guideline or some other project which has return arguments first?

symengine/cwrapper.h
certik: Fixed in c2bd01a5b090eaf15ccda7477d4321a7e2d190f7.

certik:
As of c2bd01a5b090eaf15ccda7477d4321a7e2d190f7, all tests pass.

Let me know if there are some issues that should be resolved before we merge this.

symengine/cwrapper.h
isuruf: MPC has output variable first, when there is no return variable. Eg:
http://www.multiprecision.org/index.php?prog=mpc&page=html#Projection-_0026-Decomposing

certik: Yes, those are functions like our `basic_add`, and there I think it is ok to return as the first value.

For methods of objects, I think it is customary to have the first argument as `self`.

symengine/cwrapper.h
```diff
@@ -1,45 +1,66 @@
...
+//  Assignment should be done only by using basic_assign(). Before the variable
+//  goes out of scope, basic_free() must be called.
+//
+//  For C, define a dummy struct with the right size, so that it can be
+//  allocated on the stack. For C++, the CRCPBasic is declared in cwrapper.cpp.
+#ifdef __cplusplus
+typedef struct CRCPBasic basic_struct;
+#else
+typedef struct CRCPBasic_C basic_struct;
+#endif
```
abinashmeher999: Any clues as to why it was not working when it was implemented the way `std::vector` and `std::set` were implemented?

symengine/cwrapper.h
abinashmeher999: I think when writing c interface functions for interacting with data types in C++, it makes more sense to keep `self` as the first parameter.

abinashmeher999:
@certik I need to implement `std::set` in the same way for `basic_free_symbols`. I can either send a PR to this or send a different PR after this is merged.

isuruf:
+1 to merge

certik:
@abinashmeher999 yes, send a new PR. It will work for `std::set` as well, you might have made some mistake somewhere.

abinashmeher999:
@certik  Ok I am on it. It was working for `std::set`, but not for `CRCPBasic`. I had implemented it same way as it was done for `CVecBasic`, wrapping it in a struct like [this](https://github.com/certik/symengine/pull/6/files#diff-72450588351eec10a19415c0a981ad41R67). If you remember [the error message](https://gist.github.com/abinashmeher999/38138341fbd9ac5c867e), it said `array type has incomplete element type`.

certik:
Rebase it, send a PR and let me have a look.

PR #509, opened 2015-07-06
Added c wrapper for set_basic and free_symbols

abinashmeher999:
I checked and there are no memory leaks in `test_cwrapper` after addition of these wrappers.
cc @certik @isuruf


certik:
This looks good to me. Travis passes, merging. Thanks!

PR #513, opened 2015-07-09
CMake improvements

certik:
This improves the summary, and also enables Teuchos automatically in Debug mode.


certik:
@abinashmeher999 would you mind reviewing this PR please?

CMakeLists.txt
```diff
@@ -297,10 +299,20 @@ if (WITH_PYTHON)
     add_subdirectory(symengine/python)
 endif ()
 
+if (WITH_PYTHON)
+    message("Copying source of python wrappers into: ${CMAKE_CURRENT_BINARY_DIR}/symengine/python/symengine")
+    file(COPY symengine/python/symengine/ DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/symengine/python/symengine)
+endif()
```
abinashmeher999: Why are the python source files being copied to where they already are? Is this to avoid anything I am missing? Isn't `${CMAKE_CURRENT_BINARY_DIR}` always the root dir?

abinashmeher999:
Looks good! Thanks for letting me review it.

CMakeLists.txt
certik: I just moved these lines up. When you build out of source, you need to copy the files, so that you can run the tests in the build directory directly.

certik:
Thanks for the review. Merging.

PR #514, opened 2015-07-09
Implement basic_new_heap() and basic_free_heap()

certik:
Test added.

@abinashmeher999 can you review this please? It implements the methods that you need to use to create an expression to pass to Ruby.

### DIFF
--- a/symengine/cwrapper.cpp
+++ b/symengine/cwrapper.cpp
@@ -52,6 +52,16 @@ void basic_free(basic s)
     s->m.~RCP();
 }
 
+basic_struct* basic_new_heap()
+{
+    return new CRCPBasic();
+}
+
+void basic_free_heap(basic_struct *s)
+{
+    delete s;
+}
+
 void symbol_set(basic s, char* c)
 {
     s->m = SymEngine::symbol(std::string(c));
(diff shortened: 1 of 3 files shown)


symengine/cwrapper.cpp
```diff
@@ -52,6 +52,16 @@ void basic_free(basic s)
     s->m.~RCP();
 }
 
+basic_struct* basic_new_heap()
+{
+    return new CRCPBasic();
```
abinashmeher999: Shouldn't `CRCPBasic()` be changed to `basic_struct`? Since we have used ifdefs to declare two structs `CRCPBasic` and `CRCPBasic_C` based on the compiler. They are ultimately typedeffed to `basic_struct`.

certik: Well, since this is a C++ file, they should be equivalent, and I felt it is more explicit to initialize CRCPBasic directly.

abinashmeher999: Yeah, I realised that a little late.

abinashmeher999:
+1

abinashmeher999:
Thanks @certik !

certik:
Thanks for the review @abinashmeher999 .

PR #519, opened 2015-07-11
Enumerating TypeID with namespaces in both C & C++

abinashmeher999:
Another cleaner approach that addresses #517 and also an alternative to #518 .


abinashmeher999:
cc @certik @isuruf

symengine/type_codes.h
```diff
@@ -0,0 +1,57 @@
+SYMENGINE_ENUM(INTEGER),
```
certik: We should see how other projects do this, but strictly speaking this file is not a header file anymore, since it won't work if you just include it. So perhaps we should rename it to `type_codes.inc`, or something similar.

symengine/cwrapper.cpp
```diff
@@ -62,6 +62,10 @@ void basic_free_heap(basic_struct *s)
     delete s;
 }
 
+TypeID basic_get_type(const basic s) {
+    return (TypeID) s->m->get_type_code();
```
certik: One should use probably `reinterpret_cast` or whatever the proper C++ cast is. One should not use the raw C cast in C++ codes.

certik:
I think that this looks good. Btw later we can probably extend this approach to simplify this part:

https://github.com/sympy/symengine/blob/d29cc084bb648d941955b0d746e8aa1a09f75252/symengine/visitor.cpp#L22

abinashmeher999:
@certik We should create an issue for using this approach in visitor.cpp

certik:
Cool, would you mind opening it please?

abinashmeher999:
Yeah sure.

abinashmeher999:
@certik Tests pass.

certik:
+1 to merge

PR #527, opened 2015-07-13
[WIP] Add diff.h/cpp and test it

certik:
The idea is to port all methods into separate functions. This PR tries to do it for `diff`. This is just the initial implementation, I need to port all the logic first.


Sumith1896:
Can be closed. Better approach merged here https://github.com/symengine/symengine/pull/738

certik:
Indeed.

PR #532, opened 2015-07-14
Implement EnableRcpFromThis base class

certik:
Benchmarks do not seem to be affected.

### DIFF
--- a/symengine/symengine_rcp.h
+++ b/symengine/symengine_rcp.h
@@ -261,6 +261,14 @@ public:
 
 #else
     mutable RCP<T> weak_self_ptr_;
+
+    void set_weak_self_ptr(const RCP<T> &w) {
+        weak_self_ptr_ = w;
+    }
+
+    void set_weak_self_ptr(const RCP<const T> &w) const {
+        weak_self_ptr_ = rcp_const_cast<T>(w);
+    }
 #endif // WITH_SYMENGINE_RCP
 };
 
@@ -272,7 +280,7 @@ inline RCP<T> make_rcp( Args&& ...args )
     return rcp( new T( std::forward<Args>(args)... ) );
 #else
     RCP<T> p = rcp( new T( std::forward<Args>(args)... ) );
-    p->weak_self_ptr_ = p.create_weak();
+    p->set_weak_self_ptr(p.create_weak());
     return p;
 #endif
 }
--- a/symengine/symengine_rcp.h
+++ b/symengine/symengine_rcp.h
@@ -257,7 +257,9 @@ private:
 #  else
     mutable unsigned int refcount_; // reference counter
 #  endif // WITH_SYMENGINE_THREAD_SAFE
+public:
     EnableRCPFromThis() : refcount_(0) {}
+private:
 
 #else
     mutable RCP<T> weak_self_ptr_;
@@ -271,6 +273,8 @@ private:
     }
 #endif // WITH_SYMENGINE_RCP
 
+template<class T_>
+friend class RCP;
 template<typename T_, typename ...Args>
 friend inline RCP<T_> make_rcp( Args&& ...args );
 };
--- a/symengine/symengine_rcp.h
+++ b/symengine/symengine_rcp.h
@@ -201,6 +201,20 @@ using Teuchos::print_stack_on_segfault;
 
 
 class EnableRCPFromThis {
+// Public interface
+public:
+    unsigned int use_count() const {
+#if defined(WITH_SYMENGINE_RCP)
+        return refcount_;
+#else
+        return weak_self_ptr_.strong_count();
+#endif
+    }
+
+
+
+
+// Everything below is private interface
 public:
 #if defined(WITH_SYMENGINE_RCP)
 
--- a/symengine/symengine_rcp.h
+++ b/symengine/symengine_rcp.h
@@ -203,11 +203,19 @@ class EnableRCPFromThis {
 // Public interface
 public:
     //! Get RCP<T> pointer to self (it will cast the pointer to T)
-    inline RCP<T> rcp_from_this() const {
+    inline RCP<T> rcp_from_this() {
 #if defined(WITH_SYMENGINE_RCP)
         return rcp(static_cast<T*>(this));
 #else
         return rcp_static_cast<T>(weak_self_ptr_.create_strong());
+#endif
+    }
+
+    inline RCP<const T> rcp_from_this() const {
+#if defined(WITH_SYMENGINE_RCP)
+        return rcp(static_cast<const T*>(this));
+#else
+        return rcp_static_cast<const T>(weak_self_ptr_.create_strong());
 #endif
     }
 
--- a/symengine/symengine_rcp.h
+++ b/symengine/symengine_rcp.h
@@ -220,6 +220,16 @@ public:
 #endif
     }
 
+    //! Get RCP<T2> pointer to self (it will cast the pointer to T2)
+    template <class T2>
+    inline RCP<T2> rcp_from_this_cast() const {
+#if defined(WITH_SYMENGINE_RCP)
+        return rcp(static_cast<T2*>(this));
+#else
+        return rcp_static_cast<T2>(weak_self_ptr_.create_strong());
+#endif
+    }
+
     unsigned int use_count() const {
 #if defined(WITH_SYMENGINE_RCP)
         return refcount_;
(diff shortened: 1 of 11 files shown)


certik:
Tests [pass](https://travis-ci.org/certik/symengine/builds/71122640) on Travis, this is ready for review.

@isuruf, let me know what you think.

symengine/symengine_rcp.h
```diff
@@ -198,6 +198,56 @@ using Teuchos::print_stack_on_segfault;
...
+    mutable RCP<const EnableRCPFromThis> weak_self_ptr_;
+#endif // WITH_SYMENGINE_RCP
+
+    //! Get RCP<T> pointer to self (it will cast the pointer to T)
+    template <class T>
+    inline RCP<T> get_rcp_cast() const {
+#if defined(WITH_SYMENGINE_RCP)
+        return rcp(static_cast<T*>(this));
+#else
+        return rcp_static_cast<T>(weak_self_ptr_.create_strong());
```
isuruf: This is from another PR, but what happens when you try to cast from `Add` to `Mul` in Debug mode?

isuruf:
Looks good

symengine/symengine_rcp.h
certik: Good point. It's going to segfault. We need to implement debug time
checking for casts.

certik:
I am reworking the interface to be equivalent to http://en.cppreference.com/w/cpp/memory/enable_shared_from_this.

certik:
I reworked the interface a bit. I am now happy with as it is. @isuruf please let me know if you have any objections to the latest changes.

certik:
@isuruf can you review this one? I think it's ready to be merged.

symengine/symengine_rcp.h
```diff
@@ -198,6 +198,73 @@ using Teuchos::print_stack_on_segfault;
...
+    inline RCP<T> rcp_from_this() {
+#if defined(WITH_SYMENGINE_RCP)
+        return rcp(static_cast<T*>(this));
+#else
+        return rcp_static_cast<T>(weak_self_ptr_.create_strong());
+#endif
+    }
+
+    //! Get RCP<const T> pointer to self (it will cast the pointer to const T)
+    inline RCP<const T> rcp_from_this() const {
```
isuruf: Should this be `inline const RCP<T> rcp_from_this() const {`?

certik: I used the same approach as in the standard library:

http://en.cppreference.com/w/cpp/memory/enable_shared_from_this/shared_from_this

Which also has the two methods, one returning `const T`, the other just `T`. One thing that I don't understand is that it seems the method itself never modifies the original object, so it seems both of the methods can be const. But for this PR I am just trying to follow the standard.

In the long run, @bartlettroscoe and I were thinking of trying to get the C++ standard extended, so that the `unique_ptr`, `shared_ptr` and `weak_ptr` can also return a `Ptr` class (it would probably be just called `ptr` or `view_ptr`) using a new `.ptr()` method, as opposed to a raw pointer as it is now using the `.get()` method. For self reference, one has to use `std::enable_shared_from_this`, and so our `EnableRCPFromThis` is a safe counterpart, and so I am trying to follow the same API. That way anybody using just pure C++ could use these classes. For now we have to implement these ourselves.

isuruf: Since `T` is `const Basic`, `const T` would be `const const Basic`. I guess it doesn't matter.
Anyways, +1 to merge

certik: I think that the compiler will pick up the right version of these routines, based on const. I could be wrong though.

bartlettroscoe: Ondrej,

You need the const and non-const functions to be const-correct.  That is, if a client as a const reference to the object, then it must give up `RCP<const T>`, not `RCP<T>` when it calls `rcp_from_this()`.  Otherwise, that would violate const correctness (which is very important in C++).   See Item 15 "Use const proactively" in "C++ Coding Standards".

certik: @bartlettroscoe thanks for the feedback. I would like to make it const correct, but I am a bit struggling what the right way to do this is. What I don't understand is that the C++ standard itself is using the following two functions (see http://en.cppreference.com/w/cpp/memory/enable_shared_from_this/shared_from_this):
```
shared_ptr<T> shared_from_this(); 	            (1) 	
shared_ptr<T const> shared_from_this() const;	(2) 
```
and I am just doing what they do. So it seems if there is a problem with this, it's already in the C++ standard.

bartlettroscoe: The standard is doing the right thing here.  The reason you need two overloaded functions is because the underlying weak pointer that is stored is of type `weak_ptr<T>`.  You need to function:
```
shared_ptr<T const> shared_from_this() const;
```
so that it will return a `shared_ptr<T const>` so it will not give up a non-const shared ptr to the object.  Otherwise, because C++ only enforces bitwise const, the function:
```
shared_ptr<T> shared_from_this() const
{  return weak_self_ptr_.lock(); }
```
would compile just fine and it would allow a client to change the object  For example, this would allow terrible code like:
```
void someFunc(const SomeType &obj)
{
  obj->shared_from_this()->modify(...);
}
```
where `SomeType::modify()` is a non-const member that changes the state of the object!

When programing in C++ and making things const correct, you have to implement "logical const" not just bitwise const.  The C++ compiler is too dumb to understand logical const.  We as programmers have to do that.  But if we design our classes and functions correctly, the compiler will enforce const for us (and only a const_cast<>() or a C-style cast can remove const and allow modification).  I love this about C++.

certik: @bartlettroscoe I see, thanks a lot for the explanation, it makes a lot more sense to me now. Yes, I agree. So I think the current code in this PR is correct, since these are my signatures:
```c++
    inline RCP<T> rcp_from_this();
    inline RCP<const T> rcp_from_this() const;
```
So if the user passes `const T&`, then the compiler will call the `const` version, which will return `RCP<const T>`, and if the user passes `T&`, then the compiler will call the non const version and returning `RCP<T>`, which I think is exactly what we need here.

certik:
@bartlettroscoe So I had it a bit wrong after all. Previously I had to use `class Basic : public EnableRCPFromThis<const Basic>` which always felt wrong to me. Now we can finally use `class Basic : public EnableRCPFromThis<Basic>` and things work. The problem was, as you said, that the internal pointer was just `T`, and so I had to introduce a const cast, i.e. when I set it in `make_rcp`:
```c++
template<typename T, typename ...Args>
inline RCP<T> make_rcp( Args&& ...args )
{
#if defined(WITH_SYMENGINE_RCP)
    return rcp( new T( std::forward<Args>(args)... ) );
#else
    RCP<T> p = rcp( new T( std::forward<Args>(args)... ) );
    p->set_weak_self_ptr(p.create_weak());
    return p;
#endif
}
```
the `set_weak_self_ptr()` method is overloaded as follows:
```c++
    mutable RCP<T> weak_self_ptr_;

    void set_weak_self_ptr(const RCP<T> &w) {
        weak_self_ptr_ = w;
    }

    void set_weak_self_ptr(const RCP<const T> &w) const {
        weak_self_ptr_ = rcp_const_cast<T>(w);
    }
```
In particular, in the second case I have to cast away the const. However, when the user asks for it using `rcp_from_this`, then it gets added back in as follows:
```c++
    //! Get RCP<T> pointer to self (it will cast the pointer to T)
    inline RCP<T> rcp_from_this() {
#if defined(WITH_SYMENGINE_RCP)
        return rcp(static_cast<T*>(this));
#else
        return rcp_static_cast<T>(weak_self_ptr_.create_strong());
#endif
    }

    //! Get RCP<const T> pointer to self (it will cast the pointer to const T)
    inline RCP<const T> rcp_from_this() const {
#if defined(WITH_SYMENGINE_RCP)
        return rcp(static_cast<const T*>(this));
#else
        return rcp_static_cast<const T>(weak_self_ptr_.create_strong());
#endif
    }
```
So I think this is all good. I've added thorough tests to test this behavior.

Does it look good now, or is there still some bug somewhere?

bartlettroscoe:
Yes, no const_casts<> should be needed if you can avoid it.

So other code does not manipulate your weak pointer, you will need to use friendship here to allow `make_rcp()` to access it but no other code.

certik:
I made all the internal members private and added RCP as a friend. Everything works now, tests pass. Thanks for the help @bartlettroscoe.

PR #533, opened 2015-07-15
Templated functions for **_eq and **_compare

### LINKED ISSUE #529: Templated functions for `*_eq` and `*_compare`
There many functions of the form `*_eq()` and `*_compare()` in the [dict.cpp](https://github.com/sympy/symengine/blob/master/symengine/dict.cpp) file.
Use templated functions instead as we already do for printing, so that the logic abstracts out.

From @certik's comment [here](https://github.com/sympy/symengine/pull/511#discussion_r34426918)

shivamvats:
@certik @Sumith1896 
Fixes #529

### DIFF
--- a/symengine/functions.cpp
+++ b/symengine/functions.cpp
@@ -1703,7 +1703,7 @@ bool Subs::__eq__(const Basic &o) const
 {
     if (is_a<Subs>(o) &&
             eq(*arg_, *(static_cast<const Subs &>(o).arg_)) &&
-            map_basic_basic_eq(dict_, static_cast<const Subs &>(o).dict_))
+            map_eq<map_basic_basic, map_basic_basic>(dict_, static_cast<const Subs &>(o).dict_))
         return true;
     return false;
 }
--- a/symengine/functions.cpp
+++ b/symengine/functions.cpp
@@ -1703,7 +1703,7 @@ bool Subs::__eq__(const Basic &o) const
 {
     if (is_a<Subs>(o) &&
             eq(*arg_, *(static_cast<const Subs &>(o).arg_)) &&
-            map_eq<map_basic_basic, map_basic_basic>(dict_, static_cast<const Subs &>(o).dict_))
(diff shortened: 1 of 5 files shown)


Sumith1896:
Thanks @shivamvats , it's a lot cleaner now.
+1 from my side.

symengine/functions.cpp
```diff
@@ -1703,7 +1703,7 @@ bool Subs::__eq__(const Basic &o) const
 {
     if (is_a<Subs>(o) &&
             eq(*arg_, *(static_cast<const Subs &>(o).arg_)) &&
-            map_basic_basic_eq(dict_, static_cast<const Subs &>(o).dict_))
+            map_eq<map_basic_basic>(dict_, static_cast<const Subs &>(o).dict_))
```
certik: Does it work if you call it as `map_eq(dict_, static_cast<const Subs &>(o).dict_))`? Many times you don't need to specify the template.

certik:
I think it looks good. The only question is whether we want to hide the templated functions in dict.cpp (and only expose the original non-templated versions that just call the templated ones underneath) to speed up the compilation.

symengine/functions.cpp
shivamvats: Indeed it does! Interesting.
I will make the changes.

shivamvats:
The only argument against doing so is that we'll need to define a function every time we want to use a new type with the templated function. What do you suggest?

certik:
Make the change to remove the template from the call and I guess we can keep it in the header file, it shouldn't be a big deal.

Sumith1896:
What if you pass two different types? Does it compile?
iirc it shouldn't but if it does, we need checks

shivamvats:
@certik Done
@Sumith1896 Why do you think so?

Sumith1896:
I felt it shouldn't compile. If it did, checking dicts are of same type or not would be needed.
Everything's fine now.

certik:
Try it and report back. I think it won't compile.

certik:
This looks good, merging.

shivamvats:
Thanks

PR #551, opened 2015-07-18
Host the mpir-2.7.0.zip on our github

certik:
That way if upstream server times out, we are not affected.


certik:
All tests pass, merging.

Issue #426, opened 2015-04-20
rcp_static_cast() should check that the cast is valid in Debug mode


certik:
And raise an exception. In Release mode, no check should be done (as it is now).

certik:
And also check the same for a regular `static_cast` if the argument is Basic (so we need to introduce our own safe function for that and use it instead).

janak2:
Has this been resolved? I'd like to work on it.

certik:
For `static_cast` I think we should just use https://github.com/google/protobuf/blob/master/src/google/protobuf/stubs/casts.h, i.e. `implicit_cast` and `down_cast`. Checked in Debug mode, fast in Release mode.

@janak2 no, it hasn't been resolved yet.

bollu:
@certik Is this fixed yet? If not, I'd like to take a swing at it

badcoderparashar:
is this fixed?

PR #562, opened 2015-07-24
Complex cwrapper

shivamvats:
@abinashmeher999 @isuruf 
Do we need *get* methods for rational/complex?


symengine/tests/cwrapper/test_cwrapper.c
```diff
@@ -48,6 +50,16 @@ void test_cwrapper() {
     printf("Is_a_Rational %s: %d\n", s, is_a_Rational(e));
     printf("Is_a_Integer %s: %d\n", s, is_a_Integer(e));
 
+    rational_set_ui(f, 76, 59);
```
abinashmeher999: @shivamvats You can make a different function to test the wrappers for complex, and use `SYMENGINE_C_ASSERT` to check if they are correct rather than `printf`.

abinashmeher999:
@shivamvats Thanks a lot for the help! :smiley: I can't say for sure if we will need that in ruby wrappers but they are good to have in C wrappers in case any other language needs it or some one wishes to use SymEngine in C.

symengine/tests/cwrapper/test_cwrapper.c
certik: Yes, these should rather by tested using `SYMENGINE_C_ASSERT`.

certik:
@abinashmeher999 you don't plan to support complex numbers in Ruby?

abinashmeher999:
@certik There is a little misunderstanding. By that I meant the get functions. I don't know if get* are the only ways to retrieve the value. That's why I said that. We will indeed support Complex. In fact we will support all the classes in SymEngine.

shivamvats:
@certik I am wondering why  Rational::from_two_ints takes RCP as argument while Complex::from_two_rats doesn't:
```
static RCP<const Number> from_two_rats(const Rational &re, const Rational &im);

 static RCP<const Number> from_two_ints(const RCP<const Integer> &n, const RCP<const Integer> &d);
```

certik:
@shivamvats Rational shouldn't either. You only accept RCP if you are storing the pointer somewhere. Otherwise you just accept a reference.

shivamvats:
`Rational::from_two_ints(integer(1), integer(2))` is used everywhere. Should I change it to `Rational::from_two_ints(*integer(1), *integer(2))` ?

shivamvats:
@certik Even `Rational::get_num_den` doesn't need an `RCP` as it's first argument: https://github.com/sympy/symengine/blob/master/symengine/rational.cpp#L74

Given a `RCP<const Basic> c`, how do I type-cast it to `Integer` before dereferencing? I have seen `*(rcp_static_cast<const Integer>(c))`. Can we do without `rcp_static_cast` ?

certik:
Yes, anytime the function body doesn't actually need the RCP pointer to the object, it should be converted to just a reference and use the object itself.

You can do something like `static_cast<const Integer &>(*c)`. 

We need to debug time check all these casts, that they are actually valid, see #426.

shivamvats:
Anything else to be added to this PR?

symengine/cwrapper.h
```diff
@@ -112,6 +112,13 @@ void rational_set_ui(basic s, unsigned long i, unsigned long j);
 //! Assign to s, a rational i, where is of type mpq_t.
 void rational_set_mpq(basic s, const mpq_t i);
 
+//! Assign to s, a complex re + i*im.
+void complex_set(basic s, const basic re, const basic im);
+//! Assign to s, a complex re + i*im, where re and im are rationals.
+void complex_set_rat(basic s, basic re, basic im);
```
isuruf: Change this to `void complex_set_rat(basic s, const basic re, const basic im);`

shivamvats: Done

symengine/tests/cwrapper/test_cwrapper.c
```diff
@@ -17,14 +17,14 @@ void test_cwrapper() {
     symbol_set(z, "z");
 
     s = basic_str(x);
-    printf("Symbol : %s\n", s);
+    SYMENGINE_C_ASSERT(strcmp(s, "x") == 0);
```
certik: You can remove ` == 0`, cannot you?

certik: And the same below.

shivamvats: I don't think we can do that. `strcmp` returns 0 if the two string are same.

certik: Ah, I misread the condition in `SYMENGINE_C_ASSERT`, you are right.

certik:
So the test doesn't work in this PR:
```
certik@redhawk:~/repos/symengine/symengine/tests/cwrapper(pr-562)$ ./test_cwrapper 

***
*** Warning! The following Teuchos::RCPNode objects were created but have
*** not been destroyed yet.  A memory checking tool may complain that these
*** objects are not destroyed correctly.
***
*** There can be many possible reasons that this might occur including:
***
***   a) The program called abort() or exit() before main() was finished.
***      All of the objects that would have been freed through destructors
***      are not freed but some compilers (e.g. GCC) will still call the
***      destructors on static objects (which is what causes this message
***      to be printed).
***
***   b) The program is using raw new/delete to manage some objects and
***      delete was not called correctly and the objects not deleted hold
***      other objects through reference-counted pointers.
***
***   c) This may be an indication that these objects may be involved in
***      a circular dependency of reference-counted managed objects.
***

  0: RCPNode (map_key_void_ptr=0xd99860)
       Information = {T=SymEngine::Rational, ConcreteT=SymEngine::Rational, p=0xd99860, has_ownership=1}
       RCPNode address = 0xd99940
       insertionNumber = 347
  1: RCPNode (map_key_void_ptr=0xd98110)
       Information = {T=SymEngine::Complex, ConcreteT=SymEngine::Complex, p=0xd98110, has_ownership=1}
       RCPNode address = 0xd96b60
       insertionNumber = 348

NOTE: To debug issues, open a debugger, and set a break point in the function where
the RCPNode object is first created to determine the context where the object first
gets created.  Each RCPNode object is given a unique insertionNumber to allow setting
breakpoints in the code.  For example, in GDB one can perform:

1) Open the debugger (GDB) and run the program again to get updated object addresses

2) Set a breakpoint in the RCPNode insertion routine when the desired RCPNode is first
inserted.  In GDB, to break when the RCPNode with insertionNumber==3 is added, do:

  (gdb) b 'Teuchos::RCPNodeTracer::addNewRCPNode( [TAB] ' [ENTER]
  (gdb) cond 1 insertionNumber==3 [ENTER]

3) Run the program in the debugger.  In GDB, do:

  (gdb) run [ENTER]

4) Examine the call stack when the program breaks in the function addNewRCPNode(...)
```
But the ctest test passes. @shivamvats when you develop, make sure you do that in Debug mode. That's very important to catch bugs like this one.

symengine/tests/cwrapper/test_cwrapper.c
```diff
@@ -83,11 +83,31 @@ void test_basic() {
     basic_free_heap(y);
 }
 
+void test_complex() {
+    basic e;
+    basic f;
+    char* s;
+    basic_init(e);
+    basic_init(f);
```
certik: The error is probably caused by allocating the `e` and `f` here, but never deallocating them. Remember that in C you need to deallocate by hand.

certik:
Here are some improvements: https://github.com/shivamvats/symengine/pull/3

symengine/tests/cwrapper/test_cwrapper.c
shivamvats: You are right. I am sorry, I shouldn't have overlooked this. 
I couldn't use Debug mode as my master [throws errors](https://gitter.im/sympy/symengine) while building in Debug mode. I am still trying to figure this out.

certik: What error are you getting? CMake in the log you posted seems to succeed .

shivamvats: I get errors while linking.
```
Linking CXX executable add1
CMakeFiles/add1.dir/add1.cpp.o: In function `main':
/home/aries/foss/symengine/benchmarks/add1.cpp:37: undefined reference to `SymEngine::pow(Teuchos::RCP<SymEngine::Basic const> const&, Teuchos::RCP<SymEngine::Basic const> const&)'
/home/aries/foss/symengine/benchmarks/add1.cpp:37: undefined reference to `SymEngine::mul(Teuchos::RCP<SymEngine::Basic const> const&, Teuchos::RCP<SymEngine::Basic const> const&)'
/home/aries/foss/symengine/benchmarks/add1.cpp:37: undefined reference to `SymEngine::add(Teuchos::RCP<SymEngine::Basic const> const&, Teuchos::RCP<SymEngine::Basic const> const&)'
/home/aries/foss/symengine/benchmarks/add1.cpp:38: undefined reference to `SymEngine::mul(Teuchos::RCP<SymEngine::Basic const> const&, Teuchos::RCP<SymEngine::Basic const> const&)'
collect2: error: ld returned 1 exit status
make[2]: *** [benchmarks/add1] Error 1
make[1]: *** [benchmarks/CMakeFiles/add1.dir/all] Error 2
make: *** [all] Error 2
```

certik: That's weird. Can you start from scratch, i.e. save your work, then do `git clean -dfx`, and then post the full output of
```
cmake -DCMAKE_BUILD_TYPE=Debug
make VERBOSE=1
```
This will print a lot of stuff, and if you post it, we should be able to figure out what's causing it. Also post platform and compiler version. Just create a new issue for it.

The idea is that if cmake succeed, then it should never fail like this during `make`. So we probably have to fix a bug in cmake somewhere.

shivamvats: It works after `git clean -dfx`. I think I might have corrupted some include path somehow.

Thanks a lot @certik!

certik:
Tests pass, this is +1 to merge.

shivamvats:
Thanks!

PR #578, opened 2015-08-09
[WIP] Polynomial wrappers

shivamvats:
Completes #511 

cc @certik @Sumith1896 

ToDo:

- [ ] Add encode and decode functions for `unordered_set`
- [ ] Add mul_poly


certik:
Thanks for working on this. First we need it to compile, see the Travis tests.

shivamvats:
Yes,  `encode` and ``decode` functions are missing as of now.

shivamvats:
@certik Travis(and other checks) do not test with piranha. How do I add a `WITH_PIRANHA` test?

certik:
@shivamvats I thought #585 does that.

shivamvats:
@certik Yes, he opened it after my comment.

certik:
We should make sure that at least one Travis test is testing this new code before merging.

Then we can figure out how to making it work using MSVC (and if Piranha doesn't work with it, which I think it doesn't, we can provide our own, perhaps slower, implementation of the classes that we use from Piranha).

shivamvats:
@Sumith1896  There are test failures.

shivamvats:
Closing in favour of #658

PR #583, opened 2015-08-11
Updated the beginner notebook and added example notebook

abinashmeher999:
I had to add `#to_s` method to the output of every expression to display the output. They are just a temporary workaround. The fact that the `Add`, `Mul` and `Pow` are now converted to `SymEngine::Basic` seems to have broken the expected behaviour of the `#to_iruby` method. I had so written the `#to_iruby` method, that it avoided the instances of purely `Basic` class, since they are uninitialised and cause an error. I will remove them after they(Add/Mul/Pow) are properly represented in Ruby.

### DIFF
--- a/symengine/ruby/notebooks/beginner.ipynb
+++ b/symengine/ruby/notebooks/beginner.ipynb
@@ -11,7 +11,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 1,
    "metadata": {
     "collapsed": false
    },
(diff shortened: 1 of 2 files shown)


abinashmeher999:
Also these functions that I covered in the notebook have been broken down into separate files in the [`examples/beginner`](https://github.com/sympy/sympy/tree/master/examples/beginner) directory in `sympy`. I hope keeping them in a single notebook is ok.

certik:
That looks good. +1

abinashmeher999:
I have one more notebook to add. Please hold the PR till then.

abinashmeher999:
I had added the other notebook yesterday itself. It is ready for review.

certik:
Very cool! Looks nice, thanks.

PR #585, opened 2015-08-12
Travis testing for Piranha dependency

Sumith1896:
We'll need to run the tests `WITH_PIRANHA=yes` as, at least for now, `Polynomial` will depend on `Piranha`
cc @certik @isuruf @shivamvats

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,20 @@ matrix:
     - compiler: clang
     - os: osx
   include:
+    - env: BUILD_TYPE="Debug" WITH_BFD="yes" WITH_PIRANHA="yes"
+      compiler: gcc
+      os: linux
+      addons:
+        apt:
+          sources:
+          - ubuntu-toolchain-r-test
+          - boost-latest
+          packages:
+          - libgmp-dev
+          - libmpfr-dev
+          - binutils-dev
+          - libboost-serialization1.55-dev
+          - g++-4.8
     - env: BUILD_TYPE="Debug" WITH_BFD="yes" WITH_PYTHON="yes" PYTHON_VERSION="2.7"
       compiler: clang
       os: linux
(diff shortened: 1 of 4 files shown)


certik:
You should skip building Piranhas tests and benchmarks and you should
compile it in Release mode (not sure if that's a default).

Sumith1896:
It is asking for the latest CMake, how do I install that?
Since there is no sudo access

isuruf:
You can install it into `$our_install_dir` by specifying `--prefix`.
Even though `cmake 2.8.10` is required in the CMakeLists.txt file, I think all features used are available in `cmake 2.8.7`. @bluescarni, is 2.8.10 necessary?

bluescarni:
@isuruf my recollection is that I kind of just picked an arbitrary CMake version which probably was reasonably recent at the time. I am not sure whether 2.8.7 would work or not to be honest.

isuruf:
@bluescarni, I just checked and 2.8.7 is enough for piranha. I posted https://github.com/bluescarni/piranha/pull/11.

@Sumith1896, you need to call cmake using `$our_install_dir/bin/cmake` to use the latest cmake you installed, otherwise old version is picked up.

To install piranha in `$our_install_dir` you can't give `--prefix`. instead you could give `-DCMAKE_INSTALL_PREFIX=$our_install_dir`

.travis.yml
```diff
@@ -60,6 +60,17 @@ matrix:
...
+      compiler: gcc
+      os: linux
+      addons:
+        apt:
+          sources:
+          - ubuntu-toolchain-r-test
+          - boost-latest
+          packages:
+          - libboost-serialization1.55-dev
+          - g++-4.8
```
isuruf: You have to add `libgmp-dev` and `libmpfr-dev` here as well as this entry is overriding the above project-wide entry.

Sumith1896:
@certik This is ready. Could you have a look?

certik:
I think that this looks good, great job!

Sumith1896:
Thanks for merging :)

PR #587, opened 2015-08-12
Install Sage from a tarball

certik:
@isuruf this works with `sudo: false` and it avoids saving the tarball. Here is a successful run: https://travis-ci.org/certik/symengine/jobs/75348011

The tarball itself is pretty big, since I use the tar.gz tarball, but as the next iteration, we can figure out how to use the lrzip tarball, which is 2x smaller --- I implemented it in #588, but that issue depends on Travis whitelisting `lrzip`, so for now if this PR works, I would suggest to merge it and we can improve upon it later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -135,12 +135,10 @@ matrix:
     - env: BUILD_TYPE="Release" WITH_SAGE="yes" WITH_PYTHON="yes" WITH_MPC="yes"
       compiler: gcc
       os: linux
-      sudo: true
   allow_failures:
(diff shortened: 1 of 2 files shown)


isuruf:
+1

certik:
Isuru, which error where you getting before? It seems this is working just
fine.

isuruf:
Build would just restart in the middle of the download and extraction. It seems this is working fine.

certik:
We'll leave it in the can fail section and watch it. If it works we'll move
it up.

PR #594, opened 2015-08-17
Build fix

cbehan:
Other systems' mileage may vary.

### DIFF
--- a/symengine/symengine_rcp.h
+++ b/symengine/symengine_rcp.h
@@ -12,6 +12,10 @@
 
 #if defined(WITH_SYMENGINE_RCP)
 
+#if defined(WITH_SYMENGINE_THREAD_SAFE)
+#include <atomic>
+#endif
+
 #else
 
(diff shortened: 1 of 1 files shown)


certik:
This should be included right here: https://github.com/sympy/symengine/blob/a48fa3efc6f14f91ebd1a966490812fd26997713/symengine/basic.h#L24, but I guess if you include only the `symengine_rcp.h` header, then it won't be included, so it fails. Can you post how you configured the cmake? Because we test this on Travis, so I would like to make sure we don't have a bug in tests.

cbehan:
The cmake flags I'm using are identical to the Archlinux ones: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=symengine-git

Compiling symengine_rcp.cpp is what fails because this includes symengine_rcp.h without basic.h.

certik:
Yes, all makes sense now. Thanks!

PR #629, opened 2015-10-24
resize vec_int in monomial_mul

mafiya69:
This will save from `SIGSEGV` if `C.size() < n`

### DIFF
--- a/symengine/monomials.cpp
+++ b/symengine/monomials.cpp
@@ -8,6 +8,7 @@ namespace SymEngine {
 void monomial_mul(const vec_int &A, const vec_int &B, vec_int &C)
 {
     size_t n = A.size();
+    C.resize(n);
     for(size_t i = 0; i < n; ++i) {
         C[i] = A[i] + B[i];
     }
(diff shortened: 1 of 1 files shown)


symengine/monomials.cpp
```diff
@@ -8,6 +8,7 @@ namespace SymEngine {
 void monomial_mul(const vec_int &A, const vec_int &B, vec_int &C)
 {
     size_t n = A.size();
+    C.resize(n);
```
certik: You don't want to be doing resize at this place, since this would slow things down.

Can you write a test case that shows the failure? Then we'll see how to best fix it.

Sumith1896:
You know the number of variables in `A` and in `B` and hence you accordingly pass `C` to this function. Resizing inside is not the right way to do it.

symengine/monomials.cpp
mafiya69: If `C.size() = n` then `C.resize(n)` will do nothing. So (I think) nothing slows down, else if `C.size() != n`, then function may give incorrect results.

certik: Just run the benchmarks to see.

You can add a check in Debug mode using SYMENGINE_ASSERT, that's a good idea.

However, you should not be resizing it here. As I said, try to write a test first, that segfaults. Then we'll figure out how to fix it.

Issue #362, opened 2015-01-11
Add series expansion


certik:
And compare against other codes. 

Here is one example from sympy: https://github.com/sympy/sympy/issues/2635

thilinarmtb:
+1

rwst:
Both `sympy.series` and the internal `rs_series` will need `Expr` member functions (because `Symbol` and `Function` subclasses from it). Currently everything in symengine subclasses from `Basic` not `Expression` so nothing will work that needs `Expr` functionality.

certik:
In SymEngine, we don't even have an Expr. In fact, it looks like we'll be able to implement all of the functionality using the visitor pattern or single dispatch, so the Basic class itself doesn't need many methods. However, in the Python wrappers, we certainly should make it work with both SymPy and Sage, so if it means we need to subclass Expr, then we should do it there.

rwst:
Another part would be a port of the `Order` function.

certik:
The `Order` function/class can be added in the higher level api. In the low level `rs_*` functions, the order is specified by the user and the function returns that many terms, so that things are fast.

rwst:
Univariate series expansion is now in symengine, so this can be closed.

PR #642, opened 2015-11-01
univar mpz_t series naive implementation using UnivariatePolynomial; part of #362

rwst:
This is just to have something in symengine, analogous to `UnivariatePolynomial`. After merge both should be renamed to reflect that they are `<univar, int_exp, mpz_coeff>` and independent of any optional library.


symengine/series.cpp
```diff
@@ -0,0 +1,171 @@
...
+}
+
+bool UnivariateSeries::is_canonical(const UnivariatePolynomial& poly, const unsigned int &prec)
+{
+    return true;
+}
+
+std::size_t UnivariateSeries::__hash__() const
+{
+    return poly_->__hash__() + std::size_t(prec_ * 84728863L);
```
isuruf: `poly_->hash()` would get the hash from the cache.

symengine/series.cpp
```diff
@@ -0,0 +1,171 @@
...
+    SYMENGINE_ASSERT(a.var_->get_name() == b.var_->get_name())
+    const unsigned int minprec = (a.prec_ < b.prec_)? a.prec_ : b.prec_;
+    unsigned int max = 0;
+    for (const auto &ait : a.poly_->dict_) {
+        const unsigned int aexp = ait.first;
+        if (aexp < minprec) {
+            for (const auto &bit : b.poly_->dict_) {
+                const unsigned int expsum = aexp + bit.first;
+                if (expsum < minprec)
+                    dict[expsum] += ait.second * bit.second;
```
isuruf: You can use `mpz_addmul` here

rwst:
Thanks, that gave 50% speedup. Let's see about the compile error.

rwst:
Why is that previous comment *above the other comments?

isuruf:
Daylight savings screwed up the timestamp?

rwst:
I don't get the Travis compile error here.

isuruf:
@rwst, that's probably due to the C++ version we are using. `operator bool()` is only defined for `mpz_class` if `#if __cplusplus >= 201103L`

rwst:
`error: ‘copy_if’ is not a member of ‘std’` ...come on! How old is this compiler?

isuruf:
It's GCC 4.6.3. :D
Did you try including the algorithm header?

isuruf:
Opened an issue #643

certik:
Ralf, what's the minimum compiler that Sage requires? We can require a bit
more modern gcc than 4.6 if Travis supports that.

rwst:
Sage has a minimum requirement of gcc-4.1 at the moment but will "soon" change that to 4.8 (see below). However, Piranha needs gcc-4.8 too, quote: "Piranha requires a recent compiler with robust support for the C++11 standard. The following compilers are routinely used to develop and test the library: ...GCC 4.8...". Note also that version 4.8 is the first with full C++11 support.

https://github.com/bluescarni/piranha/blob/master/doc/sphinx/getting_started.rst
http://trac.sagemath.org/ticket/19298

symengine/series.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+    poly_ = univariate_polynomial(var_, max, std::move(dict));
+
+    SYMENGINE_ASSERT(UnivariateSeries::is_canonical(*poly_, prec_))
+}
+
+UnivariateSeries::UnivariateSeries(const RCP<const Symbol> &var, const unsigned int &precision, const map_uint_mpz& dict) :
+        var_{var}, prec_{precision} {
+
+    map_uint_mpz dict_trunc = {};
```
isuruf: This line gives an error in AppleClang only.

isuruf:
Travis does support compiling with gcc 4.8. This is implemented for Piranha tests.
For other tests, we usually test with gcc 4.8 or above with the only exception being Ruby wrappers for windows with DevKit, which ships with MinGW gcc 4.7.3.

rwst:
4.7.3 is pretty complete as well, except for concurrency.
https://gcc.gnu.org/gcc-4.7/cxx0x_status.html

certik:
We can require gcc 4.8. I think that's fine to require complete C++11
support.

rwst:
Checks passed now. Coincidentally, that Sage ticket was merged yesterday and so Sage requires gcc-4.8.1 as well. I opened #644 for the compiler issue.

symengine/tests/basic/test_series.cpp
```diff
@@ -0,0 +1,167 @@
...
+TEST_CASE("Adding two UnivariateSeries", "[UnivariateSeries")
+{
+    RCP<const Symbol> x  = symbol("x");
+    map_uint_mpz adict_ = {{0, 1}, {1, 2}, {2, 1}};
+    map_uint_mpz bdict_ = {{0, 2}, {1, 3}, {2, 4}};
+
+    const UnivariateSeries a(x, 5, std::move(adict_));
+    const UnivariateSeries b(x, 4, std::move(bdict_));
+    RCP<const UnivariateSeries> c = add_uni_series(a, b);
+    REQUIRE(c->__str__() == "3 + 5*x + 5*x**2 + O(x**4)");
```
certik: Instead of testing the string representation (which can change, e.g. we might want to later decide to use `^` instead of `**` etc.), you should test the actual series itself, i.e. construct a third series that is equal to the addition, and then just compare the two series directly. The same for the rest of the tests.

symengine/series.cpp
```diff
@@ -0,0 +1,173 @@
...
+    std::copy_if(v.begin(), v.end(), std::back_inserter(vtrunc),
+        [&](decltype(v[0]) i) { return i < prec_; } );
+    poly_ = UnivariatePolynomial::create(var_, vtrunc);
+
+    SYMENGINE_ASSERT(UnivariateSeries::is_canonical(*poly_, prec_))
+}
+
+bool UnivariateSeries::is_canonical(const UnivariatePolynomial& poly, const unsigned int &prec)
+{
+    return true;
```
certik: Given the internal representation of the series, can it ever be not canonical? If not, then I would completely remove this `is_canonical` function and tests above, since they don't test anything.

certik:
Otherwise I think this looks good in general and it is a good start.

One design idea that I am not sure about is that we might want to have a lower level polynomial implementation that doesn't subclass Basic and doesn't require RCP. The advantage of it is that it is faster to create and deallocate. For bigger expressions that we want to apply series expansion to, this might provide a big speedup, i.e. if only the low level polynomial/series class is used to expand it, and only at the end it is wrapped into RCP so that it can be used as part of other expressions in SymEngine. For simple expressions, like expanding `sin(x)*cos(x)`, I expect that it wouldn't make a difference.

rwst:
Now MinGW is failing with
```
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!
```

symengine/series.cpp
certik: After 0160eb90babb59fb1f9e86fd2db057f1caebc130, `is_canonical` is never used, so you can now remove the method.

certik:
@isuruf I think that something has changed on Appveyor, that mingw doesn't work anymore. There is an error:
```
CMake Error at C:/Program Files (x86)/CMake/share/cmake-3.3/Modules/CMakeMinGWFindMake.cmake:22 (message):
  sh.exe was found in your PATH, here:
 
  C:/Program Files/Git/usr/bin/sh.exe
 
  For MinGW make to work correctly sh.exe must NOT be in your path.
 
  Run cmake from a shell that does not have sh.exe in your PATH.
 
  If you want to use a UNIX shell, then use MSYS Makefiles.
 
Call Stack (most recent call first):
  CMakeLists.txt:5 (project)
```
Which is probably causing it. So perhaps Appveyor added `sh.exe` into their image, which breaks mingw. That should be easy to fix hopefully.

certik:
@rwst just ignore the mingw errors. MSVC works, that's what matters. Mingw is just gcc I think, so if gcc work on Linux and Mac, it will work on Windows too.

isuruf:
Appveyor's git installation changed and the `sh.exe` path changed. Sent a PR #648

rwst:
So, nothing more in the way for merging?

certik:
I restarted the Appveyor build, just to be sure.

Otherwise the PR is +1 from me. Thanks for fixing it.

rwst:
Thanks for the review!

PR #646, opened 2015-11-02
Fix for #461

mafiya69:
Output of 
```
long xr = 6167950454l;
cout << " xr : " << xr << endl;
RCP<const Integer> ir = integer(xr);
cout << "*ir : " << *ir << endl;
```
Previous :
```
 xr : 6167950454
*ir : 1872983158
```
After this :
```
 xr : 6167950454
*ir : 6167950454
```

### DIFF
--- a/symengine/integer.h
+++ b/symengine/integer.h
@@ -7,6 +7,8 @@
 #ifndef SYMENGINE_INTEGER_H
 #define SYMENGINE_INTEGER_H
 
+#include <type_traits>
+
 #include <symengine/basic.h>
 #include <symengine/number.h>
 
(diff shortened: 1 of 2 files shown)


certik:
Thanks! Can you add a test that fails before and now passes?

mafiya69:
Though I have added the tests, they are passing in this PR. To see they are failing with master, see : https://travis-ci.org/mafiya69/symengine/builds/88921311

certik:
Perfect, this looks good. However, as it is currently, the `integer` function always constructs an `mpz_class`, that is then copied by value into the `Integer` constructor, so there is some overhead. I think a better design is to have multiple constructors in the `Integer` class itself, and then let the `integer` function just forward the argument. What do you think?

mafiya69:
Since all types are finally converting into `mpz_class`, I think if we have only one constructor of `Integer` class(accepting `mpz_class`) and having forwarding in `integer` function, all the values will be implicitly converted to `mpz_class`, otherwise they will be explicitly converted from their respective constructors(If we do constructor overloading). I implemented this approach in https://github.com/symengine/symengine/compare/master...mafiya69:fix461-1

certik:
That would work. I opened your patches into a PR #647.

PR #647, opened 2015-11-03
Fix for #461

certik:


symengine/integer.h
```diff
@@ -152,12 +152,13 @@ struct RCPIntegerKeyLess
         return false;
     }
 };
-//! \return RCP<const Integer> from `int`
-inline RCP<const Integer> integer(int i)
+//! \return RCP<const Integer> from integral values
+template<typename T> inline
+typename std::enable_if<std::is_integral<T>::value, RCP<const Integer>>::type
```
certik: Do we still need this `enable_if` in here? It seems it would work without it as well.

mafiya69: Also I think, we don't need the another overload for `integer` function. So I am removing `std::enable_if` and another overload of `integer`.

certik: Agreed

mafiya69:
@certik Anything more to do?

certik:
No, this looks great! Thanks for improving the code.

PR #649, opened 2015-11-05
Use const references

mafiya69:
I think 
```
for (auto& name : ds) {
    ...
}
```
can be replaced with 
```
for (const auto& name : ds) {
    ...
}
```
if `name` is not modified in `for` loop.


certik:
Excellent, good point. +1 to merge if tests pass.

Sumith1896:
+1

Sumith1896:
Thanks @mafiya69 
Good job.

certik:
Thanks @mafiya69, indeed very good job.

PR #662, opened 2015-11-14
implement skip-stop traversal, example has_function_of_symbol

rwst:


rwst:
Ultimately I did not need this, but you may be interested nevertheless.

certik:
We should add some tests showing how to use this.

Sumith1896:
@rwst There some changes in the master, `visitor.h` I think. This needs to be updated. Also, tests too.

certik:
I am going to close this for now. @rwst if you want this in, we can always finish it, add tests, etc. later.

PR #663, opened 2015-11-14
rational test file; convenience creator of rational from two longs; …

### LINKED ISSUE #660: rational syntactic sugar
I would like to do in SymEngine:
```
auto q = Rational(2,3);
```

rwst:
…fixes #660


certik:
That looks good. We should then move all the rational tests into this new file.

PR #667, opened 2015-11-23
rational is_perfect_power() and nth_root(); change rational creator

rwst:
Needed for series expansion.

### DIFF
--- a/symengine/rational.cpp
+++ b/symengine/rational.cpp
@@ -91,4 +91,63 @@ void get_num_den(const Rational &rat,
     *num = integer((rat.i).get_num());
     *den = integer((rat.i).get_den());
 }
+
+bool Rational::is_perfect_power(bool is_expected) const
+{
+    const mpz_class &num = i.get_num();
+    if (mpz_cmp_ui(num.get_mpz_t(), 0) == 0)
+        return true;
+    else if (mpz_cmp_ui(num.get_mpz_t(), 1) == 0)
+        return mpz_perfect_power_p(i.get_den().get_mpz_t()) != 0;
+
+    const mpz_class &den = i.get_den();
+    if (mpz_cmp_ui(den.get_mpz_t(), 1) == 0)
+        return mpz_perfect_power_p(num.get_mpz_t()) != 0;
+    
+    if (not is_expected) {
+        if (mpz_cmpabs(num.get_mpz_t(), den.get_mpz_t()) > 0) {
+            if (mpz_perfect_power_p(den.get_mpz_t()) == 0)
+                return false;
+        }
+        else {
+            if (mpz_perfect_power_p(num.get_mpz_t()) == 0)
+                return false;
+        }
+    }
+    mpz_t prod;
+    mpz_init(prod);
+    mpz_mul(prod, num.get_mpz_t(), den.get_mpz_t());
+    bool ret = mpz_perfect_power_p(prod) != 0;
+    mpz_clear(prod);
+    return ret;
+}
+
+bool Rational::r_nth_root(const Ptr<RCP<const Rational>> &the_rat, unsigned int n) const
+{
+    mpz_t rn, rd;
+    mpz_init(rn);
+    int ret = mpz_root(rn, i.get_num().get_mpz_t(), n);
+    if (ret == 0)
+        return false;
+    mpz_init(rd);
+    ret = mpz_root(rd, i.get_den().get_mpz_t(), n);
+    if (ret == 0)
+        return false;
+    mpq_t ratn, ratd;
+    mpq_init(ratn);
+    mpq_init(ratd);
+    mpq_set_z(ratn, rn);
+    mpq_set_z(ratd, rd);
+    mpz_clear(rn);
+    mpz_clear(rd);
+    mpq_t t;
+    mpq_init(t);
+    mpq_div(t, ratn, ratd);
+    mpq_clear(ratn);
+    mpq_clear(ratd);
+    mpq_class res(t);
+    mpq_clear(t);
+    *the_rat = make_rcp<const Rational>(res);
+    return true;
+}
 } // SymEngine
--- a/symengine/rational.cpp
+++ b/symengine/rational.cpp
@@ -85,8 +85,7 @@ int Rational::compare(const Basic &o) const
     }
     if (is_a<Integer>(o)) {
         const Integer &s = static_cast<const Integer &>(o);
-        mpq_class tmp(s.as_mpz());
-        return mpq_cmp(i.get_mpq_t(), tmp.get_mpq_t());
+        return i < s.i ? -1 : 1;
     }
     throw std::runtime_error("unhandled comparison of Rational");
 }
--- a/symengine/rational.cpp
+++ b/symengine/rational.cpp
@@ -78,10 +78,16 @@ bool Rational::__eq__(const Basic &o) const
 
 int Rational::compare(const Basic &o) const
 {
-    SYMENGINE_ASSERT(is_a<Rational>(o))
-    const Rational &s = static_cast<const Rational &>(o);
-    if (i == s.i) return 0;
-    return i < s.i ? -1 : 1;
+    if (is_a<Rational>(o)) {
+        const Rational &s = static_cast<const Rational &>(o);
+        if (i == s.i) return 0;
+        return i < s.i ? -1 : 1;
+    }
+    if (is_a<Integer>(o)) {
+        const Integer &s = static_cast<const Integer &>(o);
+        return mpq_cmp_z(i.get_mpq_t(), s.as_mpz().get_mpz_t());
+    }
+    throw std::runtime_error("unhandled comparison of Rational");
 }
 
 void get_num_den(const Rational &rat,
@@ -103,7 +109,7 @@ bool Rational::is_perfect_power(bool is_expected) const
     const mpz_class &den = i.get_den();
     if (den == 0)
         return mpz_perfect_power_p(num.get_mpz_t()) != 0;
-    
+
     if (not is_expected) {
         if (mpz_cmpabs(num.get_mpz_t(), den.get_mpz_t()) > 0) {
             if (mpz_perfect_power_p(den.get_mpz_t()) == 0)
@@ -128,8 +134,7 @@ bool Rational::r_nth_root(const Ptr<RCP<const Rational>> &the_rat, unsigned int 
     ret = mpz_root(rd.get_mpz_t(), i.get_den().get_mpz_t(), n);
     if (ret == 0)
         return false;
-    mpq_class res(rn);
-    res /= rd;
+    mpq_class res(rn, rd);
     res.canonicalize();
     *the_rat = make_rcp<const Rational>(res);
     return true;
--- a/symengine/rational.cpp
+++ b/symengine/rational.cpp
@@ -124,7 +124,7 @@ bool Rational::is_perfect_power(bool is_expected) const
     return mpz_perfect_power_p(prod.get_mpz_t()) != 0;
 }
 
-bool Rational::r_nth_root(const Ptr<RCP<const Rational>> &the_rat, unsigned int n) const
+bool Rational::nth_root(const Ptr<RCP<const Number>> &the_rat, unsigned int n) const
 {
     mpz_class rn;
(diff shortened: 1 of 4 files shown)


rwst:
Argh I just noticed a mem leak.

symengine/rational.cpp
```diff
@@ -91,4 +91,68 @@ void get_num_den(const Rational &rat,
...
+        if (mpz_cmpabs(num.get_mpz_t(), den.get_mpz_t()) > 0) {
+            if (mpz_perfect_power_p(den.get_mpz_t()) == 0)
+                return false;
+        }
+        else {
+            if (mpz_perfect_power_p(num.get_mpz_t()) == 0)
+                return false;
+        }
+    }
+    mpz_t prod;
```
certik: I would suggest to use `mpz_class`. Then you don't have to worry about memory leaks, as you find out, it is so easy to leak memory.

symengine/rational.cpp
```diff
@@ -91,4 +91,68 @@ void get_num_den(const Rational &rat,
...
+    mpz_clear(rd);
+    mpq_t t;
+    mpq_init(t);
+    mpq_div(t, ratn, ratd);
+    mpq_clear(ratn);
+    mpq_clear(ratd);
+    mpq_class res(t);
+    mpq_clear(t);
+    *the_rat = make_rcp<const Rational>(res);
+    return true;
```
certik: use `mpz_class` throughout.

rwst:
Anything else?

certik:
Thanks, it looks much better. Isn't the /= operation guaranteed to produce
canonical result? Or do you still need to canonicalize by hand.

symengine/rational.cpp
```diff
@@ -91,4 +91,47 @@ void get_num_den(const Rational &rat,
...
+    int ret = mpz_root(rn.get_mpz_t(), i.get_num().get_mpz_t(), n);
+    if (ret == 0)
+        return false;
+    mpz_class rd;
+    ret = mpz_root(rd.get_mpz_t(), i.get_den().get_mpz_t(), n);
+    if (ret == 0)
+        return false;
+    mpq_class res(rn);
+    res /= rd;
+    res.canonicalize();
```
certik: Since you call `canonicalize()` anyway, why not to just replace these three lines with just:
```c++
mpq_class res(rn, rd);
res.canonicalize();
```
?

symengine/rational.h
```diff
@@ -236,9 +239,12 @@ void get_num_den(const Rational &rat,
...
 //! convenience creator from two longs
-inline RCP<const Number> rational(long n, long d) {
-    return Rational::from_two_ints(n, d);
-    }
+inline RCP<const Rational> rational(long n, long d)
+{
+    mpq_class rat(n, d);
+    rat.canonicalize();
+    return make_rcp<const Rational>(rat);
+}
```
certik: What is the motivation of this replacement? The original code just called the following function:
```c++
RCP<const Number> Rational::from_two_ints(long n, long d)
{
    if (d == 0)
        throw std::runtime_error("Rational: Division by zero.");
    mpq_class q(n, d);

    // This is potentially slow, but has to be done, since 'n/d' might not be
    // in canonical form.
    q.canonicalize();

    return Rational::from_mpq(q);
}
```
Which does the same thing, except that one check for zero denominator.

certik: Ah I see --- the main difference is that `Rational::from_mpq` returns an Integer if `n/d` is an integer. While your new version will simply fail in Debug mode in that case (and undefined behavior in Release mode).

So we need to think how to best handle this.

symengine/rational.cpp
rwst: I can't find that constructor at https://gmplib.org/manual/C_002b_002b-Interface-Rationals.html#C_002b_002b-Interface-Rationals
Bad documentation. Do you have better docs?

certik: I think it's the "Function: mpq_class::mpq_class (integer num, integer den)", since it says that `integer` is mpz_class or standard C++ integer types.

symengine/rational.h
rwst: Again, this behaviour is not in the docs cited above. So how do you know?

rwst:
In the same way that you asked on gitter for low level and high level approach, wouldn't we need a constructor for `Rational` without check for canonical?

rwst:
BTW this restriction to nonintegers also leads to the inability to `compare` a `Rational` to an integer via https://github.com/symengine/symengine/blob/master/symengine/rational.cpp#L79

symengine/rational.h
certik: I am not sure if I understand your question, but the Rational constructor requires the numbers to actually be a rational, not an integer, see here:

https://github.com/symengine/symengine/blob/63e9d3fe44827576d9afe839d892974bb5fbb600/symengine/rational.cpp#L5

and the `is_canonical` function. Since in your new code you construct an `mpq_class` even if it is an integer, then it must fail in Debug mode in `is_canonical` in that case.

certik:
I don't mind redesigning the `Rational` class if you have a better design. As I pointed above, the current design is that `Rational` is only constructed if the `mpq_class` actually holds a rational, not an integer. If you think we should change that behavior, we can discuss it. Either way though, all the code should be consistent with whatever design we chose.

rwst:
First is that mathematically QQ contains ZZ. You don't have an equivalent restriction on `Complex` so consistency would already favor rationals with denominator 1. The only reason I see would be speed but there are no numbers, are there?

certik:
Indeed, we have a similar check for Complex, as you noted: https://github.com/symengine/symengine/blob/63e9d3fe44827576d9afe839d892974bb5fbb600/symengine/complex.cpp#L13

According to the documentation:

https://gmplib.org/manual/C_002b_002b-Interface-Rationals.html

"All arithmetic operators require their operands in canonical form, and will return results in canonical form." In other words, we have to ensure ourselves that any time we use `mpq_class` to do any kind of arithmetic operation, that it is in canonical form. We currently require that before the `Rational` is even constructed, so you never have to worry about it. Since GMP returns results in canonical form, this is a no brainer. The only place when you have to check this is if you construct `mpq_class` from integers. And so that's what we currently do.

So if I understand your proposal, your idea is to not check this when `Rational` is constructed, but rather check this every time we do anything with it? That seems more complicated both in terms of code and speed, since essentially every arithmetic method now has to check for this.

There is a special case when the denominator is 1. In that case, we already implement simplified algorithms ourselves (as you suggested above) --- by simply returning an Integer or `mpz`, and then we call integer arithmetic. Here I can see a possible slow down --- since GMP always returns `mpq_class` as a result of an arithmetic operation, even if the result is actually an integer. Currently we always check for this. The alternative is not to check for this, and just keep things rational even if they are integers. I am fine either way, I though it was cleaner to return an integer if we knew it was an integer, but perhaps we shouldn't do that. I know `piranha::integer` also doesn't automatically convert to machine integer from GMP, even if its the size would fit --- one has to do it by hand, so we could follow the same approach. But since there is already an overhead in wrapping things into Integer / Rational classes, I thought the simple check in constructor is negligible, and provides a nice clean design.

rwst:
Probably I'm focussing too much on modeling of the math, which is not SymEngine's job. I agree that, relieved of that burden, one-time canonicalizatuon is best.

rwst:
This error is due to the older gmplib version. There should be a version requirement stated in the README. Will try to work around the problem here.

symengine/rational.cpp
```diff
@@ -78,10 +78,16 @@ bool Rational::__eq__(const Basic &o) const
...
-    if (i == s.i) return 0;
-    return i < s.i ? -1 : 1;
+    if (is_a<Rational>(o)) {
+        const Rational &s = static_cast<const Rational &>(o);
+        if (i == s.i) return 0;
+        return i < s.i ? -1 : 1;
+    }
+    if (is_a<Integer>(o)) {
+        const Integer &s = static_cast<const Integer &>(o);
+        return mpq_cmp_z(i.get_mpq_t(), s.as_mpz().get_mpz_t());
```
isuruf: Arithmetic operators work for different combinations of `mpz_class` and mpq_class` right?

rwst:
@isuruf you suggest to replace the call to `mpz_cmp`? There would still a separate if case be needed I guess.

symengine/rational.cpp
```diff
@@ -78,10 +78,17 @@ bool Rational::__eq__(const Basic &o) const
...
-    return i < s.i ? -1 : 1;
+    if (is_a<Rational>(o)) {
+        const Rational &s = static_cast<const Rational &>(o);
+        if (i == s.i) return 0;
+        return i < s.i ? -1 : 1;
+    }
+    if (is_a<Integer>(o)) {
+        const Integer &s = static_cast<const Integer &>(o);
+        mpq_class tmp(s.as_mpz());
+        return mpq_cmp(i.get_mpq_t(), tmp.get_mpq_t());
```
isuruf: I was suggesting something like,

```
 if (is_a<Integer>(o)) {
     const Integer &s = static_cast<const Integer &>(o);
     if (i == s.i) return 0;
     return i < s.i ? -1 : 1;
```

rwst: You don't need the if because no Rational is integer, see above.

isuruf: Yes, forgot that.

rwst:
The next step would be to have `virtual Number::nth_root()` but I'll leave it for another PR.

certik:
So now when `rational` returns a `Rational`, the new behavior is that the following will raise an exception in Debug mode:
```
rational(3, 1)
rational(6, 2)
...
```
Because they are, or simplify (after canonicalization), to an integer. I am a bit worried that this can cause some issues.

certik:
In order to merge this PR, let's at least document the requirements. So the `rational` function should have a documentation that the arguments will be canonicalized, so they don't have to be in canonical form, but they cannot simplify to an integer.

certik:
I don't know, I feel that if your arguments are not in canonical form, then you also don't know whether they don't simplify to an integer (like `rational(9, 3)`). But if you don't know that, you can't use `rational` at all anymore. On the other hand, if you know for sure that your arguments won't simplify to an integer, then you presumably already have them in canonical form. For example, in the code from this PR in `r_nth_root`, I think the `rn` divided by `rd` is already in canonical form, isn't it? Or do you have an example when this isn't the case?

So if you know your arguments are already in canonical form, then you construct the `Rational` class directly as:
```c++
mpq_class res(rn, rd);
*the_rat = make_rcp<const Rational>(res); 
```

So it seems to me that you either construct the Rational class directly since you know what you are doing, or call `rational` if you don't know much about your arguments. But then `rational` should be converted to the old behavior, where it checks for an integer and returns an integer in that case.

certik:
The philosophy of symengine is that you impose as many restrictions in the `is_canonical` for each class as you can (and only check that in Debug mode), so that inside the class you can assume all those things and call faster algorithms (e.g. in Rational you know it's not an integer, so you don't need to worry about that special case, at the same time if you have an integer, you are forced to use the Integer class, thus automatically using faster algorithms for just integers). Then the idea is to use the information about the algorithm to construct arguments of the symengine classes in canonical form and then call the constructor without any checks. 

For cases where you can't or don't want to bother constructing in canonical form, we provide high level functions like `add, mul, pow, rational`, where you just provide arguments that are not necessarily in canonical form, and these functions will check and simplify. E.g. `add(x, x)` will check and simplify to `Mul(2, x)`, e.g. you never have the instance `Add(x, x)`. In the same spirit, `rational(2, 1)` will check and convert to `Integer(2)`, e.g. you never have `Rational(2, 1)`.

So for this reason, I actually think we should revert to the old behavior. I understand that you wanted `rational` to return an instance of the `Rational` class, but in the same way that `add` can't always return an instance of the `Add` class, `rational` can't always return the instance of the `Rational` class.

@rwst let me know what you think.

rwst:
@certik Together with one of your source comments I have wrapped this up and put it in https://github.com/symengine/symengine/wiki/Beginner-Contributor-Guide-Design-of-SymEngine

I agree but that means `virtual Number::nth_root()` and `::is_perfect_power()` is immediately necessary.

certik:
Thanks for writing it up! The PR is +1 from me. Thanks for fixing it.

certik:
I've added the integer tests for `rational()`  in #669.

PR #682, opened 2015-12-14
add order term skeleton, just for printing expansions

rwst:
Needed to get at least `series(..., method='sympy')` to produce correct output with https://github.com/symengine/symengine.py/pull/28. However, order term computations will not work:
```
In [1]: from symengine import *

In [2]: x=Symbol('x') 

In [3]: series(sin(x))
Out[3]: x + (-1/6)*x**3 + (1/120)*x**5 + O(x**6)

In [4]: _*series(sin(x))
Out[4]: (x + (-1/6)*x**3 + (1/120)*x**5 + O(x**6))**2

In [5]: _.expand()
Out[5]: 2*x*O(x**6) + (-1/3)*x**3*O(x**6) + (1/60)*x**5*O(x**6) + x**2 + (-1/3)*x**4 + (2/45)*x**6 + (-1/360)*x**8 + (1/14400)*x**10 + O(x**6)**2
```


rwst:
Hmm, shouldn't it be possible to just do `_._sympy_()` and get the original behaviour? If so I have a bug in  symengine.py#28

rwst:
It may be better to return a `sympy.core.add.Add` from `series()` until there is full series manipulation capability in SymEngine. That doesn't invalidate this PR however.

certik:
How about we instead create a Series class, that will know its order and
for printing it will print the O term? That class would store it's internal
state as a piranha polynomial, thus no unnecessary conversion is needed.
The class would have methods to convert to a symengine expression in terms
of Add/Mul etc. If you want operations with the O symbol, one just has to
convert to Series first. That seems like a clean, robust and fast design to
me. What do you think?

rwst:
I quite agree. It would be also substantially more work. The idea was first to have `series()` and atm I want to have a minimal usable version in symengine.py which is possible without this PR by returning SymPy objects. Sage users would have to convert. So this PR is useless and leads in the wrong direction. Let's remove it.

certik:
Yes, I am all for a minimal usable version. I think the version we have merged is such a version. The `Series` design can be built on top. I agree it will be some work, and we can do it after our low level piranha based code is performing optimally.

PR #703, opened 2015-12-25
documentation: merge everything from the Wiki on design plus from #445 into design.md

rwst:
Also-by: Ralf Stephan <ralf@ark.in-berlin.de>

### DIFF
--- a/doc/design.md
+++ b/doc/design.md
@@ -2,7 +2,21 @@
 
 Summary of our design decisions and some pointers to literature.
 
+## SymEngine is used from different languages
+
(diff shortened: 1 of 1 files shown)


certik:
Thanks for this! I'll read it later today.

doc/design.md
```diff
@@ -47,6 +61,53 @@ subset of all the functinoality and it requires all our objects to have a
...
+#endif
+public:
+    Ptr<std::map<int, int>> get_access() {
+#ifdef DEBUG_MODE
+        return m.ptr();
+#else
+        return ptrFromRef(m);
+#endif
+    }
+};
```
certik: Note that this is easier done using the `Viewable` class in https://github.com/certik/trilinos/pull/2, but we can update the docs once I implement it.

certik:
I read through this, and I think it is good enough. We can improve upon it later. Thanks for adding this in!

Issue #711, opened 2015-12-28
Move dependencies to the symengine/utilities dir


Sumith1896:
All the dependencies should be together. Hence, move them inside the newly made `utilities` directory.
Currently, the `catch` and `teuchos` needs to be moved.

Sumith1896:
Fixed via https://github.com/symengine/symengine/pull/715. Thanks @CodeMaxx for the work.

PR #712, opened 2015-12-28
Moved catch and teuchos dir to utilities dir

CodeMaxx:
Moved the `catch` and `teuchos` dir to `symengine/utilities` dir and changed the  `cmake` files accordingly so as it compiles properly.
But since I used an in-source compile the build files are also a part of the this commit . 
How are these to be removed?


CodeMaxx:
fix for #711

certik:
You committed too many files in this commit. You should only commit files that you changed.

CodeMaxx:
Yes I know . The build files have been committed as well. Is there a way to delete the files created by running `cmake .` and then `make` from my local directory? Otherwise I'll manually search for the files and push a commit.

CodeMaxx:
What about the moved files?

Sumith1896:
do a `git clean -dfx`

PR #719, opened 2015-12-29
[WIP] Symbol: use global counter

certik:
This compares integers instead of strings when comparing Symbols. This should be a lot faster. Currently it uses a global `int -> string` and `int-> hash` table. It changes the ordering a bit, i.e previously we compared e.g. "x" < "y", now we do e.g. 1 < 2, and so some tests fail. Most tests pass, so it should be working.

I am trying to see if it speeds up any benchmarks.

### DIFF
--- a/symengine/symbol.cpp
+++ b/symengine/symbol.cpp
@@ -4,21 +4,18 @@
 
 namespace SymEngine {
 
-Symbol::Symbol(const std::string &name)
-    : name_{name}
-{
-}
+std::map<int, std::string> Symbol::global_symbols;
+std::map<int, std::size_t> Symbol::global_symbols_hashes;
 
 std::size_t Symbol::__hash__() const
 {
-    std::hash<std::string> hash_fn;
-    return hash_fn(name_);
+    return global_symbols_hashes[name_counter_];
 }
 
 bool Symbol::__eq__(const Basic &o) const
 {
     if (is_a<Symbol>(o))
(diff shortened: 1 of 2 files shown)


certik:
This should just use `std::vector` instead of `std::map`. I'll fix it.

rikardn:
I know that this is a very old PR, but I still like the idea of trying to speed up symbol comparison. I would however like to bring up three problems with this particular approach.

1. Thread safety

The implementation is not thread safe since the global symbol map is written to without any guard.

2. Garbage collection

After creating many symbol objects and then freeing them again the symbol map will not be updated. This will effectively work as a memory leak. Since there is no global registry of symbols objects it would be impossible to create a garbage collector or to free elements in the map when they are no longer needed.

3. Global state

The suggested approach will introduce global state to symengine. Global state always gives me a bad feeling, since it makes it harder to debug, test and reason about program correctness. I cannot directly see any other issues than 1 and 2 above, but it still feels like future problems might arise. Serialization will potentially create problems for example.

1 and 2 should be possible to solve, but even though 3 is vague, it feels a bit scary.

One potential solution would be to introduce a `Context` class that can keep track of the symbol cache and be instantiated by the user and having a method that creates new symbols. Whenever the `Context` is freed the symbol map is freed and there would be no global state. This would solve both 2 and 3. I don't know if the added complexity makes it worth it though.

PR #720, opened 2015-12-30
SpeedUp Mul::__hash__()

mafiya69:


isuruf:
Did you check performance? Specifically `./expand6 1000`, `./expand2`, `./add1` and `./expand7` ?

mafiya69:
The performance of `./expand2` decreases from `1509 ms` to `3806 ms`!

certik:
Any ideas why the performance decreased so much?

isuruf:
I think it's because of hash collisions. In `Add` this wasn't noticed, because there was a speedup due to not ordering anymore.

certik:
You are probably right. std:map is already ordered, but we can play with
different data structures for Mul. Do you have examples of the hash
collisions, especially for Add? Maybe there is a stronger commutative hash
function.

isuruf:
With #718, `2*x + 2**(1/2)` hash equals `x + 2*2**(1/2)`, before #718, there were None.
What really matters is the bucket collisions. `std::unordered_map` uses buckets where each hash is mapped to a bucket and number of buckets is significantly lower than the number of hashes. So, even if all the hashes are different, what matters for performance is the bucket the hash ends up in.

You are right. There might be a stronger commutative hash.

PR #726, opened 2015-12-31
Latest gcc travis

CodeMaxx:
An additional travis-ci test added which compiles with latest gcc and g++ compilers(version 5.3 currently).
#684


CodeMaxx:
First 3 commits were from some other branch . I don't know why they are showing up here.

PR #734, opened 2016-01-01
[WIP] DiffVisitor 

Sumith1896:
Todo:
- [ ] - Get rid of member `diff()` functions
- [ ] - Add exhaustive set of functions
- [ ] - Add tests

cc @certik @isuruf


certik:
This looks great. We need to create benchmarks specifically for
differentiation and benchmark old and new diff.

P.S. You can probably unite the number methods into just one method for
type Number.

symengine/tests/basic/test_basic.cpp
```diff
@@ -314,40 +314,6 @@ TEST_CASE("Mul: Basic", "[basic]")
...
-    RCP<const Basic> i2  = integer(2);
-    RCP<const Basic> i3  = integer(3);
-    RCP<const Basic> i5  = integer(5);
-    RCP<const Basic> i10  = integer(10);
-    r1 = integer(5);
-    r2 = r1->diff(x);
-    REQUIRE(eq(*r2, *zero));
-
-    r1 = Rational::from_two_ints(*integer(2), *integer(3));
-    r2 = r1->diff(x);
```
srajangarg: @Sumith1896 The purpose of this, was to remove member functions as stated in https://github.com/symengine/symengine/pull/527#issue-94633819 and https://github.com/symengine/symengine/pull/731#discussion_r48684668
Right now, I think the member function is being called, not the ones you wrote.
Call it as `diff(r1, x)` instead (and everywhere else) . And also remove `Basic::diff()` and `diff()` from all other classes.

symengine/diff.cpp
```diff
@@ -0,0 +1,233 @@
...
+        result_ = zero;
+    }
+
+    void bvisit(const RealMPFR &x) {
+        result_ = zero;
+    }
+
+    void bvisit(const ComplexMPC &x) {
+        result_ = zero;
+    }
```
certik: All these can probably just be written as:
```c++
    void bvisit(const Number &x) {
        result_ = zero;
    }
```

symengine/tests/basic/test_basic.cpp
```diff
@@ -314,40 +314,6 @@ TEST_CASE("Mul: Basic", "[basic]")
...
-    REQUIRE(eq(*r1, *r2));
-
-    r1 = add(add(i2, mul(i3, x)), mul(i5, pow(x, i2)));
-    REQUIRE(eq(*r1->diff(x), *add(i3, mul(i10, x))));
-    REQUIRE(eq(*r1->diff(x)->diff(x), *i10));
-
-    r1 = add(mul(mul(pow(x, y), pow(y, x)), i2), one)->diff(x);
-    r2 = add(mul(i2, mul(pow(x, y), mul(pow(y, x), log(y)))), mul(i2, mul(pow(x, y), mul(pow(y, x), div(y, x)))));
-    REQUIRE(eq(*r1, *r2));
-}
```
certik: Do not remove these yet. You can move this test function into the `test_diff.cpp` file, but don't change it, as we need to keep the diff methods for now and benchmark it and make sure the visitor approach is actually faster before removing the old diff methods.

symengine/tests/diff/test_diff.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+    RCP<const Basic> r1;
+    RCP<const Basic> r2;
+
+    r1 = integer(5);
+    r2 = r1->diff(x);
+    REQUIRE(eq(*r2, *zero));
+
+    r1 = Rational::from_two_ints(*integer(2), *integer(3));
+    r2 = r1->diff(x);
```
certik: This should be converted to the new `diff`.

symengine/tests/diff/test_diff.cpp
```diff
@@ -0,0 +1,110 @@
...
+    RCP<const Basic> i6 = integer(6);
+    RCP<const Basic> i9 = integer(9);
+    RCP<const Basic> i10 = integer(10);
+    RCP<const Basic> i27 = integer(27);
+
+    RCP<const Basic> r1;
+    RCP<const Basic> r2;
+
+    r1 = integer(5);
+    r2 = r1->diff(x);
```
certik: This should be converted to the new `diff`.

certik:
This looks good, the new approach is simple and concise. As I said above, in this PR just implement the new algorithm, but keep the old one in place. It might take us a while to convince ourselves using benchmarks (and various compilers/platforms) that the new approach is not slower. 

If it is slower, then we'll have to think hard how to speed it up. I really like that we are taking stuff out of the core, it makes the code modular. If it turns out it is slower, then one idea would be to hook the new visitor methods as the old diff virtual methods using some macros, i.e. that one can still define a new algorithm in a separate file (without explicitly touching any core files), but use the old style virtual methods to implement it under the hood. We'll see.

symengine/diff.cpp
Sumith1896: Will do.

symengine/tests/basic/test_basic.cpp
Sumith1896: Oh, I was considering benchmarking this branch against my master. Will that do?

certik: Sure.

certik:
@Sumith1896 just finish this, but don't touch the core `diff`. See #738 where I am trying to do refactor the diff without using visitors. We'll benchmark this later. I would like to have both implementations in though.

certik:
@Sumith1896  I finished the #738, I think that's in good shape. @isuruf and I were trying to figure out how to unify the approach in #738 and the visitor pattern. It is actually possible, though there are some technical issues. The idea is that the visitor pattern has 2 virtual function calls, which in some situations must create an overhead. In some other situations it should be negligible if the amount of work done in the methods is large. 

So I think a good strategy is to refactor the rest of core using the approach in #738, which shouldn't create any overhead. And then figure out if there is a way to unify the two approaches further.

Sumith1896:
Much better approach here https://github.com/symengine/symengine/pull/738 
That's the way to go forward. We can close this PR for now.
@certik Shall I?

certik:
We can close this, I think I ported everything from here into #738. I didn't add new tests, since I am using all the old tests for diff.

PR #739, opened 2016-01-05
fixed travis check, wasn't exiting correctly

srajangarg:
Sorry, the initial pull request wasn't exiting with the correct return codes.
Fixed.


srajangarg:
@Sumith1896

certik:
What was it doing before? Always rising an exception since `exit` was not defined?

Sumith1896:
Yes. I didn't know of system exit return status. Thanks @srajangarg

PR #742, opened 2016-01-06
rcp_static_cast() should check that the cast is valid in Debug mode

abhimanyu-siwach:
This pull request is in reference to issue #426, and is by no means a complete solution to the problem, but more on the lines of a discussion. 

I added a typeid check in rcp_static_cast after reading from here: http://stackoverflow.com/questions/28002/regular-cast-vs-static-cast-vs-dynamic-cast 

Apart from the first test, all the tests threw cast error, here is debug trace for test_basic:
>Reading symbols from ./symengine/tests/basic/test_basic...done.
(gdb) run
Starting program: /home/sdetu/projects/symengine/symengine/tests/basic/test_basic 
terminate called after throwing an instance of 'std::runtime_error'
  what():  Type not same in rcp_static_cast N9SymEngine8RationalE N9SymEngine7IntegerE

>Program received signal SIGABRT, Aborted.
0x00007ffff6a84cc9 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56	  return INLINE_SYSCALL (tgkill, 3, pid, selftid, sig);
(gdb) backtrace 
#0  0x00007ffff6a84cc9 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff6a880d8 in __GI_abort () at abort.c:89
#2  0x00007ffff73996dd in __gnu_cxx::__verbose_terminate_handler() ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff7397746 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff7397791 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff73979a8 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00000000005c7ff5 in Teuchos::rcp_static_cast<SymEngine::Integer const, SymEngine::Basic const> (p1=...) at /home/sdetu/projects/symengine/symengine/utilities/teuchos/Teuchos_RCP.hpp:731
#7  0x00000000005d1154 in SymEngine::mul (a=..., b=...)
    at /home/sdetu/projects/symengine/symengine/mul.cpp:363
#8  0x00000000005d1640 in SymEngine::div (a=..., b=...)
    at /home/sdetu/projects/symengine/symengine/mul.cpp:374
#9  0x00000000005bdfb0 in SymEngine::sqrt_ (arg=...)
    at /home/sdetu/projects/symengine/symengine/constants.cpp:53
#10 0x00000000005be628 in __static_initialization_and_destruction_0 (__initialize_p=1, 
    __priority=65535) at /home/sdetu/projects/symengine/symengine/constants.cpp:63
#11 0x00000000005c0af2 in _GLOBAL__sub_I_constants.cpp(void) ()
    at /home/sdetu/projects/symengine/symengine/constants.cpp:121
#12 0x00000000006d903d in __libc_csu_init ()
#13 0x00007ffff6a6fe55 in __libc_start_main (main=0x68e7ba <main(int, char* const*)>, argc=1, 
>    argv=0x7fffffffdf88, init=0x6d8ff0 <__libc_csu_init>, fini=<optimized out>,

First of all I am not completely able to understand why this problem is arising, T2 is an Integer and *check is Rational, please correct me if I am wrong but I think an upcast should have happened.

Coming on to soultions, first one that came to my mind is to correct the cast parameters but that doesn't seems to be correct.
Second would be to check casts in a different way but I am not able come up with anything.

I am new to open source so kindly please try to ignore my vagueness.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ matrix:
     - compiler: clang
     - os: osx
   include:
-    - env: BUILD_TYPE="Debug" WITH_BFD="yes" WITH_PIRANHA="yes"
+    - env: BUILD_TYPE="Debug" WITH_BFD="yes" WITH_PIRANHA="yes" WITH_BENCHMARKS_NONIUS="yes"
       compiler: gcc
       os: linux
       addons:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,6 +203,22 @@ if (WITH_PIRANHA)
     set(WITH_BOOST yes)
 endif()
 
+# BENCHMARKS
+set(BUILD_BENCHMARKS yes
+    CACHE BOOL "Build SymEngine benchmarks")
+
+set(BUILD_BENCHMARKS_NONIUS no
+    CACHE BOOL "Build SymEngine Nonius benchmarks")
+
+if (BUILD_BENCHMARKS)
+    if (BUILD_BENCHMARKS_NONIUS)
+        if(NOT WIN32)
+            set(WITH_PTHREAD yes)
+        endif()
+        set(WITH_BOOST yes)
+    endif()
+endif()
+
 if (WITH_PTHREAD)
     find_package(PTHREAD REQUIRED)
     include_directories(${PTHREAD_INCLUDE_DIRS})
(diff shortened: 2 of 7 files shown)


certik:
Thanks for working on this. Why not to use dynamic_cast in Debug mode? That
should raise an exception or return null if the cast is invalid.

PR #744, opened 2016-01-06
rcp_static_cast() should check that the cast is valid in Debug mode

abhimanyu-siwach:
@certik is this what you were looking for?

With reference to #426.

### DIFF
--- a/symengine/utilities/teuchos/Teuchos_RCP.hpp
+++ b/symengine/utilities/teuchos/Teuchos_RCP.hpp
@@ -724,11 +724,14 @@ inline
 Teuchos::RCP<T2>
 Teuchos::rcp_static_cast(const RCP<T1>& p1)
 {
+#if defined(TEUCHOS_DEBUG)
+  return rcp_dynamic_cast<T2>(p1, true);
+#else
   // Make the compiler check if the conversion is legal
   T2 *check = static_cast<T2*>(p1.get());
   return RCP<T2>(check, p1.access_private_node());
+#endif
 }
-
 
 template<class T2, class T1>
 inline
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,7 @@ doc/html
 
 # Config Files
 symengine/symengine_config.h
-symengine/teuchos/Teuchos_config.h
+symengine/utilities/teuchos/Teuchos_config.h
 symengine/python/symengine/lib/config.pxi
 
 # Python build files


symengine/symengine_rcp.h
```diff
@@ -146,8 +146,12 @@ template<class T2, class T1>
 inline RCP<T2> rcp_static_cast(const RCP<T1>& p1)
 {
     // Make the compiler check if the conversion is legal
+#if defined(TEUCHOS_DEBUG)
+    return rcp_dynamic_cast<T2>(p1);
+#else
     T2 *check = static_cast<T2*>(p1.get());
     return RCP<T2>(check);
+#endif
```
certik: You don't need to modify this file at all, because this file is only used in Release mode. The reason we use Teuchos in Debug mode is that it is actually quite difficult to catch all segfaults and invalid memory accesses, Teuchos does it right, and so we just use it. In Release mode we use our faster versions. So there is no point of adding Debug checks into our code, since it won't be perfect anyway and it's not used.

symengine/basic-inl.h
```diff
@@ -57,6 +57,20 @@ inline void hash_combine(std::size_t& seed, const T& v)
...
+{
+    // Make the compiler check if the conversion is legal
+#if defined(TEUCHOS_DEBUG)
+    T2& dynamicCheck = dynamic_cast<T2&>(p1);
+    return dynamicCheck;
+#else
+    T2 staticCheck = static_cast<T2>(p1);
+    return staticCheck;
+#endif
+}
```
certik: This can be removed, right? I don't see you use it anywhere in the PR.

symengine/utilities/teuchos/Teuchos_RCP.hpp
```diff
@@ -725,8 +725,12 @@ Teuchos::RCP<T2>
 Teuchos::rcp_static_cast(const RCP<T1>& p1)
 {
   // Make the compiler check if the conversion is legal
```
certik: Move this comment into the `#else` section.

certik:
Yes, I think that's it. This checks the `rcp_static_cast` in Debug mode. I left some comments to further improve the PR.

abhimanyu-siwach:
@certik for the second part of issue #426, after implementing implicit_cast() and down_cast() I would be required to change all static_cast where Basic or it's derivative is present as argument to down_cast(), right?

certik:
@abhimanyu-siwach that's right, I think all our usage of `static_cast` is `down_cast`. Let's do that in a separate PR.

In order to finish this PR, would you mind tidying the history a bit, and just create 2 simple commits, one modifying the `.gitignore` and the other adding the check into `rcp_static_cast`?

certik:
Thanks!

PR #745, opened 2016-01-07
added polygamma function

abhinavagarwal07:
- [x] added `polygamma` function
- [x] added tests

https://github.com/symengine/symengine/pull/722


certik:
I think this looks good. Thanks @abhinavagarwal07 !

PR #746, opened 2016-01-07
Added support for down_cast() for safe static cast in debug mode.

abhimanyu-siwach:
This pull request would be in two phases:
- [x] Implementing down_cast()
- [ ] changing static_cast() to down_cast() in the existing code base.

@certik there aren't many changes from the [link](https://github.com/google/protobuf/blob/master/src/google/protobuf/stubs/casts.h) you gave in #426, but I would like you to confirm that it's fine. I have couple of queries that I wanted to ask: 
- What should be the scope of down_cast()? (currently defined in SymEngine namespace)
- Where to define it? (currently defined in basic-inl.h)

Meanwhile, I am going to start working on the second phase.

### DIFF
--- a/symengine/symengine_casts.h
+++ b/symengine/symengine_casts.h
@@ -54,7 +54,7 @@ template<typename T> struct remove_reference<T&> { typedef T type; };
 //   implicit_cast<ToType>(expr)
 
 template <typename To, typename From>
-inline To implicit_cast(const From &f){
+inline To implicit_cast(const From &f) {
     return f;
 }
 
@@ -77,7 +77,7 @@ inline To implicit_cast(const From &f){
 // You should design the code some other way not to need this.
 
 template <typename To, typename From>	 // use like this: down_cast<T*>(foo).
-inline To down_cast(From* f){				// only accept pointers.
+inline To down_cast(From* f) {				// only accept pointers.
     // Ensures that To is a sub-type of From *.  This test is here only
     // for compile-time type checking, and has no overhead in an
     // optimized build at run-time, as it will be optimized away
@@ -86,9 +86,8 @@ inline To down_cast(From* f){				// only accept pointers.
         implicit_cast<From*, To>(0);
     }
 
-#if defined(TEUCHOS_DEBUG) 
     SYMENGINE_ASSERT(f == NULL || dynamic_cast<To>(f) != NULL);  // debug mode only!
-#endif
+    
     return static_cast<To>(f);
 }
 
@@ -103,9 +102,8 @@ inline To down_cast(From& f) {
         implicit_cast<From*, ToAsPointer>(0);
     }
 
-#if defined(TEUCHOS_DEBUG) 
     SYMENGINE_ASSERT(dynamic_cast<ToAsPointer>(&f) != NULL); // debug mode only!
-#endif
+
     return *static_cast<ToAsPointer>(&f);
 }
 


symengine/basic-inl.h
```diff
@@ -40,6 +40,53 @@ inline bool is_a_sub(const Basic &b)
...
+    // Ensures that To is a sub-type of From *.  This test is here only
+    // for compile-time type checking, and has no overhead in an
+    // optimized build at run-time, as it will be optimized away
+    // completely.
+    if (false) {
+        implicit_cast<From*, To>(0);
+    }
+
+#if defined(TEUCHOS_DEBUG) 
+    assert(f == NULL || dynamic_cast<To>(f) != NULL);  // debug mode only!
```
certik: Use `SYMENGINE_ASSERT`, because the regular `assert` is enabled/disabled with `NDEBUG` I think.

symengine/utilities/teuchos/Teuchos_RCP.hpp
```diff
@@ -733,6 +733,7 @@ Teuchos::rcp_static_cast(const RCP<T1>& p1)
 #endif
 }
 
+
```
certik: You can remove this new line.

certik:
Define it in a new header file called e.g. `casts.h`. Also, copy the license from the original link at the top of the `casts.h` file.

certik:
Otherwise it looks good.

certik:
As to the namespace, keep it in `SymEngine` for now.

certik:
Also keep some of the original comments about how it works. I think those are useful.

symengine/utilities/teuchos/Teuchos_RCP.hpp
abhimanyu-siwach: I noticed later that rest of the functions in this file have two empty line in between them. That's why I added the new line.

certik: Ok.

symengine/symengine_casts.h
```diff
@@ -0,0 +1,114 @@
...
+    // for compile-time type checking, and has no overhead in an
+    // optimized build at run-time, as it will be optimized away
+    // completely.
+    if (false) {
+        implicit_cast<From*, To>(0);
+    }
+
+#if defined(TEUCHOS_DEBUG) 
+    SYMENGINE_ASSERT(f == NULL || dynamic_cast<To>(f) != NULL);  // debug mode only!
+#endif
```
certik: Actually, since SYMENGINE_ASSERT is only enabled in Debug mode, you can now remove the `TEUCHOS_DEBUG` ifdef. It makes sense, since this `down_cast` is not part of Teuchos anyway.

symengine/symengine_casts.h
```diff
@@ -0,0 +1,114 @@
...
+    // for compile-time type checking, and has no overhead in an
+    // optimized build at run-time, as it will be optimized away
+    // completely.
+    if (false) {
+        implicit_cast<From*, ToAsPointer>(0);
+    }
+
+#if defined(TEUCHOS_DEBUG) 
+    SYMENGINE_ASSERT(dynamic_cast<ToAsPointer>(&f) != NULL); // debug mode only!
+#endif
```
certik: Same here.

certik:
I think it looks good. Now we just have to use it.

symengine/symengine_casts.h
```diff
@@ -0,0 +1,112 @@
...
+    // Ensures that To is a sub-type of From *.  This test is here only
+    // for compile-time type checking, and has no overhead in an
+    // optimized build at run-time, as it will be optimized away
+    // completely.
+    if (false) {
+        implicit_cast<From*, To>(0);
+    }
+
+    SYMENGINE_ASSERT(f == NULL || dynamic_cast<To>(f) != NULL);  // debug mode only!
+
```
certik: It's hard to see, but there is a trailing white space here, so that's why tests fail.

certik:
@abhimanyu-siwach do you want to finish this PR? I think it looks great and we should use it. It needs to be merged with master to resolve conflicts.

rwst:
@abhimanyu-siwach ping

certik:
This has been fixed by #1124. Thanks @abhimanyu-siwach for your patches, which have been reused in #1124.

PR #749, opened 2016-01-09
type id variable added

### LINKED ISSUE #387: Try to store type_code_id as a variable
So each constructor will assign the type into this variable, and then the `get_type_code()` method, which currently is virtual, will become a non-virtual method that just returns the `type_code_id` variable.

So the invocation of the `get_type_code()` method should become a lot faster, which should speedup single dispatch and any other code the requires frequent type checks. The slow down will be twofold:

* each constructor now must assign to an extra variable, so it will be slower
* each Basic instance will have an extra variable, so more space is needed on heap to allocate it

So overall I expect some benchmarks will get slower and some other ones faster. It might be a good deal overall, but we need to benchmark this carefully.

While playing with this, it is essential to keep in mind, that the most accessed variable needs to be declared first in the Basic class (I forgot the exact technical platform dependent reasons, it has something to do with cache or the way CPU works). Currently we have `hash_`, `refcount_` in this order. I tried to swap them and I think things got a little faster on the `expand2` benchmark.

Things to try:

* So we should try to put the type as the first variable.

* Also we should experiment with using perhaps `char` (or `short`) as the type, instead of `int`, it might be faster. I expect we'll have less than 256 types, but if we have more, we'll use `short` -- it might not even have any significant performance benefit, in which case we should use `int`.

kanchanarp:
type_id_ variable added in place of the virtual method

Fixes #387


isuruf:
There are only 49 occurrences of `this->type_code_` excluding multiple constructors in your changes, but there are 53 classes listed in `type_codes.inc`. Add them as well to prevent to prevent failures.

Can you check benchmarks to see if there's a speedup due to this? (i.e. run `./expand2` `./expand6 1000` `./expand7`)

kanchanarp:
Attached herewith is a summary on the benchmark times and the original log files of the benchmark runs in text format.
[benchmark results.xlsx](https://github.com/symengine/symengine/files/83983/benchmark.results.xlsx)
[log_type_id_variable.txt](https://github.com/symengine/symengine/files/83984/log_type_id_variable.txt)
[log_virtual_method.txt](https://github.com/symengine/symengine/files/83985/log_virtual_method.txt)

kanchanarp:
average time
with type_id_
expand2: 1549.1	
expand7: 211.2 	
expand6 1000: 6073.8

with virtual method
expand2: 1626.4	
expand7: 211.7	
expand6 1000: 9233.5

certik:
That looks very encouraging. I thought we got a slowdown last time we tried
it, didn't we?

kanchanarp:
@certik @isuruf 
There has been a mistake in my calculation in expand6 in "with virtual method" average should be 6246ms.

10 benchmark results:
6199ms
6184ms
6212ms
6187ms
6269ms
6248ms
6245ms
6548ms
6184ms
6184ms

symengine/basic.h
```diff
@@ -86,6 +86,8 @@ enum TypeID {
 #include "basic-methods.inc"
 
 class Basic : public EnableRCPFromThis<Basic> {
+protected:
+    TypeID type_code_;
```
mafiya69: [CppCoreGuildelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c133-avoid-protected-data) says avoid `protected` data, I think you shouldn't use `protected` members.

certik:
I finally got to test this branch. I merged it (e51668b) with the latest master
(7728026). I compiled with:
```
cmake -DCMAKE_PREFIX_PATH=$HASHSTACK -DWITH_TCMALLOC=yes -DBUILD_BENCHMARKS_NONIUS=yes ..
```
It's essential to compile with TCMALLOC. Some results:

**This branch**:
The fastest out of about 10 runs (by hand):
expand1: 63ms
expand2: 629ms
expand3: 14ms
```
$ ./benchmarks/bench
clock resolution: mean is 964.075 ns (640002 iterations)

benchmarking expand1
collecting 100 samples, 1 iterations each, in estimated 7.8398 s
mean: 66.9491 ms, lb 66.6146 ms, ub 67.2389 ms, ci 0.95
std dev: 1579.69 μs, lb 1410.61 μs, ub 1920.81 μs, ci 0.95
found 1 outliers among 100 samples (1%)
variance is moderately inflated by outliers

benchmarking expand2
collecting 100 samples, 1 iterations each, in estimated 67.3823 s
mean: 700.397 ms, lb 699.792 ms, ub 700.724 ms, ci 0.95
std dev: 2.20336 ms, lb 1364.89 μs, ub 3.61918 ms, ci 0.95
found 7 outliers among 100 samples (7%)
variance is unaffected by outliers

benchmarking expand3
collecting 100 samples, 1 iterations each, in estimated 2.03817 s
mean: 10.0574 ms, lb 10.0471 ms, ub 10.0817 ms, ci 0.95
std dev: 76.0163 μs, lb 38.2069 μs, ub 152.322 μs, ci 0.95
found 15 outliers among 100 samples (15%)
variance is unaffected by outliers
$ ./benchmarks/symbench
Time for R1 :        0.000402026
Time for R2 :        0.004701725
Time for R3 :        0.000002305
Time for R5 :        0.000007752
Time for R7 :        0.025192250
Time for R8 :        0.011029580
Time for S1 :        0.010228323
Time for S2 :        0.011218300
Time for S3 :        0.008236285
Time for S3a :       0.941355921
```



**vs master**:
The fastest out of about 10 runs (by hand):
expand1: 63ms
expand2: 655ms
expand3: 15ms
```
$ ./benchmarks/bench
clock resolution: mean is 1045.01 ns (640002 iterations)

benchmarking expand1
collecting 100 samples, 1 iterations each, in estimated 7.82994 s
mean: 68.8462 ms, lb 68.4641 ms, ub 69.158 ms, ci 0.95
std dev: 1750.51 μs, lb 1512.21 μs, ub 2.35664 ms, ci 0.95
found 1 outliers among 100 samples (1%)
variance is moderately inflated by outliers

benchmarking expand2
collecting 100 samples, 1 iterations each, in estimated 73.7778 s
mean: 758.108 ms, lb 756.037 ms, ub 759.255 ms, ci 0.95
std dev: 7.64767 ms, lb 4.5078 ms, ub 11.3127 ms, ci 0.95
found 12 outliers among 100 samples (12%)
variance is unaffected by outliers

benchmarking expand3
collecting 100 samples, 1 iterations each, in estimated 1893.86 ms
mean: 11.062 ms, lb 11.0574 ms, ub 11.0692 ms, ci 0.95
std dev: 28.8902 μs, lb 20.4739 μs, ub 41.9436 μs, ci 0.95
found 7 outliers among 100 samples (7%)
variance is unaffected by outliers
$ ./benchmarks/symbench
Time for R1 :        0.000431710
Time for R2 :        0.005255947
Time for R3 :        0.000002515
Time for R5 :        0.000008102
Time for R7 :        0.027809747
Time for R8 :        0.011288285
Time for S1 :        0.010803214
Time for S2 :        0.012881860
Time for S3 :        0.010254935
Time for S3a :       1.128350458
```


So it looks like this PR does speed things up across the board.

symengine/add.cpp
```diff
@@ -13,6 +13,7 @@ Add::Add(const RCP<const Number> &coef, umap_basic_num&& dict)
     : coef_{coef}, dict_{std::move(dict)}
 {
     SYMENGINE_ASSERT(is_canonical(coef, dict_))
+    this->type_code_ = type_code_id;
```
certik: Put this above `SYMENGINE_ASSERT(is_canonical(coef, dict_))`

certik:
Instead of the lines `this->type_code_ = type_code_id;`, put there instead `ASSIGN_TYPEID`. And define the `ASSIGN_TYPEID` macro right below `IMPLEMENT_TYPID`.

certik:
I didn't finish my thought --- by doing that, we can then easily switch between the current version and the new version with a simple define. And I think this is something which we want to control, and perhaps switch back later if it turns out to be faster.

isuruf:
@kanchanarp, let me know if you need any help with this

srajangarg:
@kanchanarp @isuruf can I pick this up?

isuruf:
@srajangarg, sure

srajangarg:
@certik >It's essential to compile with TCMALLOC

Why do you say so?

certik:
@srajangarg we observed good speedups with TCMALLOC, also the timings were more consistent.

srajangarg:
closing. Work has been continued in https://github.com/symengine/symengine/pull/1153

PR #758, opened 2016-01-14
Minor fix in .travis.yml

CodeMaxx:
Since the code quality script was replaced, this line is no more necessary.
Also any specific reason why other lint checks mentioned in #743 were not implemented?
@isuruf @certik


abhinavagarwal07:
:+1

PR #780, opened 2016-01-26
Flint univariate series expansion

rwst:
Flint polys cannot do `cot()`, `csc()`, and `lambert_w()` because they cannot handle Laurent polys.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,6 +135,11 @@ if (WITH_PRIMESIEVE)
     set(PKGS ${PKGS} "PRIMESIEVE")
 endif()
 
+# FLINT
+set(WITH_FLINT no
+    CACHE BOOL "Build with Flint")
+set(HAVE_SYMENGINE_FLINT False)
+
 # ARB
 set(WITH_ARB no
     CACHE BOOL "Build with Arb")
@@ -150,6 +155,19 @@ set(WITH_MPC no
     CACHE BOOL "Build with MPC")
 set(HAVE_SYMENGINE_MPC False)
 
+if (WITH_FLINT)
+    find_package(FLINT REQUIRED)
+    include_directories(${FLINT_INCLUDE_DIRS})
+    set(LIBS ${LIBS} ${FLINT_TARGETS})
+    set(HAVE_SYMENGINE_FLINT True)
+    set(PKGS ${PKGS} "FLINT")
+
+    # Workaround for https://github.com/fredrik-johansson/arb/issues/24
+    include_directories(${FLINT_INCLUDE_DIRS}/flint)
+
+    set(WITH_MPFR yes)
+endif()
+
 if (WITH_ARB)
     find_package(FLINT REQUIRED)
     include_directories(${FLINT_INCLUDE_DIRS})
@@ -159,6 +177,8 @@ if (WITH_ARB)
     # Workaround for https://github.com/fredrik-johansson/arb/issues/24
     include_directories(${FLINT_INCLUDE_DIRS}/flint)
 
+    set(WITH_FLINT yes)
+    set(HAVE_SYMENGINE_FLINT True)
     set(WITH_MPFR yes)
 
     find_package(ARB REQUIRED)
@@ -384,6 +404,7 @@ configure_file(cmake/SymEngineConfigVersion.cmake.in SymEngineConfigVersion.cmak
 
 install(FILES "${PROJECT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/SymEngineConfig.cmake"
               ${PROJECT_BINARY_DIR}/SymEngineConfigVersion.cmake
+              cmake/FindFLINT.cmake
               cmake/FindARB.cmake
               cmake/FindBFD.cmake
               cmake/FindECM.cmake
@@ -461,6 +482,12 @@ message("WITH_PRIMESIEVE: ${WITH_PRIMESIEVE}")
 if (WITH_PRIMESIEVE)
     message("PRIMESIEVE_INCLUDE_DIRS: ${PRIMESIEVE_INCLUDE_DIRS}")
     message("PRIMESIEVE_LIBRARIES: ${PRIMESIEVE_LIBRARIES}")
+endif()
+
+message("WITH_FLINT: ${WITH_FLINT}")
+if (WITH_FLINT)
+    message("FLINT_INCLUDE_DIRS: ${FLINT_INCLUDE_DIRS}")
+    message("FLINT_LIBRARIES: ${FLINT_LIBRARIES}")
 endif()
 
 message("WITH_ARB: ${WITH_ARB}")
--- a/benchmarks/series_expansion_sincos_flint.cpp
+++ b/benchmarks/series_expansion_sincos_flint.cpp
@@ -1,6 +1,7 @@
 #include <symengine/symengine_config.h>
 
-#ifdef HAVE_SYMENGINE_PIRANHA
+#ifdef HAVE_SYMENGINE_FLINT
+#include <symengine/series_flint.h>
 
 #include <iostream>
 #include <chrono>
@@ -13,7 +14,10 @@
 using SymEngine::Basic;
 using SymEngine::Symbol;
 using SymEngine::symbol;
+using SymEngine::integer;
+using SymEngine::add;
 using SymEngine::mul;
+using SymEngine::pow;
 using SymEngine::sin;
 using SymEngine::cos;
 using SymEngine::RCP;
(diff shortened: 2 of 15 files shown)


certik:
Great job. This looks good to me at first sight.

symengine/series_flint.cpp
```diff
@@ -0,0 +1,172 @@
...
+        hash = hash * 101  +  *s++;
+    }
+    return hash;
+}
+
+std::size_t URatPSeriesFlint::__hash__() const {
+    std::size_t seed = URATPSERIESPIRANHA;
+    hash_combine(seed, var_);
+    hash_combine(seed, degree_);
+    hash_combine(seed, str_hash(p_.to_string().c_str(), seed));
```
certik: So in https://github.com/symengine/symengine/blob/3f0c0947ae09bac9545334a6903d9d72c7043a52/symengine/symbol.cpp#L14 we use:
```c++
std::hash<std::string> str_hash;
hash_combine(seed, str_hash(p_.to_string()));
```
is there a reason to use a different hash here?

rwst:
Um, I get only 1370ms vs 1600ms (URatP-Piranha), not even 20% speedup with the `sincos-1000`. Callgrind will show.

CMakeLists.txt
```diff
@@ -150,6 +155,19 @@ set(WITH_MPC no
...
+    include_directories(${FLINT_INCLUDE_DIRS})
+    set(LIBS ${LIBS} ${FLINT_TARGETS})
+    set(HAVE_SYMENGINE_FLINT True)
+    set(PKGS ${PKGS} "FLINT")
+
+    # Workaround for https://github.com/fredrik-johansson/arb/issues/24
+    include_directories(${FLINT_INCLUDE_DIRS}/flint)
+
+    set(WITH_MPFR yes)
+endif()
```
isuruf: If this block is after `WITH_ARB` then there's no need of code duplication.

symengine/series_flint.cpp
rwst: There is no function for getting a hash in Flint so I convert to string.

rwst: Ah you mean the extra `c_str()` and seed? Will try without.

certik: Yes, that's what I meant.

CMakeLists.txt
certik: I've reworked how this is handled in #757, it works great. So don't worry about it for now, just make sure it works. Once it gets in, I'll then merge it with my PR.

certik:
@rwst how does it compare to your series benchmarks that you have on github? Is there some overhead in SymEngine? If so, we need to fix it.

rwst:
It's worse because of Flint's creation of the two series `sin(x)` and `cos(x)` as explained above, which takes most of the time.

certik:
There is `series_sin` defined here: https://github.com/symengine/symengine/blob/3f0c0947ae09bac9545334a6903d9d72c7043a52/symengine/series.h#L279, and then you redefine it here: https://github.com/symengine/symengine/pull/780/files#diff-16c8b61b394d518f97214d67f2372e7eR65

But you don't use virtual methods, how does it work? Which method will be called? 

Also, is the first template implementation faster than the one in flint?

isuruf:
@certik, that's the beauty of CRTP. Since we get the class `Series` as a template argument, when you call `Series::series_sin`, it will call the static method in `Series` class. If it was not defined in `Series` class as in Piranha, it will call the static method in the parent of `Series` class

certik:
So in this PR, it will call the Flint's implementation, but if we remove `series_sin`, it will call our generic implementation?

isuruf:
@certik, yes.
Additional info: If you remove `series_cosh`, the generic implementation will call Flint implementation's `series_exp` and `series_invert` instead of the generic implementation.

certik:
Ah I see, nice. That's very cool, I like this design.

rwst:
I made the other experiment and tried `sin(x+x^2)*cos(x+x^2), N=1000` and surprise: Flint 1.7s, Pari 12s, Piranha `URatP` is not finished after 5 minutes... so I scaled down to `N=300`: Flint 0.071s, Pari 0.37s, Piranha `URatP` 11.1s. These are worlds apart.

isuruf:
URatP is slower because of clearing pow caches

isuruf:
Solution would be to clear it only if a higher precedence is called. 
CC @bluescarni

rwst:
I will replace `sin(x)*cos(x)` with `sin(x+x^2)*cos(x+x^2)` in the benchmark and make two separate files for flint and piranha. It makes no sense to have both in one.

rwst:
There is still the wider issue that affects all packages, even Pari: that it is faster to substitute if the argument is small, e.g. Pari:
```
? sin(x+x^2)*cos(x+x^2)
...
? ##
  ***   last result computed in 361 ms.
? subst(sin(x)*cos(x),x,x+x^2)
...
? ##
  ***   last result computed in 76 ms.
```
So someone will have to patch Pari and Flint as well with that optimization (first implemented in ring series).

bluescarni:
@rwst @isuruf I will be able to take a look later today. I would like to understand exactly what is going on, I expect flint to be faster but not by that amount.

rwst:
I would prefer to add optimizations and the integration in the `series()` function in another PR.

benchmarks/series_expansion_sincos_flint.cpp
```diff
@@ -0,0 +1,48 @@
...
+    //std::cout << *res[N-1] << std::endl;
+    std::cout
+        << std::chrono::duration_cast<std::chrono::milliseconds>(t2-t1).count()
+        << "ms" << std::endl;
+
+    return 0;
+}
+#else
+int main() {}
+#endif
```
certik: In here I would simply not build the benchmark if Flint is not installed. We can fix this after the PR is in.

benchmarks/series_expansion_sincos_piranha.cpp
```diff
@@ -39,7 +44,5 @@ int main(int argc, char* argv[])
     return 0;
 }
 #else
-
-int main(int, char*[]) {}
-
+int main() {}
 #endif
```
certik: The same here.

symengine/series_flint.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+unsigned int str_hash(const char* s, unsigned int seed = 0)
+{
+    unsigned int hash = seed;
+    while (*s)
+    {
+        hash = hash * 101  +  *s++;
+    }
+    return hash;
+}
```
certik: This function is not used, so you can remove it.

certik:
Just remove `str_hash` and it is +1 to merge from me if tests pass.

rwst:
To be continued.

fredrik-johansson:
Right now `sin(x+x^2+O(x^1000))` with Flint is slow, because it doesn't have a basecase algorithm for short input. You can compare with `exp(x+x^2+O(x^1000))`, which does have a basecase and is 700 times faster. Adding a similar basecase for sin/cos in on the todo list...

fredrik-johansson:
I've fixed this problem in flint -- `sin(x+x^2+O(x^1000))` takes 0.003 seconds now.

Some other functions still need to be optimized this way.

certik:
Thanks Fredrik! Which Flint commits do that?

fredrik-johansson:
The latest commit 57176eb1c02632340076a5f4ad6a69911824be5c in fredrik-johansson/flint2.git. I expect that Bill Hart will merge it in a moment.

certik:
Here is the link: https://github.com/fredrik-johansson/flint2/commit/57176eb1c02632340076a5f4ad6a69911824be5c

fredrik-johansson:
Thanks. It's in wbhart/flint2 now.

certik:
@fredrik-johansson according to your opinion now, if we expand sin or cos using Flint, does it now run at optimal speed? Or is there still space for improvement?

fredrik-johansson:
There are still some improvements that can be done in Flint, but we're talking about very small constant factor overhead (< 2).

certik:
Ok, perfect.

fredrik-johansson:
Actually, it's not going to be optimal for certain very sparse inputs, like
x + x^100 or indeed even x^100. You can always add more special cases for
such things. I think this is of somewhat limited practical utility, though.

certik:
Ok. How about multivariate polynomials? We've seen huge speedups in Piranha compared to Flint for multivariate polynomials, especially sparse. For single variate ones, Flint seems to be faster, as Piranha is not optimized for those.

fredrik-johansson:
There's no code in Flint for multivariate polynomials. There have been some
plans, but no one to work on them.

certik:
Nemo can do them.

fredrik-johansson:
Nemo has (dense) univariate polynomials, which allows you to create
polynomials of polynomials. This is fine for representing dense
multivariate polynomials in a small number of variables.

Fredrik

PR #781, opened 2016-01-26
Constants in C wrappers

abinashmeher999:
I have experimented on how the interface for the SymEngine's constants from C should be.

The logic was the `basic_struct*` that we obtain from these functions should not be available for referencing to any other `basic_struct`, i.e. it should be `const`. Some changes needed to be made to support that.

@certik @isuruf Please have a look.

### DIFF
--- a/symengine/cwrapper.cpp
+++ b/symengine/cwrapper.cpp
@@ -66,6 +66,46 @@ void basic_free_heap(basic_struct *s)
     delete s;
 }
 
+void basic_const_set(basic s, char* c)
+{
+    s->m = SymEngine::constant(std::string(c));
+}
+
+void basic_const_zero(basic s)
+{
+    s->m = SymEngine::zero;
+}
+
+void basic_const_one(basic s)
+{
+    s->m = SymEngine::one;
+}
+
+void basic_const_minus_one(basic s)
+{
+    s->m = SymEngine::minus_one;
+}
+
+void basic_const_I(basic s)
+{
+    s->m = SymEngine::I;
+}
+
+void basic_const_pi(basic s)
+{
+    s->m = SymEngine::pi;
+}
+
+void basic_const_E(basic s)
+{
+    s->m = SymEngine::E;
+}
+
+void basic_const_EulerGamma(basic s)
+{
+    s->m = SymEngine::EulerGamma;
+}
+
 TypeID basic_get_type(const basic s) {
     return static_cast<TypeID>(s->m->get_type_code());
 }
--- a/symengine/cwrapper.h
+++ b/symengine/cwrapper.h
@@ -84,6 +84,19 @@ void basic_free_stack(basic s);
 basic_struct* basic_new_heap();
 void basic_free_heap(basic_struct *s);
 
+//! Use these functions to get the commonly used constants as basic.
+//! Assigns to s a SymEngine constant with name c
+void basic_const_set(basic s, char* c);
+
+void basic_const_zero(basic s);
+void basic_const_one(basic s);
+void basic_const_minus_one(basic s);
+void basic_const_I(basic s);
+
+void basic_const_pi(basic s);
+void basic_const_E(basic s);
+void basic_const_EulerGamma(basic s);
+
 //! Assign value of b to a.
 void basic_assign(basic a, const basic b);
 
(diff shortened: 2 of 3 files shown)


isuruf:
I don't understand.
Why not have a `void basic_const_pi(basic s)`

abinashmeher999:
The motivation was to make `basic`/`basic_struct*` constant. If it is constant then the above function can't be used. The `basic_struct*` was made constant so that it can't be pointed to something else.

symengine/tests/cwrapper/test_cwrapper.c
```diff
@@ -383,6 +383,26 @@ void test_subs() {
...
+    basic p, z;
+    basic_new_stack(p);
+    basic_new_stack(z);
+
+    const basic_struct *pi = basic_const_pi();
+    const basic_struct *zero = basic_const_zero();
+
+    integer_set_si(z, 0);
+
+    //TODO: add test for pi
```
abinashmeher999: @isuruf any suggestions on how to test for `pi`?

isuruf:
@abinashmeher999, if the user wants to get a const value, the user can write a small function to achieve that.
```C
const basic func() {
   basic b = basic_new_heap()
   basic_const_pi(b);
   return b;
}
```
Problem with your approach is that,
1. you are allocating in the heap.
2. this function is different from others which might surprise the user.

abinashmeher999:
You are right. I didn't think of that. Only minor problem that the user might face is freeing it. Because the free functions take `basic_struct *` as an argument. That can be solved by changing the function prototype to take `const basic_struct*`.

Should I proceed with this apporach then
```c
// It will be first initialized with basic_new_stack or basic_new_heap
basic_const_<name>(basic_struct *)  // assigns the symengine constant to the variable
basic_free_stack(const basic_struct *)
basic_free_heap(const basic_struct *) //in case the user decides to obtain the value as a const and later wants to free it
```

isuruf:
How can you take in a `const` argument to a free function and then change it?

isuruf:
I think it should just be `basic_const_pi(basic b)`

abinashmeher999:
Let's say the user did this
```c
const basic func() {
   basic b = basic_new_heap()
   basic_const_pi(b);
   return b;
}
```
How will the user free it?

isuruf:
Right. So, we have to scrap that idea.

symengine/tests/cwrapper/test_cwrapper.c
```diff
@@ -383,6 +383,26 @@ void test_subs() {
...
+    SYMENGINE_C_ASSERT(basic_eq(z, zero));
+
+    basic_free_stack(p);
+    basic_free_stack(z);
+
+    basic_free_const(pi);
```
isuruf: You can't do this because `pi` is a const pointer

abinashmeher999: Yes, I will correct it.

abinashmeher999:
@isuruf I have updated the PR.

symengine/cwrapper.cpp
```diff
@@ -41,6 +41,8 @@ extern "C" {
 
 struct CRCPBasic {
     SymEngine::RCP<const SymEngine::Basic> m;
+    CRCPBasic() {};
```
isuruf: You don't need this anymore. I think this is what is causing the error, because an RCP object can't be null in Teuchos

abinashmeher999: I see. Let me check.

*Regards,*

*Abinash Meher*

abinashmeher999: The error was because of a silly mistake that I made in the tests. I fixed them now.

isuruf:
+1 to merge

symengine/cwrapper.h
```diff
@@ -84,6 +84,19 @@ void basic_free_stack(basic s);
 basic_struct* basic_new_heap();
 void basic_free_heap(basic_struct *s);
 
+//! Use these functions to get the commonly used constants as basic.
+//! Assigns to s a SymEngine constant with name c
+void basic_const_set(basic s, char* c);
```
certik: Can you document here, that this function creates a new `SymEngine::constant` from a **copy** of the string in `c`, thus the caller is free to use `c` afterwards, and also the caller must free `c`.

certik:
I think that this looks good, so I am merging it. But @abinashmeher999 if you could add the comment I mentioned in a new PR, that would be great.

PR #784, opened 2016-01-28
speed up several Piranha series functions

rwst:
The sincos benchmark shows an 8x speedup at N=1000 with this.

### DIFF
--- a/symengine/series.h
+++ b/symengine/series.h
@@ -222,8 +222,10 @@ public:
         if (s == var) {
             //! fast atan(x)
             short sign = 1;
+            Poly monom(var), vsquare(var*var);
             for (unsigned int i = 1; i < prec; i += 2, sign *= -1) {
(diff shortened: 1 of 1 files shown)


certik:
+1

symengine/series.h
```diff
@@ -222,8 +222,10 @@ class SeriesBase : public SeriesCoeffInterface {
         if (s == var) {
             //! fast atan(x)
             short sign = 1;
+            Poly monom(var), vsquare(var*var);
             for (unsigned int i = 1; i < prec; i += 2, sign *= -1) {
-                res_p += Series::pow(var, i, prec) * (Coeff(sign) / Coeff(i));
+                res_p += monom * (Coeff(sign) / Coeff(i));
+                monom *= vsquare;
```
isuruf: Shouldn't this be done in a truncated manner?

rwst: The degree of `monom` and `res_p` is always `< prec`.

isuruf:
+1

PR #785, opened 2016-01-28
further fixes to series

rwst:
This includes
 - add tests for series interface
 - speed up flint sin/cos by 2.5x
 - fix: make `SeriesInterface` members public
 - fix bugs in `series_flint/piranha.as_basic()`
 - remove conditional code in test files

### DIFF
--- a/symengine/series_flint.h
+++ b/symengine/series_flint.h
@@ -61,6 +61,8 @@ public:
         if (not s.get_coeff(0).is_zero()) {
             throw std::logic_error("sin(const) not Implemented");
         }
+        if (s.degree() > 8)
+            return fp_t(sin_series(s, prec));
         //! fast sin(x)
         fp_t res_p, monom(s), ssquare;
         ssquare = mullow(s, s, prec);
@@ -79,10 +81,11 @@ public:
     }
 
     static inline fp_t series_cos(const fp_t &s, const fp_t& var, unsigned int prec) {
-//        return fp_t(cos_series(s, prec));
         if (not s.get_coeff(0).is_zero()) {
             throw std::logic_error("cos(const) not Implemented");
         }
+        if (s.degree() > 8)
+            return fp_t(cos_series(s, prec));
         fp_t res_p, ssquare, monom;
         ssquare = mullow(s, s, prec);
         monom = ssquare;
--- a/symengine/series_flint.h
+++ b/symengine/series_flint.h
@@ -6,6 +6,7 @@
 #include <symengine/expression.h>
 
 #ifdef HAVE_SYMENGINE_FLINT
+#include "flint.h"
 #include <flint/fmpq_polyxx.h>
 
 namespace SymEngine {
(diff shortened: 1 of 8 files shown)


symengine/series_flint.h
```diff
@@ -56,14 +57,48 @@ class URatPSeriesFlint : public SeriesBase<fp_t, flint::fmpqxx, URatPSeriesFlint
...
+        prod.set_one();
+        for (unsigned int i = 1; i <= prec / 2; i++) {
+            const short j = 2 * i;
+            if (i != 0)
+                prod /= flint::fmpzxx(1 - j);
+            prod /= flint::fmpzxx(j);
+            res_p += monom * prod;
+            monom *= ssquare;
+        }
+        return res_p;
```
isuruf: Is this faster than flint's `cos_series`?

symengine/series_flint.h
```diff
@@ -56,14 +57,48 @@ class URatPSeriesFlint : public SeriesBase<fp_t, flint::fmpqxx, URatPSeriesFlint
...
+        res_p.set_one();
+        flint::fmpqxx prod;
+        prod.set_one();
+        for (unsigned int i = 1; i <= prec / 2; i++) {
+            const short j = 2 * i;
+            if (i != 0)
+                prod /= flint::fmpzxx(1 - j);
+            prod /= flint::fmpzxx(j);
+            res_p += monom * prod;
+            monom *= ssquare;
```
isuruf: Is this a truncated multiplication?

certik:
Is the "overloaded" sin series implementation faster than the generic one?
What makes it faster?

rwst:
The code in sin/cos before this PR called Flint's implementation. This one now is an adaptation of our generic code. I did not use the generic code because 1. it's awkward or maybe impossible to write generic code that can handle native flint polys, and 2. I didn't want to write a wrapper class. Also it is quite possible that it turns out that for optimal results we need code that switches from this algorithm to the native flint implementation when the argument reaches a certain degree.

So nothing is final atm.

certik:
I guess what I don't understand is how the generic code can handle Piranha polys, but not Flint polys.

symengine/tests/basic/test_series.cpp
```diff
@@ -0,0 +1,40 @@
...
+{
+    RCP<const Symbol> x = symbol("x"), y = symbol("y");
+    auto ex = div(integer(1), add(integer(1), x));
+    auto ser = series(ex, x, 10);
+    std::cerr << ser->as_basic()->__str__() << std::endl;
+
+    REQUIRE(rcp_static_cast<const Number>(ser->get_coeff(7))->is_minus_one());
+    REQUIRE(rcp_static_cast<const Number>(ser->as_dict()[8])->is_one());
+    REQUIRE(ser->as_basic()->__str__()=="1 - x + x**2 - x**3 + x**4 - x**5 + x**6 - x**7 + x**8 - x**9");
+}
```
certik: You need to add ifdef piranha around here to fix the Travis failures.

certik:
The changes are ok with me. We just need to make sure tests pass.

certik:
Thanks for working on this @rwst, it's very exciting. It is an iterative process and it will take us a while to make sure we use both Flint and Piranha optimally. I think it is a great start what we already have.

rwst:
There is already a sincos speedup with *truncated multiplication for the small arg `x+x^2` of 20%.

The following are results for `sin(arg)` with argument a polynomial `x+x^2+x^3+...` up to degree `M`, i.e. `M=2` is `sin(x+x^2)` as in the original `sincos_flint` benchmark. All tests with `N=1000`.
```
       current impl.     native flint
M=2        0.25             0.77
M=10      0.88             0.81
M=100    1.3               0.82
M=1000  1.8               0.91
```
So the 2.5x speedup with this patch only lasts until `sin(x+x^2+..,+x^10)`.

certik:
Interesting. Maybe we can somehow allow both implementations to be used by the user.

rwst:
>I guess what I don't understand is how the generic code can handle Piranha polys, but not Flint polys.
Two issues: 1. flint functions return a complicated templated poly expression type whch different from the poly type and, 2. intialization of the poly type with e.g. `Poly pol(0)` or any other `int` does not initialize the poly to a constant but just reserves that much memory

certik:
Ah, so the general wrapper compiles, but doesn't even work, if `pol(0)` doesn't initialize it?

So if we wanted to use it, we would need to create a Flint C++ wrapper.

rwst:
It doesn't even compile due to the flint return type. Yes, a wrapper in the future, and additional casts in the generic code (which cost nothing in flint as far as I understand, it's a quite interesting concept, see ch.64 of the documentation, not online).

rwst:
Thanks for the review.

certik:
These changes are +1 from me. Next time please wait for a final review if you push more patches. It's a good habit.

rwst:
You're right. That went beyond the 'obvious fix' threshold.

PR #790, opened 2016-02-05
Mertens Function

malayaleecoder:
Added Mertens Function, see https://en.wikipedia.org/wiki/Mertens_function for more details.
Please review and suggest me further improvements!

### DIFF
--- a/symengine/ntheory.cpp
+++ b/symengine/ntheory.cpp
@@ -1498,11 +1498,8 @@ int mobius(const Integer &a)
     }
 }
 
-int mertens(const int long a)
+int mertens(const unsigned long a)
 {
-    if (a <= 0) {
-        throw std::runtime_error("mertens : Mertens function is only defined for positive integers!");
-    }
     unsigned mertens = 0;
     for (unsigned i = 1; i <= a; i++) {
         mertens += mobius(*integer(i));
--- a/symengine/ntheory.cpp
+++ b/symengine/ntheory.cpp
@@ -1498,9 +1498,9 @@ int mobius(const Integer &a)
     }
 }
 
-int mertens(const unsigned long a)
+long mertens(const unsigned long a)
 {
-    unsigned long mertens = 0;
+    long mertens = 0;
     for (unsigned long i = 1; i <= a; i++) {
         mertens += mobius(*integer(i));
     }
--- a/symengine/ntheory.cpp
+++ b/symengine/ntheory.cpp
@@ -1500,8 +1500,8 @@ int mobius(const Integer &a)
 
 int mertens(const unsigned long a)
 {
-    unsigned mertens = 0;
-    for (unsigned i = 1; i <= a; i++) {
+    unsigned long mertens = 0;
+    for (unsigned long i = 1; i <= a; i++) {
         mertens += mobius(*integer(i));
     }
     return mertens;
--- a/symengine/ntheory.cpp
+++ b/symengine/ntheory.cpp
@@ -1498,4 +1498,15 @@ int mobius(const Integer &a)
     }
 }
 
+int mertens(const int long a)
+{
+    if (a <= 0) {
+        throw std::runtime_error("mertens : Mertens function is only defined for positive integers!");
+    }
+    unsigned mertens = 0;
+    for (unsigned i = 1; i <= a; i++) {
+        mertens += mobius(*integer(i));
+    }
+    return mertens;
+}
 } // SymEngine
--- a/symengine/ntheory.h
+++ b/symengine/ntheory.h
@@ -176,7 +176,7 @@ void powermod_list(std::vector<RCP<const Integer>> &pows, const RCP<const Intege
 int mobius(const Integer &a);
 //Mertens Function
 //mertens(n) -> Sum of mobius(i) for i from 1 to n
-int mertens(const unsigned long a);
+long mertens(const unsigned long a);
 }
 #endif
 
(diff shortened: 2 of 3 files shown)


Sumith1896:
Do add tests too.

symengine/ntheory.cpp
```diff
@@ -1498,4 +1498,15 @@ int mobius(const Integer &a)
     }
 }
 
+int mertens(const Integer &a){
+    if(a.as_int() <= 0){
+        throw std::runtime_error("mertens: Integer <= 0");
```
Sumith1896: Improve the message

symengine/ntheory.cpp
```diff
@@ -1498,4 +1498,15 @@ int mobius(const Integer &a)
...
+    }
+    unsigned mertens = 0;
+    for(unsigned i = 1; i <= a.as_int(); i++){
+        RCP<const Integer> j = integer(i);
+        mertens = mertens + mobius(*j);
```
certik: You can write these two lines as:
```
mertens += mobius(*integer(i));
```

symengine/ntheory.cpp
```diff
@@ -1498,4 +1498,15 @@ int mobius(const Integer &a)
     }
 }
 
+int mertens(const Integer &a){
+    if(a.as_int() <= 0){
```
certik: Add space after `if` and before `{`.

symengine/ntheory.cpp
```diff
@@ -1498,4 +1498,15 @@ int mobius(const Integer &a)
...
+        throw std::runtime_error("mertens: Integer <= 0");
+    }
+    unsigned mertens = 0;
+    for(unsigned i = 1; i <= a.as_int(); i++){
```
certik: Add space after `if` and before `{`.

isuruf:
Is this a widely used function? I don't see it in SymPy nor Sage

malayaleecoder:
@Sumith1896 @certik: Any more improvisations required? And why is Appveyor and Travis CI still showing that the build is in progress?

malayaleecoder:
@isuruf : I don't think it is widely used. I heard about this function from it's related conjecture, https://en.wikipedia.org/wiki/Mertens_conjecture.

symengine/ntheory.cpp
```diff
@@ -1498,4 +1498,15 @@ int mobius(const Integer &a)
     }
 }
 
+int mertens(const Integer &a)
+{
+    if(a.as_int() <= 0) {
```
certik: Space after `if`.

symengine/ntheory.cpp
```diff
@@ -1498,4 +1498,15 @@ int mobius(const Integer &a)
...
+        throw std::runtime_error("mertens : Mertens function is only defined for positive integers!");
+    }
+    unsigned mertens = 0;
+    for(unsigned i = 1; i <= a.as_int(); i++) {
```
certik: space after `for`.

certik:
Given the way the function is implemented, it only works if the argument fits into an int. Why not to then change the argument from `const Integer &a` to `const long int a`.

And change the tests to just `REQUIRE(mertens(31) == -4);`.

It's not widely used, but I think there is no harm in having it, as long as it is well tested. Also can you add a test for the negative integer? Test that it raises an exception.

malayaleecoder:
@certik : Do verify and let me know about any improvements :)

Sumith1896:
Looks good. You can squash your commits.

malayaleecoder:
@Sumith1896 : Done. Please check!

symengine/ntheory.cpp
```diff
@@ -1498,4 +1498,15 @@ int mobius(const Integer &a)
     }
 }
 
+int mertens(const int long a)
```
isuruf: This can just be `unsigned long` and then you don't have to check for negative values

malayaleecoder:
@isuruf : If I make it unsigned long, should I add tests where negative values throw an exception?

certik:
If you make it unsigned, I don't think you can assign a negative value to it, can you? So that seems like a nice solution.

isuruf:
+1

certik:
This looks very simple and nice. +1 from me as well.

symengine/ntheory.cpp
```diff
@@ -1498,4 +1498,12 @@ int mobius(const Integer &a)
     }
 }
 
+int mertens(const unsigned long a)
+{
+    unsigned mertens = 0;
+    for (unsigned i = 1; i <= a; i++) {
```
isuruf: Either all of `a, mertens, i` should be `unsigned` or `unsigned long`

certik: I missed that, yes.

malayaleecoder: Sorry for missing that.

malayaleecoder:
@isuruf @certik : Should I squash the commits again and make it final?

symengine/ntheory.cpp
```diff
@@ -1498,4 +1498,12 @@ int mobius(const Integer &a)
     }
 }
 
+int mertens(const unsigned long a)
+{
+    unsigned long mertens = 0;
```
Sumith1896: summation of `mobius` can be negative. `unsigned long` is not the right type.

srajangarg: Your function returns `int` but inside, you have returned an `unsigned long`, the test cases passes only because the answer is `int` casted.

malayaleecoder:
I think this should fix it. Please have a look.

malayaleecoder:
@certik : Is there any more improvements to be implemented? Please check.

certik:
@malayaleecoder it looks good to me. Would you mind squashing the commits? The final change is very simple, I think a single commit is fine.

malayaleecoder:
@certik I am unable to squash the commits in this branch. So I am closing this and am creating a new PR.

certik:
That works. Thanks!

PR #791, opened 2016-02-06
series_flint: take advantage of newest flint

rwst:
Timings of `benchmark/series_sincos_flint` are:
 * before (using old Flint): 1620 ms
 * before (using own code): 605 ms
 * now (using flint git master, removing own code): 125 ms


certik:
Nice, good job! This is +1 from me.

rwst:
Thanks.

Issue #788, opened 2016-02-01
minor cmake issue with setting flint=no


rwst:
Setting `-DWITH_FLINT=no` with cmake will not set it to No if `WITH_ARB=yes` but it should set both to No.

Discussion in #795 showed that using `No` as default is indistinguisable to `No` as commanded. I think a solution would be to set the default to `Maybe`.

malayaleecoder:
@rwst : I would like to work on this issue. I am very new to symengine. Please guide me through.

rwst:
What do you not understand?

malayaleecoder:
@rwst : Which are the files to be modified with the above changes?

rwst:
The files that must be modified are our cmake configuration files, see https://cmake.org/

malayaleecoder:
@rwst : Very sorry that I could not understand correctly what You are trying to convey. When I run "cmake ." I see that "WITH_ARB no". Is it this what you are referring?

rwst:
SymEngine optionally uses arb and flint. Arb depends on flint and includes it. So, if all is No and WITH_ARB is then set to yes, WITH_FLINT is set to Yes too. That works already. Now with both Yes WITH_FLINT is set to No (because the user deinstalled it) then WITH_ARB should also be back at No. Your task is to implement that.

CodeMaxx:
@malayaleecoder are you still working on this? Oth

malayaleecoder:
@CodeMaxx : No, you can go ahead with it.

CodeMaxx:
@rwst is `-DWITH_ARB=yes` by default ?

rwst:
It depends on if arb is installed on the system.

rwst:
Invalid as per discussion in #795

PR #795, opened 2016-02-06
set WITH_ARB to false when WITH_FLINT is false

CodeMaxx:
@rwst @isuruf For #788

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,6 +166,9 @@ if (WITH_FLINT)
     include_directories(${FLINT_INCLUDE_DIRS}/flint)
 
     set(WITH_MPFR yes)
+else()
+    set(WITH_ARB no)
+    set(HAVE_SYMENGINE_ARB False)
 endif()
 
 if (WITH_ARB)


isuruf:
This won't work. If a user only sets `WITH_ARB=yes` and doesn't set `WITH_FLINT` this will make `WITH_ARB=no`

CodeMaxx:
Yeah I had thought of that ... but how do I detect if the user sets `WITH_FLINT=no` or hasn't set it at all?

rwst:
If the user doesn't give `-DWITH_FLINT=` it is set to the previously used value. If that is unknown I think the default is No. Note however that the previous value may be Yes although Flint may no longer be installed on the system. So setting `-DWITH_FLINT=yes` is not more than giving an intention with the final value depending on installation status.

CodeMaxx:
Isn't `WITH_FLINT` set to no [here](https://github.com/symengine/symengine/blob/master/CMakeLists.txt#L139) even if the previously used value was yes ?

rwst:
That was really easy to answer: https://cmake.org/cmake/help/v3.3/command/set.html#set-cache-entry
"Set the given cache <variable> (cache entry). Since cache entries are meant to provide user-settable values this does not overwrite existing cache entries by default. Use the FORCE option to overwrite existing entries."

CodeMaxx:
@isuruf @rwst this is what we are aiming for right?

|         | Flint_yes        | Flint_no | Flint_x |
|---------|------------------|----------|---------|
| ARB_yes | yes              | no       | yes     |
| ARB_no  | Flint_yes ARB_no | No       | No      |
| ARB_x   | Flint_yes ARB_no | No       | No      |

`yes` implies both are yes. `x` implies user has not set this option.

rwst:
Yes that's right.

CodeMaxx:
I still don't understand how to I detect the difference between `WITH_FLINT` set to `no` and `WITH_FLINT` not set at all(previously set value was `no`)? We need different configurations for these two scenarios according the table in my prev comment.

rwst:
You mean reading the cache before using the set command?

CodeMaxx:
No. My question is that if in case `WITH_ARB=yes` but `WITH_FLINT` is not specified by the user, then the final configuration will depend on the value of `WITH_FLINT` from the cache(which is the value the user specified in a previous run) i.e.
|         | flint_prev_no   | flint_prev_yes    |
|---------|-----------------|-------------------|
| arb_yes | arb_no flint_no | arb_yes flint_yes |
So what I want to know is if the user has set the value of `WITH_FLINT`

rwst:
Actually, if the user sets `WITH_ARB=yes` cmake should check again if flint is installed as well and set `WITH_FLINT` accordingly.

isuruf:
@rwst, problem is that you cannot make the distinction of the first run or a subsequent run.
When `WITH_ARB=yes` is set by the user at first. `WITH_FLINT=no` is there because it is the default. Then since `WITH_ARB=yes`, `WITH_FLINT=yes` is set.

When the user sets `WITH_FLINT=no` on the command line this is the same as the first case, so you cannot differentiate between the two

rwst:
I see. I think a solution would be to set the default to `Maybe`---I just confirmed this is possible. What do you think?

CodeMaxx:
@rwst Can you give a reference to the source.... What does `Maybe` actually mean in this case. Also what is the syntax to detect it.. `if(WITH_FLINT=Maybe)` or something else?

CodeMaxx:
Also I tried this
```
set(WITH_FLINT Maybe
    CACHE BOOL "Build with Flint" FORCE)
set(HAVE_SYMENGINE_FLINT False)
```
It treats `Maybe` are `true`

rwst:
Of course you need `STRING` instead of `BOOL` in the `SET`, see the cmake doc link above.

CodeMaxx:
The Maybe solution doesn't work.
The cache is set to Maybe only on the first run... otherwise the variable retains the value of the previous run.
if I use FORCE then the cache remains Maybe no matter what the user specifies.

I tried to search for alternate methods but couldn't find any. Just [this something](http://stackoverflow.com/questions/27350560/how-to-tell-whether-cmake-used-initial-value-for-an-option) from stackexchange but I guess thats not the solution we are looking for.

rwst:
That nails it. Many thanks for your efforts. The issue can be considered
invalid.

CodeMaxx:
Cool...closing the PR then.

CodeMaxx:
@rwst Shouldn't we atleast turn off ARB when FLINT is off?

isuruf:
@CodeMaxx, that's what you were trying to do and realized you cannot right?

CodeMaxx:
We can do that when user specifies Arb=yes and Flint=no... But if user gives only Arb=yes and doesn't specify Flint,then whether Arb is used or not will depend on the value of Flint from the previous run.

CodeMaxx:
@isuruf @rwst What is the final decision then?

rwst:
The original problem in #788 was that I gave Flint=No and expected Arb to be set to No from it, as well. Would this be distinguishable from Flint=No cached and the user sets Arb=yes? If so, please reopen #788

CodeMaxx:
No, it won't be distinguishable..... so should I close this PR?

rwst:
Yes, I think so.

PR #796, opened 2016-02-08
catch all flint aborts with asserts

rwst:
part of #793


rwst:
No travis?

certik:
This is +1 from me.

rwst:
Thanks.

Issue #672, opened 2015-12-04
improve sin_table, add cos, tan


rwst:
Recently, I improved these in Pynac to include all values expressible as roots of roots. We could have that in SymEngine too:
https://github.com/pynac/pynac/blob/master/ginac/inifcns_trans.cpp#L421
https://github.com/pynac/pynac/blob/master/ginac/inifcns_trans.cpp#L581
https://github.com/pynac/pynac/blob/master/ginac/inifcns_trans.cpp#L735

certik:
Are those the ones and only ones that can be converted to closed form? Or are these the ones and only ones that can be converted to closed form expressible as a root of a root?

If I remember well, some of the values can be converted to three nested roots, four nested roots.

The other question is whether there is a noticeable slowdown, since you need to compare lots of cases now. I wonder if the if statements can be replaced by a hash table.

rwst:
These are the ones and only ones that can be converted to closed form expressible as a root of a root. I determined the characteristic polynomials of high precision values of `f(pi/n)` with `n` up to 144 IIRC. Of course, if I were to write a paper, I would recheck.

rwst:
No in a paper I would have to prove it even.

certik:
Interesting. Do you have some notes about the math? It would be nice to have a script to do this.

rwst:
Unfortunately no notes. It was not really difficult but I did many things by hand using Sage.

rwst:
However, their correctness of all of those implemented is tested to 1e-30 in Sage:
https://github.com/sagemath/sage/blob/master/src/sage/functions/trig.py#L73

isuruf:
What do you mean by a root of a root?

Does this count?
![image](https://cloud.githubusercontent.com/assets/5234427/11607505/2e2fc00e-9b6f-11e5-8311-0e88d26086bd.png)

rwst:
Well, then Sage lied:
```
sage: QQbar(sin(pi/120)).radical_expression()
1/2*sqrt(-sqrt(1/2*sqrt(sqrt(5) + 128*sqrt(3/8192*sqrt(5) + 15/8192) + 7) + 2) + 2)
```

rwst:
Solving the minimal polynomial with Sympy does not better. So if you really want to automatize this a better solver or simplifier is needed. Try yourself:
```
sage: QQbar(sin(pi/120))               
0.02617694830787315? + 0.?e-19*I
sage: _.minpoly()
[...]

PR #800, opened 2016-02-10
Added cos table

nishnik:
Addressing issue #672 .
A naive table for `cos`, the same way `sin_table` is implemented.
Novice question: Do we have methods to handle infinity, will be needed for `tan_table`.

### DIFF
--- a/symengine/constants.cpp
+++ b/symengine/constants.cpp
@@ -86,6 +86,13 @@ RCP<const Basic> sin_table[] = {
         zero, mC0, mC1, mC2, mC3, mC4, minus_one, mC4, mC3, mC2, mC1, mC0
     };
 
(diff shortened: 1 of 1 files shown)


certik:
You should always start with a test. I think this is already implemented. See here:

https://github.com/symengine/symengine/blob/18c3c617d60f1cdaf571a450bfe2169583485fca/symengine/functions.cpp#L436

But the tests seem to be there only for `sin`:

https://github.com/symengine/symengine/blob/18c3c617d60f1cdaf571a450bfe2169583485fca/symengine/tests/basic/test_functions.cpp#L1070

So write them for `cos` and see if it works. I think it might. Either way, we should add tests for this.

nishnik:
Okay! I will add tests for cos and send a PR.

PR #823, opened 2016-02-24
Misc parser

srajangarg:
Added some missing functionality to the parser, including multi argument function support. (and also some newly added functions)
Also added some tests to check if errors were being thrown correctly.
Also, the current master has trailing whitespaces, which I removed. Why are they creeping in (repeatedly), and why isn't travis reporting them?

### DIFF
--- a/symengine/parser.cpp
+++ b/symengine/parser.cpp
@@ -58,10 +58,10 @@ class ExpressionParser {
         {"asec", asec}, {"acsc", acsc}, {"acot", acot},
 
         {"sinh", sinh}, {"cosh", cosh}, {"tanh", tanh},
-        {"coth", coth}, // {"sech", sech}, {"csch", csch},
+        {"coth", coth}, {"sech", sech}, {"csch", csch},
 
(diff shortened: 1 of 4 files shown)


certik:
I think this looks great. Thanks.

I don't know why the whitespace check doesn't work, we need to fix it.

Sumith1896:
Looks good to me too.

Sumith1896:
Thanks @srajangarg 
Yes, we'll need to fix the whitespace thing.

CodeMaxx:
I reviewed the whitespace checking code. The code is correct. I believe this has something to do with the  position of the code in `test_travis.sh`

Sumith1896:
@CodeMaxx Possibly. Abhinav was trying the same here https://github.com/symengine/symengine/pull/827
Do send in a PR in case you fix it.

CodeMaxx:
@Sumith1896 I tried placing the code in different positions in the file but no success. Could it be a problem with Travis?

CodeMaxx:
Otherwise I think we should add a python script just like we had before.

PR #838, opened 2016-03-01
Working whitespace check

nishnik:
I googled a lot about this issue.
As a good practice it is advised to use [Git hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks).
The default `pre-commit` hook is sufficient for whitespace check.
We can write a bash or python script to activate `pre-commit` hook (It just needs `mv .git/hooks/pre-commit.sample .git/hooks/pre-commit`). *Need suggestions on that!*

And adding this: `git diff HEAD^1 --check` in travis will serve our purpose. As it will exit the travis build if it finds trailing whitespace in recently modified code.

[Instance](https://travis-ci.org/nishnik/symengine/builds/112823175) where travis build fails due to trailing whitespace.
[Instance](https://travis-ci.org/nishnik/symengine/builds/112830342) when it runs fine when there aren't any whitespace. One job fails because piranha has been updated. Unrelated to this PR.

### DIFF
--- a/bin/test_travis.sh
+++ b/bin/test_travis.sh
@@ -106,10 +106,4 @@ export LD_LIBRARY_PATH=$our_install_dir/lib:$LD_LIBRARY_PATH
 echo "Checking whether all header files are installed:"
 python $SOURCE_DIR/symengine/utilities/tests/test_make_install.py $our_install_dir/include/symengine/ $SOURCE_DIR/symengine
 
-# check trailing whitespace:
-if !  egrep " $" -nr --include=\*.{cpp,h,inc}  --exclude-dir=*{teuchos,/build/}* $SOURCE_DIR ; then
-    echo No trailing whitespace;
-else
-    exit -1;
-fi
-# TODO: Add similar grep checks for space after comma,, space after `if`, space between `)` and `{` also
+git diff HEAD@{50} --check #test for whitespaces
--- a/symengine/mp_wrapper.h
+++ b/symengine/mp_wrapper.h
@@ -10,7 +10,7 @@
 #endif
 
 #define SYMENGINE_UI(f) f ## _ui
-#define SYMENGINE_SI(f) f ## _si 
+#define SYMENGINE_SI(f) f ## _si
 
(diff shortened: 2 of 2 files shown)


abhinavagarwal07:
Seems good to me.

abhinavagarwal07:
@nishnik 
Please merge with latest master to get tests working.

nishnik:
ping @isuruf @certik

bin/test_travis.sh
```diff
@@ -106,10 +106,4 @@ export LD_LIBRARY_PATH=$our_install_dir/lib:$LD_LIBRARY_PATH
...
 python $SOURCE_DIR/symengine/utilities/tests/test_make_install.py $our_install_dir/include/symengine/ $SOURCE_DIR/symengine
 
-# check trailing whitespace:
-if !  egrep " $" -nr --include=\*.{cpp,h,inc}  --exclude-dir=*{teuchos,/build/}* $SOURCE_DIR ; then
-    echo No trailing whitespace;
-else
-    exit -1;
-fi
-# TODO: Add similar grep checks for space after comma,, space after `if`, space between `)` and `{` also
+git diff HEAD^1 --check #test for whitespaces
```
Sumith1896: What is the idea here? 
`remove the top commit and take a diff` as it looks to me.

nishnik: It doesn't remove the top commit. Just returns the `diff` introdued at it, i.e. the changes introduced in previous commit.
And `diff --check` is to check for trailing whitespaces.

> --check

    Warn if changes introduce whitespace errors.
    What are considered whitespace errors is controlled by core.whitespace configuration.
    By default, trailing whitespaces (including lines that solely consist of whitespaces) and a space character that is immediately followed by a tab character inside the initial indent of the line are considered whitespace errors.
    Exits with non-zero status if problems are found.

Sumith1896: This is incorrect.

Sumith1896:
@nishnik This misses out on some cases. Say you have a set of 10 commits, this checks only the top two. If the whitespace was introduced in the initial commits then they are missed out.

nishnik:
Oh! My bad!
I will look into this.

PR #858, opened 2016-03-07
Implement Error function

abhinavagarwal07:
#832 
Erf
- [x] basic functionality, diff etc

- [x] print/ parse

- [x] evals

- [x] subs

- [x]  ~~add more cases as in sympy~~ add `erf(-x) = -erf(x)`


isuruf:
Can you take a look at what SymPy does and implement all of the relevant cases?

abhinavagarwal07:
@isuruf there are `rewrite` methods for `erf` but they need `erfc` / `erfi` to implemented first.
`could_extract_minus` can be used to write `erf(-x)` as `-erf(x)`

What are other relevant cases ?

isuruf:
I think only writing `erf(-x)` as `erf(x)` is left here

abhinavagarwal07:
@isuruf http://www.wolframalpha.com/input/?i=erf(-x)
`erf(-x) = -erf(x)`
> `erf(-x)` as `erf(x)`

isuruf:
Sorry, that was a typo

thisch:
support for complex numbers would also be nice

isuruf:
For complex numbers, there's no implementation in C++ standard library. Maybe we could use Faddeeva package? It's a MIT licensed header file and an implementation file.

@abhinavagarwal07, as @rwst mentioned in #859, it's easy to implement the series of erf by generating the coefficients

rwst:
I think this is fine. Complex numerics should not hold up this PR when the same is missing with a lot of other functions, as well. What's more, there should be a discussion about makinig flint+arb default, as they are clearly superior to MPFR stuff.

PR #876, opened 2016-03-20
could_extract_minus handles x-y and y-x

vyasgiridhar:
#871 
Can be used for abs(y-x) == abs(x-y).

### DIFF
--- a/symengine/functions.cpp
+++ b/symengine/functions.cpp
@@ -152,21 +152,43 @@ bool could_extract_minus(const RCP<const Basic> &arg)
             return false;
     }
     else if (is_a<Add>(*arg)) {
+        int flag = 0;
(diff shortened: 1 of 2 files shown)


CodeMaxx:
Hi @vyasgiridhar Welcome to SymEngine! 
So there's a problem with your code. Though the issue didn't mention it properly but we are looking to differentiate between all such expressions as `x-y-z` and `z+y-x` and not just for the special case of `x-y`.

vyasgiridhar:
this works out?

CodeMaxx:
No. What exactly are you trying to do? You are getting segmentation faults as well.

vyasgiridhar:
Solved the segmentation faults.

CodeMaxx:
You should add more tests to check if your code works.

PR #884, opened 2016-03-24
revert commit 1617f970; Add::as_two_terms() needed by symengine.py

rwst:
I propose to fix compile problems in symengine.py by putting back the removed function


certik:
Looks like codecov works. :) 

What's the long term idea ---- do we want to support this function or not? If yes, then we need to write a test. If not, then we should remove it and fix the Python wrappers.

rwst:
I think we should support it---it's also in SymPy. I'll add a test.

certik:
Let's add a test, we can support it.

PR #898, opened 2016-03-31
Clang format final

srajangarg:
Discussions https://github.com/symengine/symengine/pull/882 and https://github.com/symengine/symengine/issues/866
The code coverage test can be ignored, for this PR.


certik:
This is +1, if tests pass.

Sumith1896:
+1 from me too. 
@isuruf Your views would be great too.

bin/travis_clang.sh
```diff
@@ -0,0 +1,30 @@
...
+    for FILE in $FILES; do
+
+        $CLANG $FILE | cmp  $FILE >/dev/null
+
+        if [ $? -ne 0 ]; then
+            echo "[!] INCORRECT FORMATTING! $FILE" >&2
+            RETURN=1
+        fi
+
+    done
```
certik: This script seems to only check the files changed against master. This seems fragile. Why not to check all the files, always?

srajangarg: We anyways have master all formatted up. That's why I thought it was the way to go. Checking all files doesn't hurt. Should I go ahead with thath?

Sumith1896: Yes, preferably.

certik:
I think this is good enough, let's merge this. Then we can improve the scripts once it is in master. There might be more things we'll have to iron out.

certik:
@srajangarg I've created #902 and #903 for the remaining issues.

certik:
I am still struggling setting up Vim the right way. I have asked here: http://stackoverflow.com/questions/36364338/

Issue #903, opened 2016-04-01
Improve bin/travis_clang.sh


certik:
As described here: https://github.com/symengine/symengine/pull/898#discussion-diff-58224489

srajangarg:
Can be closed!

certik:
Fixed by #906.

PR #904, opened 2016-04-01
Travi-clang checks all files. Linux version error removed

CodeMaxx:
Addresses #902 #903 
@certik @isuruf

### DIFF
--- a/bin/travis_clang.sh
+++ b/bin/travis_clang.sh
@@ -11,7 +11,7 @@ if [[ "${TRAVIS_OS_NAME}" == "linux" ]] && \
         exit 1
     fi
 
-    FILES=`git diff master --name-only | grep -E "\.(cpp|h)$"`
+    FILES=`grep -E "\.(cpp|h)$"`
 
     for FILE in $FILES; do
 
--- a/symengine/polynomial.cpp
+++ b/symengine/polynomial.cpp
@@ -1,8 +1,8 @@
-#include <symengine/polynomial.h>
 #include <symengine/add.h>
-#include <symengine/mul.h>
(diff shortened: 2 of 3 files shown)


CodeMaxx:
@certik Shouldn't we have a similar test like [this](https://github.com/symengine/symengine/blob/master/.travis.yml#L33) for OS X ?

symengine/polynomial.cpp
```diff
@@ -27,8 +27,7 @@ UnivariateIntPolynomial::UnivariateIntPolynomial(
     SYMENGINE_ASSERT(is_canonical(degree_, dict_))
 }
 
-bool UnivariateIntPolynomial::is_canonical(const unsigned int &degree_,
-                                           const map_uint_mpz &dict) const
+bool UnivariateIntPolynomial::is_canonical(const unsigned int &degree_, const map_uint_mpz &dict) const
```
certik: What caused this change?

bin/travis_clang.sh
```diff
@@ -11,7 +11,7 @@ if [[ "${TRAVIS_OS_NAME}" == "linux" ]] && \
         exit 1
     fi
 
-    FILES=`git diff master --name-only | grep -E "\.(cpp|h)$"`
+    FILES=`grep -E "\.(cpp|h)$"`
```
abhinavagarwal07: This is better unchanged

srajangarg:
@CodeMaxx I think this PR is not needed, the changes to clang-format have already been made.

CodeMaxx:
@srajangarg Cool :+1:

PR #906, opened 2016-04-02
clang-format of benchmarks/ directory

srajangarg:
Also, changes according to https://github.com/symengine/symengine/issues/903


certik:
That looks good, thanks!

PR #914, opened 2016-04-12
Travis: print which files are processed

certik:


Sumith1896:
Cool. +1

CodeMaxx:
+1

Sumith1896:
@certik I don't see the result here https://travis-ci.org/symengine/symengine/jobs/122611136 
What's wrong?

certik:
@Sumith1896 that's why I added it there, because I suspect it is not being run... No idea what's wrong.

certik:
And there are failures, as I suspected...

bin/travis_clang.sh
```diff
@@ -13,7 +18,11 @@ if [[ "${TRAVIS_OS_NAME}" == "linux" ]] && \
 
     FILES=`git ls-files | grep -E "\.(cpp|h|hpp)$" | grep -Ev "symengine/utilities" | grep -Ev "cmake/"`
 
+    echo "FILES="
+    echo $FILES
```
Sumith1896: Are we interested in printing this out?

Sumith1896:
Awesome! It now works as expected. 
We can merge it once tests pass.

bin/travis_clang.sh
certik: No, we don't need it, I removed it in 07828975d5030d71f318db182459ceaf9cdffb52.

certik:
The fix in 3ce37c55f25c39b47c988d9cd4ef73cbbafa36e4 should hopefully prevent such failures in the future. I don't know why the extra checks where there in the first place.

certik:
Tests pass, so I am going to merge this.

Issue #905, opened 2016-04-02
Clang_format changing the order of #include


CodeMaxx:
`clang-format` arranges the `#include`s in lexicographic order
This causes certain tests to fail in PR #893.
See this question on stackoverflow
http://stackoverflow.com/questions/865785/why-does-the-order-of-my-includes-matter-c
 @certik @isuruf

CodeMaxx:
Ok wait... I have clang-format 3.8 on my mac. This change has been introduced in this version.

certik:
Yes, right now we use 3.7.

CodeMaxx:
@certik What I want to say that, we will not be able to use version 3.8(latest) in future.

certik:
@CodeMaxx you can use the `IncludeCategories` option (http://clang.llvm.org/docs/ClangFormatStyleOptions.html) to order header files in the right order.

CodeMaxx:
@certik Take a look at #917

rikardn:
This has now been fixed in `.clang-format`

PR #917, opened 2016-04-15
Prevent clang-format from sorting includes

CodeMaxx:
Required for clang-format 3.8 or higher/
Changing the order of includes causes errors in some cases. 
Addresses #905 

@isuruf @certik


CodeMaxx:
This will work after we shift to version clang-format 3.8 or higher. Version 3.7 doesn't recognise it.

Sumith1896:
What exactly is the problem here and why is this needed?

rikardn:
This options is now in the new `.clang-format`

PR #930, opened 2016-05-05
[WIP] Initial implementation of the ks2 algorithm

srajangarg:
Paper [here](http://arxiv.org/pdf/0712.4046v1.pdf)

I've replaced the tests for `mul_poly` with `mul_poly_ks2` and they work. Anyway's I'll create another testbench for the ks2 algorithm. 

I would also like to implement the naive algorithm for multiplication, maybe useful for checking our other functions, and also for benchmarking reasons.

Also, I'll probably follow up this PR with a complete benchmarking suite for polynomials, where we can vary polynomial length and coefficients. Also, compare all the different function (including creation, addition/subtraction, multiplication (and others as they are added on later)). This will also help me decide on the unordered vs ordered map implementation of `UnivariatePolynomials`

### DIFF
--- a/symengine/polynomial.cpp
+++ b/symengine/polynomial.cpp
@@ -172,14 +172,16 @@ integer_class UnivariateIntPolynomial::eval(const integer_class &x) const
     return result;
 }
 
-integer_class UnivariateIntPolynomial::eval_bit(const int &x) const
+integer_class UnivariateIntPolynomial::eval_bit(const int &x, bool neg) const
 {
     unsigned int last_deg = dict_.rbegin()->first;
     integer_class result(0);
+    int mul;
 
     for (auto it = dict_.rbegin(); it != dict_.rend(); ++it) {
 
-        result = (*it).second + (result << x * (last_deg - (*it).first));
+        mul = ((*it).first % 2 == 0 or not neg) ? 1 : -1;
+        result = (*it).second * mul + (result << x * (last_deg - (*it).first));
         last_deg = (*it).first;
     }
     result = result << x * last_deg;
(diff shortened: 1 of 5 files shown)


srajangarg:
@Sumith1896 can you explain why the build fails?

symengine/polynomial.cpp
```diff
@@ -313,51 +315,126 @@ RCP<const UnivariateIntPolynomial> mul_poly(const UnivariateIntPolynomial &a,
...
           + bit_length(a.max_abs_coef()) + bit_length(b.max_abs_coef());
 
-    integer_class a1(1), b1, res;
-    a1 <<= N;
-    integer_class a2 = a1 / 2;
-    integer_class mask = a1 - 1;
-    integer_class a_val(a.eval_bit(N)), b_val(b.eval_bit(N));
-    integer_class s_val(a_val * b_val);
-    integer_class r = mp_abs(s_val);
+    integer_class full = 1, temp, res;
```
Sumith1896: @srajangarg Should be `integer_class full = integer_class(1)` IIRC

srajangarg:
I've added the benchmarking code I've used. It basically generates random polynomials and tests both the `mul_poly`s on it. Also, the benchmarks for the function run on my machine, as well as the python script for plotting the data can be found in this [gist].(https://gist.github.com/srajangarg/1164fa9c23443be55065bbbc0062cc90)

I've tested it out for 4 different degrees of sparseness, and 4 different coefficient lengths. Also, the polynomial length varies right from 1 to 10,000. The plots can be found [here](http://imgur.com/a/zqTxR).

Log scale :
[[http://i.imgur.com/uSjTrKg.png]]

Linear scale :
[[http://i.imgur.com/JhRGgyk.png]]

srajangarg:
I'll add my observations on these graphs here, soon. @Sumith1896 @isuruf Please tell me if anything else is required.

PR #982, opened 2016-06-09
Function symbol CWrappers

rajithv:
Wrappers for the FunctionSymbol class

### DIFF
--- a/symengine/cwrapper.cpp
+++ b/symengine/cwrapper.cpp
@@ -660,9 +660,9 @@ void basic_subs2(basic s, const basic e, const basic a, const basic b)
     s->m = e->m->subs({{a->m, b->m}});
 }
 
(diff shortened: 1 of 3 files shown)


symengine/cwrapper.cpp
```diff
@@ -26,6 +26,8 @@ using SymEngine::Basic;
 using SymEngine::RCP;
 using SymEngine::zero;
 using SymEngine::Symbol;
+using SymEngine::FunctionSymbol;
```
abinashmeher999: Is this line required?

srajangarg:
Hey @rajithv, can you follow the workflow as mentioned [here](https://github.com/symengine/symengine/wiki/Development-workflow-for-SymEngine) especially about the `clang-format`. Otherwise style changes creep into master, and show up in diffs of other PRs.

PR #1005, opened 2016-06-29

rwst:


symengine/visitor.h
```diff
@@ -72,7 +72,7 @@ void postorder_traversal_stop(const Basic &b, StopVisitor &v);
 class HasSymbolVisitor : public BaseVisitor<HasSymbolVisitor, StopVisitor>
 {
 protected:
-    RCP<const Symbol> x_;
+    const Symbol* x_;
```
certik: This should be using `Ptr<const Symbol>`, not raw pointers.

symengine/visitor.h
```diff
@@ -86,17 +86,17 @@ class HasSymbolVisitor : public BaseVisitor<HasSymbolVisitor, StopVisitor>
 
     void bvisit(const Basic &x){};
 
-    bool apply(const Basic &b, const RCP<const Symbol> &x)
+    bool apply(const Basic &b, const Symbol& x)
     {
-        x_ = x;
+        x_ = &x;
```
certik: This can become a dangling pointer if `x` ever gets deallocated. But since this visitor is only used from the `has_symbol` function, then this should be safe.

symengine/visitor.h
certik: I submitted #1013 with this fix.

rwst: `Ptr` cannot have `nullptr` so it cannot have state. However, you have two distinct states with visitors: unapplied and applied. In other words, the `Ptr<const Symbol>` is very much like a reference that must be immediately initialized.

rwst:
Closed in favor of #1014

PR #1006, opened 2016-06-29
coeff(): replace unnecessary RCP function args

rwst:
Depends on #1005 because of merge conflicts.


symengine/visitor.h
```diff
@@ -101,8 +101,8 @@ bool has_symbol(const Basic &b, const Symbol& x);
 class CoeffVisitor : public BaseVisitor<CoeffVisitor, StopVisitor>
 {
 protected:
-    RCP<const Symbol> x_;
-    RCP<const Basic> n_;
+    const Symbol* x_;
+    const Basic* n_;
```
certik: Use `Ptr<const Symbol>` etc., instead of the raw pointers.

certik:
Otherwise +1.

rwst:
`Ptr` cannot have `nullptr` so it cannot have state. However, you have two distinct states with visitors: unapplied and applied. In other words, the `Ptr<const Symbol>` is very much like a reference that must be immediately initialized.

rwst:
Closed in favor of #1018

PR #1010, opened 2016-06-29
pow_upoly changes

srajangarg:
move `pow_upoly` to upolybase (so that `UExprPoly` can also use it) and make expand use `pow_upoly` instead of it's own implementation.


certik:
That looks good to me.

PR #1013, opened 2016-06-30

certik:


rwst:
Compile fail because apparently Ptr must be initialized.

certik:
I see -- I only tested in Debug mode, which uses Teuchos, whose Ptr is a bit more general (and perhaps slower). SymEngine's Ptr is more restrictive. This should fix it.

rwst:
I think #1014 has a cleaner solution.

certik:
I agree. Let's keep this open just to make sure tests pass. But then close it in favor of #1014, which is cleaner, because it doesn't need to initialize the Ptr to null, and then assign something else to it.

Sumith1896:
https://github.com/symengine/symengine/pull/1016 got merged

PR #1014, opened 2016-07-02

rwst:
Take 3. There is still a coverage fail which led me to the investigation of a hidden bug in series_visitor.h (probably). Later.

### DIFF
--- a/symengine/visitor.h
+++ b/symengine/visitor.h
@@ -76,7 +76,10 @@ protected:
     bool has_;
 
 public:
-    HasSymbolVisitor(const Symbol &x) : x_(&x) {}
+    HasSymbolVisitor(const Symbol &x) : x_(&x)
+    {
+    }
+
     void bvisit(const Symbol &x)
     {
         if (x_->__eq__(x)) {
@@ -100,7 +103,7 @@ public:
     }
 };
 
-bool has_symbol(const Basic &b, const Symbol& x);
+bool has_symbol(const Basic &b, const Symbol &x);
 
 class CoeffVisitor : public BaseVisitor<CoeffVisitor, StopVisitor>
 {
--- a/symengine/visitor.h
+++ b/symengine/visitor.h
@@ -76,6 +76,7 @@ protected:
     bool has_;
 
 public:
+    HasSymbolVisitor(const Symbol &x) : x_(&x) {}
     void bvisit(const Symbol &x)
     {
         if (x_->__eq__(x)) {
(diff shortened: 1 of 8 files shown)


rwst:
The code suddenly complained about by coverage is unreachable because `has_symbols` is called first, and any unhandled type in `SeriesVisitor` is also unhandled in `HasSymbolVisitor` and so has no symbols, which leads to the simple constructor called:
```
28          if (not has_symbol(*ex, *var))
29              return make_rcp<const UPSeriesPiranha>(p_expr{Expression(ex)},
30                                                     var->get_name(), prec);
```
Thus there is no way to reach `SeriesVisitor::bvisit(const Basic&)` using the `series()` call, regardless of type. BTW a `FunctionWrapper` would trigger `bvisit(const Function&)`.

I will now try to reach the code in other ways.

symengine/visitor.h
```diff
@@ -86,17 +90,20 @@ class HasSymbolVisitor : public BaseVisitor<HasSymbolVisitor, StopVisitor>
...
     void bvisit(const Basic &x){};
 
-    bool apply(const Basic &b, const RCP<const Symbol> &x)
+    bool apply(const Basic &b)
     {
-        x_ = x;
+        // We are breaking a rule and assigning to a Ptr from a raw pointer
+        // directly. But since HasSymbolVisitor is only instantiated inside the
+        // has_symbol() function, the `x` can never go out of scope, so this is
+        // safe.
```
certik: You can remove this comment, as it doesn't apply anymore.

symengine/visitor.h
```diff
@@ -72,10 +72,14 @@ void postorder_traversal_stop(const Basic &b, StopVisitor &v);
 class HasSymbolVisitor : public BaseVisitor<HasSymbolVisitor, StopVisitor>
 {
 protected:
-    RCP<const Symbol> x_;
+    Ptr<const Symbol> x_;
     bool has_;
 
 public:
+    HasSymbolVisitor(const Symbol &x) : x_(&x)
```
certik: If you want to make this even cleaner, you can pass in `Ptr<const Symbol> x`, then you just assign Ptr instances, which is safe and clean. And you move up the responsibility of making sure the Symbol doesn't get deallocated one level up --- into `has_symbol`, which is the place which knows that Symbol won't go out of scope.

certik:
This looks good. I left a minor comment, but that can be fixed even after we merge this. I don't know what's going on with the coverage, but as far as I am concerned, feel free to merge it as is.

rwst:
I think codecov stats for patches are wrong in that they count all uncovered lines appearing in the diff (eg also the first and last), not only those changed. So I had to cover quite unrelated code lines.

certik:
@rwst I pushed in a few fixes here: #/1016, now I think it looks clean.

certik:
@rwst  see also #1017.

Issue #1017, opened 2016-07-03
Usage of ptrFromRef


certik:
This came up in:

https://github.com/symengine/symengine/pull/1016/files#diff-2c18f2e846e56fc460fb04a8379d21edL52

the old code looks like (CODE 1):
```c++
bool has_symbol(const Basic &b, const RCP<const Symbol> &x)
{
    HasSymbolVisitor v(x);
    return v.apply(b);
}
```
and the new code looks like (CODE 2):
```c++
bool has_symbol(const Basic &b, const Symbol &x)
{
    HasSymbolVisitor v(ptrFromRef(x));
    return v.apply(b);
}
```
In the old code, `HasSymbolVisitor` accepts `const RCP<const Symbol> &x` and so there is no problem. Except that one increments and then decrements the refcount, so it's a bit slower.

In the new code, `HasSymbolVisitor` accepts `Ptr<const Symbol> x`, so there is no unnecessary refcount increment/decrement. However, the `ptrFromRef` must be called, which is unsafe, because if HasSymbolVisitor saved the pointer somewhere, then it can become dangling once this function finishes. In this particular case, `HasSymbolVisitor` gets deallocated before the function finishes, *and* it doesn't save any pointers in other places (but you don't know that from this piece of code, that's the problem), and so it is safe.

The problem is that just by looking at a local piece of code, it must be possible to statically determine, whether the code is safe. Eventually we'll have plugins to Clang and other compilers that will check that automatically. In the meantime, we have to check by hand. Either way, the rule is that using `ptrFromRef` is unsafe, for the above reason.

A safe solution then is (CODE 3):
```c++
bool has_symbol(const Basic &b, Ptr<const Symbol> x)
{
    HasSymbolVisitor v(x);
    return v.apply(b);
}
```

When it is called, let's initialize some example first:
```c++
    RCP<const Basic> r1;
    RCP<const Symbol> x=symbol("x"), y=symbol("y");
    r1 = add(x, pow(y, integer(2)));
```
then CODE 1 looks like:
```c++
    has_symbol(*r1, x);
```
CODE 2 looks like:
```c++
    has_symbol(*r1, *x);
```
and CODE 3 looks like:
```c++
    has_symbol(*r1, x.ptr());
```
From the caller's perspective, all 3 cases are safe. Just the implementation of `has_symbol` in CODE 2 is not safe.

PR #1018, opened 2016-07-04
coeff(): replace unnecessary RCP function args

rwst:
cleaned up and updated version of #1006.


certik:
It's consistent with what we already merged, and I think the code as it is is safe, so I am merging it. However, my objections is written up in #1017, we should never be using `ptrFromRef`, so I think we need to change
```c++
RCP<const Basic> coeff(const Basic &b, const Basic &x, const Basic &n) 
```
to
```
RCP<const Basic> coeff(const Basic &b, const Ptr<const Basic> x, const Ptr<const Basic> n) 
```
and then just assign pointers inside the `coeff` method. The caller then takes an `RCP` instance and calls `.ptr()` on it to obtain a `Ptr` instance. The `Ptr` if used by value, as indicated by this comment, it as fast in Release mode as a raw pointer, I checked the generated assembly and it is equivalent. So the performance is not affected by using Ptr instead of a raw pointer or a raw reference. But it is 100% safe, if you never use `ptrFromRef`, but only `.ptr()` to obtain it.

Issue #1012, opened 2016-06-30
test_series_generic is too slow


certik:
In Debug mode it takes 8s on my computer, while every other test takes < 0.1s. This should be fixed to run in 0.1s or less, if at all possible.
```
Test project /home/certik/repos/symengine
      Start  1: test_rcp
 1/28 Test  #1: test_rcp .........................   Passed    0.00 sec
      Start  2: test_basic
 2/28 Test  #2: test_basic .......................   Passed    0.03 sec
      Start  3: test_arit
 3/28 Test  #3: test_arit ........................   Passed    0.09 sec
      Start  4: test_poly
 4/28 Test  #4: test_poly ........................   Passed    0.01 sec
      Start  5: test_series
 5/28 Test  #5: test_series ......................   Passed    0.01 sec
      Start  6: test_series_generic
 6/28 Test  #6: test_series_generic ..............   Passed    8.07 sec
      Start  7: test_functions
 7/28 Test  #7: test_functions ...................   Passed    0.11 sec
      Start  8: test_subs
 8/28 Test  #8: test_subs ........................   Passed    0.01 sec
      Start  9: test_integer
 9/28 Test  #9: test_integer .....................   Passed    0.01 sec
      Start 10: test_rational
10/28 Test #10: test_rational ....................   Passed    0.01 sec
      Start 11: test_number
11/28 Test #11: test_number ......................   Passed    0.01 sec
      Start 12: test_as_numer_denom
12/28 Test #12: test_as_numer_denom ..............   Passed    0.01 sec
      Start 13: test_parser
13/28 Test #13: test_parser ......................   Passed    0.02 sec
      Start 14: test_sets
14/28 Test #14: test_sets ........................   Passed    0.01 sec
      Start 15: test_fields
15/28 Test #15: test_fields ......................   Passed    0.03 sec
      Start 16: test_ntheory
16/28 Test #16: test_ntheory .....................   Passed    0.08 sec
      Start 17: test_diophantine
17/28 Test #17: test_diophantine .................   Passed    0.03 sec
      Start 18: test_matrix
18/28 Test #18: test_matrix ......................   Passed    0.13 sec
      Start 19: test_printing
19/28 Test #19: test_printing ....................   Passed    0.01 sec
      Start 20: test_uintpoly
20/28 Test #20: test_uintpoly ....................   Passed    0.02 sec
      Start 21: test_mintpoly
21/28 Test #21: test_mintpoly ....................   Passed    0.03 sec
[...]

PR #1019, opened 2016-07-04
reduce precision in many series_generic tests; part of #1012

rwst:
gives a 5.5x speedup in `test_series_generic`.


Sumith1896:
+1, I hope the coverage is unaffected...

rwst:
Since you say +1, I'm feeling free to merge. Thanks for reviewing.

PR #1022, opened 2016-07-04
fix expanding of 2**(x+1)

srajangarg:
This is a quick solution I came up with, I think there might be better ways to do what I have wanted to do. (eg. Get the contents of an `Add` without the `coef_` etc.)


certik:
Doesn't this suffer from the problem I described in https://github.com/symengine/symengine/issues/1021#issuecomment-230898784?

PR #1041, opened 2016-07-20
add cotire to speed up builds

srajangarg:
with a cmake option `WITH_COTIRE`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,8 @@ if ((CMAKE_CXX_COMPILER_ID MATCHES Clang) AND (CMAKE_BUILD_TYPE STREQUAL "Releas
     endif()
 endif()
 
+# Cotire
+set(WITH_COTIRE yes CACHE BOOL "Use cotire to speed up build")
 
 set(INTEGER_CLASS "gmp"
     CACHE STRING "Integer class for symengine. Either gmp, gmpxx, flint or piranha")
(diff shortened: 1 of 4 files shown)


srajangarg:
@isuruf How is `cmake` installed on the travis machines? I tried, but couldn't find the specific line in `install_travis.h`. `cotire` complains of using an older version which is not supported.

certik:
I think `cmake` is available on Travis by default, isn't it?

srajangarg:
Is that so? Then it wouldn't be possible to use a newer CMake version, right?

certik:
No, we can use a newer cmake. We just have to install it, probably using Conda.

isuruf:
What's the latest cmake needed for cotire?

srajangarg:
2.8.12

isuruf:
Okay, then we either have to 
1 - bump the version of cmake needed in `CMakeLists.txt` to 2.8.12 or 
2 - disable cotire by default or
3 - enable cotire by default only if cmake version >= 2.8.12
I'm for (3)

isuruf:
Also, can you enable the tests on appveyor?

srajangarg:
Yes (3) makes sense. Will you be adding necessary commits to the PR?

isuruf:
I can do that, but won't be able to for a while though

srajangarg:
@isuruf please review. Added in your suggestion. You can see that linux builds are done without cotire while mac builds use cotire (have latest cmake using brew)

certik:
I am fine with 1 actually. Once I finish #757, we will have to require CMake 3.0 (or perhaps even a later version) anyway.

srajangarg:
@certik How about we go with this for now, and when you finish #757, all you have to do is uncomment the cmake version check.

CMakeLists.txt
```diff
@@ -92,7 +92,11 @@ if ((CMAKE_CXX_COMPILER_ID MATCHES Clang) AND (CMAKE_BUILD_TYPE STREQUAL "Releas
 endif()
 
 # Cotire
-set(WITH_COTIRE yes CACHE BOOL "Use cotire to speed up build")
+if(CMAKE_VERSION VERSION_GREATER 2.8.12)
```
isuruf: Shouldn't this be `> 2.8.11` (i.e. `>=2.8.12`)?

isuruf: space after `if`

certik:
That's fine with me.

srajangarg:
@certik merging this in, I'll add a todo point in #757 as a reminder.

PR #1096, opened 2016-10-18
complimentary error function

RiteshKB:


certik:
Hi @RiteshKB thanks for the patch. Can you add some tests for the erfc function as well please? Otherwise I think it looks good.

symengine/functions.h
```diff
@@ -924,6 +924,30 @@ class Erf : public OneArgFunction
 //! Canonicalize Erf:
 RCP<const Basic> erf(const RCP<const Basic> &arg);
 
+class Erfc : public OneArgFunction
+{
+    /*   The Gauss error function. This function is defined as:
+     *
+     *   .. math::
+     *      \mathrm{erfc}(x) =1 neg( \frac{2}{\sqrt{\pi}} \int_0^x e^{-t^2})
```
certik: Should `neg` be repleaced with `-`?

Also, the parentheses should probably be removed. Like this:

```
\mathrm{erfc}(x) = 1 - \frac{2}{\sqrt{\pi}} \int_0^x e^{-t^2}
```

symengine/functions.h
```diff
@@ -924,6 +924,30 @@ class Erf : public OneArgFunction
 //! Canonicalize Erf:
 RCP<const Basic> erf(const RCP<const Basic> &arg);
 
+class Erfc : public OneArgFunction
+{
+    /*   The Gauss error function. This function is defined as:
```
certik: Isn't this the complementary error function?

RiteshKB: yes @certik

RiteshKB:
Im working on tests of erfc function.

nano.save
```diff
@@ -0,0 +1 @@
+
```
certik: You should remove this file.

symengine/functions.h
```diff
@@ -924,6 +924,30 @@ class Erf : public OneArgFunction
 //! Canonicalize Erf:
 RCP<const Basic> erf(const RCP<const Basic> &arg);
 
+class Erfc : public OneArgFunction
+{
+    /*   The complementary error function. This function is defined as:
+     *
+     *   .. math::
+     *      \mathrm{erfc}(x) =1 - \frac{2}{\sqrt{\pi}} \int_0^x e^{-t^2}
+     *\mathrm{d}t.
```
certik: I would put a space after `=`, and also indent the second line with `\mathrm{d}t.` to the right, something like this:
```
*                \mathrm{d}t.
```

Issue #1112, opened 2016-11-01
Consolidate some of the Travis tests


certik:
We currently test many of our libraries as a separate test. But I think we should try to consolidate tests of libraries that do not clash with each other. For example, I think the ECM, Primesieve and MPC tests can all be done in one test.

More tests can probably be consolidated, while still testing everything.

mohd-omama:
Hi, I'm new to symengine

I would like to fix this bug!

Issue #503, opened 2015-07-03
Add Clang memory sanitizer and address sanitizer checks to Travis


certik:
We should run all our tests with the Clang's memory and address sanitizers and run the whole C++ testsuite (and the C suite as well).

http://clang.llvm.org/docs/MemorySanitizer.html
http://clang.llvm.org/docs/AddressSanitizer.html

These checks are good to always run (even if the way we use C++ should guarantee to have no leaks there, but there can be some in the C wrappers and they are also able to catch other problems with the code).

If the whole code passes with these checks on, then we should make the test fail if the checks fail.

bollu:
I'd like to take the issue up if no one is working on it

Sumith1896:
Go ahead @bollu

PR #1115, opened 2016-11-03
[WIP] Consolidation of Tests regarding Issue #1112

ShikharJ:
The ECM, MPC and PRIMESIEVE libraries seem to work without conflict on a single build. This patch reduces the total number of travis test builds from 20 to 18 ( regarding PR #1111 ) for the issue #1112.


ShikharJ:
@isuruf I believe this should work for now ( for Issue #503 and PR #1111 ). Any other combination of builds that are worth trying?

ShikharJ:
I believe that the changes implemented here will eventually be implemented in PR #1116. Hence it only makes sense to close this PR.

Issue #136, opened 2014-04-11
Converting public members to private members


sushant-hiray:
There are quite a few classes which currently have members defined as public.
We need to convert them to private members. Such `TODO's` scatter over the relevant files. 

For instance:
-   [add.h](https://github.com/sympy/csympy/blob/master/src/add.h#L-16-18) 
-   [integer.h](https://github.com/sympy/csympy/blob/master/src/integer.h#L-20-21)
-   [mul.h](https://github.com/sympy/csympy/blob/master/src/mul.h#L-15-17)
-   [pow.h](https://github.com/sympy/csympy/blob/master/src/pow.h#L-17-19)
-   [rational.h](https://github.com/sympy/csympy/blob/master/src/rational.h#L-18-20)

These are quite a few files, but are quite easy to fix.

pradyotprakash:
Hi! I am new to symengine development.
Can I get started on this?

certik:
Hi Pradyot, sure, go ahead!

pradyotprakash:
Hi Certik,
I tinkered with the code a bit and it seems that most of the public variables can't be made private without considerable amount of changes to the entire codebase. There are several instances where the public members are being used directly, and new functions would be needed to access them or a lot of friend declarations would be needed.

For eg, by changing the public in line 17 of add.h, error pops up at line 251 of add.cpp.

What should I do?

pradyotprakash:
I have converted the public variables of the above mentioned 5 files to private. Should I go ahead and look for similar changes in other files as well?

isuruf:
Can you update the PR first?

ShikharJ:
Fixed via #1150.

isuruf:
@ShikharJ, is this fixed for all Basic subclasses?

ShikharJ:
@isuruf The ones mentioned above and a few others. I could not make changes to all, as some of them such as `ComplexDouble` and `RealDouble` used a certain template routine, that I didn't want to touch.

PR #1150, opened 2016-12-11
Changed Public and Protected Data Members to Private In Select Classes

ShikharJ:
Relevant: #136 
@certik @isuruf 
Currently, the builds fail due to a single failure in test_series_generic (line 318 onward), which came up while applying the changes for the Rational class. Any suggestions?
UPDATE: I have pushed all the non-erroneous changes, and I'd like to address the Rational and other classes through another PR later. Please review and suggest improvements.
UPDATE 2: The issues have been resolved for most of the intended classes.

### DIFF
--- a/symengine/add.h
+++ b/symengine/add.h
@@ -14,7 +14,7 @@ namespace SymEngine
 
 class Add : public Basic
 {
-public:                      // TODO: make this private
+private:
     RCP<const Number> coef_; //! The coefficient (e.g. `2` in `2+x+y`)
     umap_basic_num dict_; //! The dictionary of the rest (e.g. `x+y` in `2+x+y`)
 
@@ -69,11 +69,11 @@ public:
 
     virtual vec_basic get_args() const;
 
-    RCP<const Number> get_coef() const
+    inline const RCP<const Number> &get_coef() const
     {
         return coef_;
     }
-    const umap_basic_num &get_dict() const
+    inline const umap_basic_num &get_dict() const
     {
         return dict_;
     }
--- a/symengine/polys/basic_conversions.cpp
+++ b/symengine/polys/basic_conversions.cpp
@@ -72,7 +72,7 @@ public:
 
     void bvisit(const Add &x)
     {
-        for (auto it : x.dict_)
+        for (auto it : x.get_dict())
             it.first->accept(*this);
     }
 
@@ -112,10 +112,10 @@ public:
 
     void bvisit(const Add &x)
     {
-        if (not x.coef_->is_zero())
-            x.coef_->accept(*this);
+        if (not x.get_coef()->is_zero())
+            x.get_coef()->accept(*this);
 
-        for (auto it : x.dict_) {
+        for (auto it : x.get_dict()) {
             RCP<const Number> mulx = one, divx = one;
 
             if (it.second->is_negative())
--- a/symengine/polys/basic_conversions.cpp
+++ b/symengine/polys/basic_conversions.cpp
@@ -78,7 +78,7 @@ public:
 
     void bvisit(const Mul &x)
     {
-        for (auto it : x.dict_)
+        for (auto it : x.get_dict())
             it.first->accept(*this);
     }
 
@@ -134,13 +134,13 @@ public:
         // needs `expand` to have been called
         RCP<const Number> mulx = one, divx = one;
 
-        if (x.coef_->is_negative())
+        if (x.get_coef()->is_negative())
             mulx = minus_one;
 
-        if (is_a<const Rational>(*x.coef_))
-            divx = down_cast<const Rational &>(*x.coef_).get_den();
+        if (is_a<const Rational>(*x.get_coef()))
+            divx = down_cast<const Rational &>(*x.get_coef()).get_den();
 
-        auto dict = x.dict_;
+        auto dict = x.get_dict();
         gen_set[Mul::from_dict(mulx, std::move(dict))] = divnum(one, divx);
     }
 
--- a/symengine/sets.h
+++ b/symengine/sets.h
@@ -138,7 +138,7 @@ public:
 
 class Interval : public Set
 {
-public:
+private:
     RCP<const Number> start_;
     RCP<const Number> end_;
     bool left_open_, right_open_;
@@ -165,6 +165,23 @@ public:
     virtual RCP<const Set> set_intersection(const RCP<const Set> &o) const;
     virtual RCP<const Boolean> contains(const RCP<const Basic> &a) const;
     virtual vec_basic get_args() const;
+
+    inline const RCP<const Number> &get_start() const
+    {
+        return start_;
+    }
+    inline const RCP<const Number> &get_end() const
+    {
+        return end_;
+    }
+    inline const bool &get_left_open() const
+    {
+        return this->left_open_;
+    }
+    inline const bool &get_right_open() const
+    {
+        return this->right_open_;
+    }
 };
 
 class Union : public Set
--- a/symengine/sets.h
+++ b/symengine/sets.h
@@ -110,7 +110,7 @@ public:
 
 class FiniteSet : public Set
 {
-public:
+private:
     set_basic container_;
 
 public:
@@ -129,6 +129,11 @@ public:
     virtual RCP<const Set> set_union(const RCP<const Set> &o) const;
     virtual RCP<const Set> set_intersection(const RCP<const Set> &o) const;
     virtual RCP<const Boolean> contains(const RCP<const Basic> &a) const;
+
+    inline const set_basic &get_container() const
+    {
+        return this->container_;
+    }
 };
 
 class Interval : public Set
@@ -227,8 +232,8 @@ inline RCP<const Set> set_union(const set_set &in, bool solve = true)
     for (auto it = in.begin(); it != in.end(); ++it) {
         if (is_a<FiniteSet>(**it)) {
             const FiniteSet &other = down_cast<const FiniteSet &>(**it);
-            combined_FiniteSet.insert(other.container_.begin(),
-                                      other.container_.end());
+            combined_FiniteSet.insert(other.get_container().begin(),
+                                      other.get_container().end());
         } else if (is_a<UniversalSet>(**it)) {
             return universalset();
         } else if (not is_a<EmptySet>(**it)) {
--- a/symengine/sets.h
+++ b/symengine/sets.h
@@ -169,7 +169,7 @@ public:
 
 class Union : public Set
 {
-public:
+private:
     set_set container_;
 
 public:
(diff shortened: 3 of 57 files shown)


ShikharJ:
@certik @isuruf 
Here are some benchmarks that I tried to run on master and this branch (in a `before -- after` order). Only single runs are reported here (no averages):
```
add1
3649ms -- 3644ms

expand1
2276ms -- 2292ms

expand2b
313ms -- 313ms

expand3
706ms -- 772ms

expand6
704ms -- 715ms

expand6b
877ms -- 851ms

expand7
4583ms -- 4471ms

lwbench
Time for A : 	     0.023280959 -- 0.010350169
Time for B : 	     0.041580161 -- 0.019532893
Time for C : 	     0.027206264 -- 0.016515596
Time for D : 	     0.002486950 -- 0.001866515
Time for E : 	     0.002445461 -- 0.001560530

matrix_add1
186 microseconds -- 185 microseconds

matrix_add2
120 microseconds -- 106 microseconds

matrix_mul1
1479 microseconds -- 1515 microseconds

matrix_mul2
630 microseconds -- 628 microseconds

ntheorybench
mertens(1):0ms -- 0ms
mertens(2):0ms -- 0ms
mertens(3):0ms -- 0ms
mertens(4):0ms -- 0ms
mertens(8):0ms -- 0ms
mertens(16):0ms -- 0ms
mertens(32):1ms -- 1ms
mertens(64):3ms -- 2ms
mertens(113):5ms -- 4ms
mobius(2): 0ms -- 0ms
mobius(3): 0ms -- 0ms
mobius(4): 0ms -- 0ms
mobius(8): 0ms -- 0ms
mobius(16): 0ms -- 0ms
prime_factor_multiplicities(primes_mul,2): 0ms -- 0ms 
prime_factor_multiplicities(primes_mul,3): 0ms -- 0ms
prime_factor_multiplicities(primes_mul,4): 0ms -- 0ms
prime_factor_multiplicities(primes_mul,8): 0ms -- 0ms
prime_factor_multiplicities(primes_mul,16): 0ms -- 0ms
mp_sqrt(2): 0ms -- 0ms
mp_sqrt(3): 0ms -- 0ms
mp_sqrt(4): 0ms -- 0ms

series
13506ms -- 13389ms

symbench
Time for R1 : 	     0.018723835 -- 0.021276283
Time for R2 : 	     0.299823418 -- 0.309304477
Time for R3 : 	     0.000025688 -- 0.000023863
Time for R5 : 	     0.000062277 -- 0.000063088
Time for R7 : 	     2.192311288 -- 2.219490902
Time for R8 : 	     0.259632106 -- 0.257633402
Time for S1 : 	     0.593725027 -- 0.597372557
Time for S2 : 	     0.648153065 -- 0.649468341
Time for S3 : 	     0.564077721 -- 0.571550247

symengine_bench
93ms -- 98ms
```

ShikharJ:
@isuruf Ping.

ShikharJ:
@certik @isuruf 
Final Benchmarks (averaged out of five, on `i5-2410M/ 6GB`):
```
benchmark      before - after
./add1         3636   - 3643
./expand1      2306   - 2290
./expand2b     306    - 312
./expand3      734    - 719
./expand6      721    - 708
./expand6b     857    - 838
./expand7      4413   - 4442
```

ShikharJ:
@isuruf @certik Ping.

certik:
Please do not merge yet. This is a big change, I want to run the benchmarks myself to make sure things didn't get slower.

certik:
I ran thorough benchmarks, and I didn't observe any slowdown against master.

symengine/add.h
```diff
@@ -69,11 +69,11 @@ class Add : public Basic
 
     virtual vec_basic get_args() const;
 
-    RCP<const Number> get_coef() const
+    inline const RCP<const Number> &get_coef() const
     {
         return coef_;
     }
```
certik: So the only issue I have with this PR are changes like these. A safe rule is to never return `&` from a function, because it's too easy to do something like:
```c++
// create `e`
const auto a = e.get_coef();
// destroy `e`

std::cout << *a; // UNDEFINED
```
I'll write soon about some workarounds for this.

certik: Ok, so the get_coef() is being passed around. In that case, one can't return `&`, because I think the RCP might not get activated properly.

The solution for things where the result is not passed around is to return a `Ptr`:
```diff
-    inline const RCP<const Number> &get_coef() const
+    inline const Ptr<const Number> get_coef() const
     {
-        return coef_;
+        return coef_.ptr();
     }
```
That is safe.

certik:
@ShikharJ can you apply the following patch:
```diff
--- a/symengine/add.h
+++ b/symengine/add.h
@@ -69,11 +69,11 @@ public:
 
     virtual vec_basic get_args() const;
 
-    inline const RCP<const Number> &get_coef() const
+    inline const RCP<const Number> get_coef() const
     {
         return coef_;
     }
-    inline const umap_basic_num &get_dict() const
+    inline const umap_basic_num get_dict() const
     {
         return dict_;
     }
```
and a similar patch to all the other places. It compiles, but I get a failure in `test_arit`.

symengine/add.h
isuruf: How is
```C++
// create `e`
const auto a = e.get_dict();
// destroy `e`

std::cout << *a; // UNDEFINED
```

different from

```C++
// create `e`
const auto a = e.dict_;
// destroy `e`

std::cout << *a; // UNDEFINED
```
?

isuruf:
Returning a const reference in case of non-RCP is cheaper. Returning `const umap_basic_num&` is cheaper when all you want is to iterate the dictionary and you don't want to make a copy which is expensive.
For RCP, I'm fine with either way as there is very little difference in performance and shouldn't affect any benchmarks.

ShikharJ:
@certik I had referred to your comment here: https://github.com/symengine/symengine/pull/717#r48550788
Hence the return type as `const RCP<const Number> &`.

symengine/polys/basic_conversions.cpp
```diff
@@ -134,13 +134,13 @@ class PolyGeneratorVisitorPow : public BaseVisitor<PolyGeneratorVisitorPow>
         // needs `expand` to have been called
         RCP<const Number> mulx = one, divx = one;
 
-        if (x.coef_->is_negative())
+        if (x.get_coef()->is_negative())
             mulx = minus_one;
```
certik: Here is an example that is correct with returning a `RCP <> &`, because it is used on the same line. So this usage is fine.

symengine/subs.h
```diff
@@ -32,15 +32,15 @@ class SubsVisitor : public BaseVisitor<SubsVisitor>
...
         RCP<const Number> coef;
 
-        auto it = subs_dict_.find(x.coef_);
+        auto it = subs_dict_.find(x.get_coef());
         if (it != subs_dict_.end()) {
             coef = zero;
             Add::coef_dict_add_term(outArg(coef), d, one, it->second);
         } else {
-            coef = x.coef_;
+            coef = x.get_coef();
```
certik: Here is an example, where the result of `get_coef()` is saved. What this means is that the refcount is *not* incremented in this case. Which is a bug, because if `x` goes out of scope before `coef`, this will segfault. That is my understanding --- do you agree with me, or am I wrong? If I am right, then this is inherently unsafe, and thus this usage should not be used.

isuruf: Yes, a const ref should never be stored.

isuruf:
As I said earlier, it is totally fine to return a RCP type that is not a reference. (We don't need `get_coef_rcp`, just `get_coef` that returns a `RCP<>` with no `&` is fine.)
I'm only concerned about non-RCP types.

certik:
@ShikharJ I apologize for that comment --- I didn't realize that it would be also used to assign it. Then it's unsafe. The problem is that I am not sure if `x = e.get_coef()` is equivalent to `x = e.coef_`, if `get_coef()` returns `RCP <>` (no reference). I am hoping that the compiler can elide the result, but we should check that (by looking at the assembly).

This is pretty tricky, and that's why I decided to just use the `coef_` directly, to avoid these issues. I agree that we should figure out a solution, that doesn't require accessing the internal variable, but that is still safe, but as fast.

certik:
@isuruf right, we can just do the following change:
```diff
-    inline const RCP<const Number> &get_coef() const
+    inline const RCP<const Number> get_coef() const
```

isuruf:
Following is better.
```diff
-    inline const RCP<const Number> &get_coef() const
+    inline RCP<const Number> get_coef() const
```

isuruf:
> This is pretty tricky, and that's why I decided to just use the coef_ directly, to avoid these issues. 

I'd like to mention again that accessing `.coef_` is exactly the same as returning a const reference.

certik:
Wait. So with this current PR (returning const reference), is this usage safe or not safe:
```
coef = x.get_coef();
```
Because previously we did `coef = x.coef_` and that is safe.

isuruf:
```
coef = x.get_coef();
```
and
```
coef = x.coef_;
```
are exactly the same for the compiler. If one is safe, so is the other.

certik:
Ok. Here you commented: https://github.com/symengine/symengine/pull/1150#discussion_r95070311, that you agree with me that it's not safe. If this PR is safe, as it is, then I have no objections. I thought it is unsafe.

certik:
What happens if you do:
```
auto x = e.get_coef()
```
Will `x` be a reference, or not?

symengine/subs.h
isuruf: I meant that both `coef = x.coef_;` and `coef = x.get_coef();` can be unsafe.

isuruf:
`auto x = e.get_coef()`, here `x` will be a const reference.
`auto x = e.coef_`, here `x` will be a const reference.

ShikharJ:
@isuruf In addition to changing `inline const RCP<const Number> &get_coef()` to `inline RCP<const Number> get_coef()`, what should be done of the Non-RCP data members? Should I add an extra function which returns a non-reference data member?

certik:
Actually, I think that if you do:
```
auto x = e.coef_
```

Then `x` is of the type `RCP<Basic>`, not a const reference.

certik:
@ShikharJ wait with further changes. Perhaps this PR is ok after all. I think there are fundamental misconceptions that we need to figure out.

As I said, my only requirement is that things are 100% safe in Debug mode, but no performance hit in Release mode.

@ShikharJ if you want to help, you can help us resolve this issue.

certik:
Do not merge.

Based on our discussion on gitter, it seems it should return a reference after all. Also, since the PR got changed, I have to rerun benchmarks again.

symengine/add.h
certik: @isuruf, per our gitter discussion, indeed they are not different. We should return a reference.

symengine/sets.h
```diff
@@ -160,11 +165,28 @@ class Interval : public Set
...
+
+    inline const RCP<const Number> &get_start() const
+    {
+        return start_;
+    }
+    inline const RCP<const Number> &get_end() const
+    {
+        return end_;
+    }
+    inline const bool &get_left_open() const
```
isuruf: Returning a bool by reference is unnecessary

isuruf:
Thanks @ShikharJ

symengine/sets.h
certik: Right, simple types like `int`, `double`, `bool` (also I think `std::complex`) should be returned by value.

certik:
Thanks @ShikharJ for the PR. Much appreciated.

PR #1156, opened 2016-12-16
Replaced `rcp_static_cast<>()->` with `down_cast<>().`

ShikharJ:
Relevant: #1155


ShikharJ:
ping @certik

certik:
Sorry, I need to think about this a bit. The previous way should be safe, but this might not be in all cases. @isuruf if you like it, we can merge it.

ShikharJ:
@isuruf Could you please take a look at this also?

certik:
It's fine with me if @isuruf gives this +1.

Issue #1161, opened 2016-12-27
Missing else-if statement in rsub function in complex.h


ranjithkumar007:
In Complex.h 

```
virtual RCP<const Number> rsub(const Number &other) const
    {
        if (is_a<Rational>(other)) {
            return rsubcomp(down_cast<const Rational &>(other));
        } else if (is_a<Integer>(other)) {
            return rsubcomp(down_cast<const Integer &>(other));
        } else {
            throw NotImplementedError("Not Implemented");
        }
    };
```

The above function is missing an else-if statement for Number "other" of type "Complex".
rsubcomp(const Complex &other) function is already implemented but it is not called in rsub()

I think it should be like
```
virtual RCP<const Number> rsub(const Number &other) const
    {
        if (is_a<Rational>(other)) {
            return rsubcomp(down_cast<const Rational &>(other));
        } else if (is_a<Integer>(other)) {
            return rsubcomp(down_cast<const Integer &>(other));
        } else if (is_a<Complex>(other)) {
            return rsubcomp(down_cast<const Complex &>(other));
        } else {
            throw NotImplementedError("Not Implemented");
        }
    };
```

ranjithkumar007:
else-if is present in complex_double.h ...mis-interpreted the function with function in complex.h

PR #1162, opened 2016-12-27
added missing else-if statement in rsub

ranjithkumar007:
Relevant: #1161 
Can you review this PR.. @certik @isuruf


certik:
Can you also add a test, that failed before, but now it passes?

ranjithkumar007:
Sorry for mis-interpreting .
I thought the function as if is in complex_double.h and not containing rsubcomp for Complex type.

PR #1211, opened 2017-02-11
flexc++ / bisonc++ based parser

srajangarg:
- [x] Add in support for doubles and MPFR numbers
- [x] integrate properly into cmake
- [ ] integrate into travis
- [ ] Remove binary files from history

I have gone with flexc++ and bisonc++. Are they as commonly used as bison and flex? Also I'm not sure as to how the commands for generating the parser/scanner files have to be added into cmake. For now I have hardcoded them using `add_custom_command` in `symengine/CMakeLists.txt`


srajangarg:
@certik @isuruf Please let me know what you think.

srajangarg:
Maybe we can include the generated files `parser.cpp`, `parserbase.h`, `scanner.cpp` and `scannerbase.h`, into the repository. They will be regenerated each time, only if the user has `flexc++` and `bisonc++` available. This will ensure that these two do not become a hard dependency for SymEngine.

srajangarg:
This new flex/bions based parser is performing roughly **4.8 times slower** than our previous handwritten parser. Thoughts?

isuruf:
> For a 1.5x10^6 length string, the bison generated parser did 1.7 times as fast! This suggests that the auto-generated parser has more overhead of processing, while the actual parsing of the string is done faster.

This is probably because the old parser has lots of unnecessary copying.

Sumith1896:
@isuruf What are your views on this? Do we stick to the old parser or move on to (a more scalable) new parser?

isuruf:
One of my concerns is the speed for smaller strings

certik:
Yes, we should only merge it if it is faster or as fast as our current parser. Alternatively, we can have two parsers if this generated one has some other advantages over our current parser.

It the reason it is slower due to the fact that it has lots of recursion?

isuruf:
@srajangarg, I'm open to a generated parser even if it is slow due to https://github.com/symengine/symengine/issues/1444. I tried last weekend to fix the current parser, but it was hopeless.

certik:
Ok. In that case, let's do the following in this PR:

* add this parser in addition to our current parser.

* make this new parser (slower but more correct) the default, but keep the older parser (faster, but with bugs) as an option for a user to use.

isuruf:
I've merged with master and implemented all the features in the hand written parser. All tests pass now.
I've also removed the hack of writing the string to a file and using the file as input to the scanner.

Now for small strings the generated parser is not so bad. Less than 5% slowdown for a small string like, `x + y - sin(x)/(z**2-4) - x**(y**z)`

isuruf:
Tests pass. Needs a review

isuruf:
> The only issue I have is that the old parser is not tested at all, but the new parser passes all the old tests plus a few new ones. Is that correct?

Yes, that is correct

isuruf:
@certik, can you review this once more?

isuruf:
>  Should we test the old parser?

Apart from duplicating the tests for the old parser, I don't see how.

> How do you generate the parser files --- do we have the sources for that also? 

Yes, they are included. If you do `-DWITH_GENERATE_PARSER=yes`, then the generated files will be updated if they are out of date.

certik:
@srajangarg mentions there are some binary files in the history of this PR. They should be removed. Otherwise +1 to merge as it is.

isuruf:
Hmm, I don't see any. @srajangarg, can you confirm?

isuruf:
I checked https://patch-diff.githubusercontent.com/raw/symengine/symengine/pull/1211.patch and it doesn't contain any binary files

srajangarg:
Hi everyone, sorry I've been away for most part of the discussion. But yes,
I don't remember committing any binaries. So the merge should be good!

PR #1216, opened 2017-02-21
remove redundant check

ranjithkumar007:
numerator = zero case is already handled in the previous [if](https://github.com/symengine/symengine/blob/master/symengine/functions.cpp#L1094) statement.


ranjithkumar007:
ping @isuruf

srajangarg:
Thanks

PR #1223, opened 2017-03-15
to_string changed to ostrstream, mingw bug

srajangarg:


srajangarg:
@isuruf Should we merge?

PR #1255, opened 2017-05-10
poly file name changes + making symengine mpolys inherit from a base class

srajangarg:
this base class can be used to inherit the piranha mpoly class

I think that the file names are much more well defined now.


symengine/polys/mpoly_symengine.h
```diff
@@ -412,48 +406,50 @@ class MSymEnginePoly : public Basic
...
             return true;
         } else {
-            return (unified_eq(vars_, o_.vars_)
-                    && unified_eq(poly_.dict_, o_.poly_.dict_));
+            return (unified_eq(this->get_vars(), o_.get_vars())
+                    && unified_eq(this->get_poly().dict_, o_.get_poly().dict_));
         }
     }
+
+    static set_basic get_translated_container(Container &x, Container &y,
```
srajangarg: Was moved inside the class, as different classes can have different `get_translated_container` methods, which serve different purposes. Hence the `add/sub/mul/div` functions use that particular class's `get_translated` method, and remain the same for all classes. (just how `+, -, x, /` were overloaded for univariate poly classes and implemented commonly for all classes)

srajangarg:
@isuruf can we merge? I'm unable to write multivariate piranha polynomials at the moment. Will do it another PR.

certik:
That looks good to me. @isuruf any objections?

PR #1260, opened 2017-05-12
Remove static keyword from LLVM double ONE_ARG_EXTERNAL_FUNCTION macro.

jppelteret:
This fixes a segmentation fault that may arise when sequential calls are
made to the functions defined with this macro.


jppelteret:
All credit for this bugfix goes to @isuruf :-)

PR #1337, opened 2017-10-04
Implement Symbolic Factorial and Binomial

ShikharJ:
Relevant: #1336


symengine/functions.cpp
```diff
@@ -3269,6 +3269,49 @@ RCP<const Basic> trigamma(const RCP<const Basic> &x)
...
+    if (is_a<NaN>(*arg)) {
+        return Nan;
+    } else if (is_a<Infty>(*arg)) {
+        if (down_cast<const Infty &>(*arg).is_positive()) {
+            return Inf;
+        }
+    } else if (is_a<Integer>(*arg)
+               or (is_a_Number(*arg)
+                   and not down_cast<const Number &>(*arg).is_exact())) {
+        return gamma(add(arg, one));
```
isuruf: There's a `factorial` in `ntheory` that you can use here.

ShikharJ: My concern is regarding code duplication that would occur. `gamma` already has the cases defined. It would be a literal copy-paste.

isuruf: Okay.

isuruf: This returns an instance of `Gamma` sometimes. Can you return a `Factorial` instance in that case?

ShikharJ:
@isuruf I think this has progressed enough for a review.

ShikharJ:
@isuruf Can I please have a review of this?

symengine/functions.cpp
```diff
@@ -3270,6 +3270,161 @@ RCP<const Basic> trigamma(const RCP<const Basic> &x)
...
+        }
+    } else if (is_a<Integer>(*arg)
+               or (is_a_Number(*arg)
+                   and not down_cast<const Number &>(*arg).is_exact())) {
+        if (not(is_a<Rational>(*arg)
+                and neq(*down_cast<const Rational &>(*arg).get_den(),
+                        *integer(2)))) {
+            return gamma(add(arg, one));
+        }
+    }
```
isuruf: SymEngine's gamma and sympy's factorial evaluates this for floating points. How about just returning `gamma(arg + 1)`?

symengine/functions.cpp
```diff
@@ -3270,6 +3270,161 @@ RCP<const Basic> trigamma(const RCP<const Basic> &x)
...
+               and (is_a<Integer>(*n)
+                    and not down_cast<const Number &>(*n).is_negative())) {
+        return zero;
+    } else if (is_same_type(*n, *k)) {
+        if (is_a<Integer>(*n)) {
+            return binomial(down_cast<const Integer &>(*n),
+                            down_cast<const Integer &>(*k).as_uint());
+        }
+        if (is_a<Rational>(*n) or is_a<RealDouble>(*n)) {
+            if (down_cast<const Number &>(*sub(n, k)).is_negative()) {
```
isuruf: Move this to under `is_a_Number(*n) and is_a_Number(*k)`

symengine/functions.cpp
```diff
@@ -3270,6 +3270,161 @@ RCP<const Basic> trigamma(const RCP<const Basic> &x)
...
+{
+    return binomial(n, k);
+}
+
+RCP<const Basic> binomial(const RCP<const Basic> &n, const RCP<const Basic> &k)
+{
+    if (is_a_Number(*k) and down_cast<const Number &>(*k).is_zero()) {
+        return one;
+    } else if (is_a_Boolean(*n) or is_a_Boolean(*k) or is_a_Set(*n)
+               or is_a_Set(*k)) {
```
isuruf: Don't check this. Garbage values in -> garbage values out

symengine/functions.cpp
```diff
@@ -3270,6 +3270,161 @@ RCP<const Basic> trigamma(const RCP<const Basic> &x)
...
+        throw SymEngineException(
+            "Boolean and Set objects not allowed in this context.");
+    } else if (eq(*n, *k) and not(is_a<NaN>(*n) or is_a<Infty>(*n))) {
+        return one;
+    } else if (is_a<NaN>(*n)) {
+        if ((is_a<Integer>(*k)
+             and down_cast<const Number &>(*k).is_positive())) {
```
isuruf: Why? Just return `NaN` if any one of `n` or `k` is a `NaN`

symengine/functions.cpp
```diff
@@ -3270,6 +3270,161 @@ RCP<const Basic> trigamma(const RCP<const Basic> &x)
...
+            return n;
+        }
+    } else if (eq(*k, *Inf)
+               and (is_a<Integer>(*n)
+                    and not down_cast<const Number &>(*n).is_negative())) {
+        return zero;
+    } else if (is_same_type(*n, *k)) {
+        if (is_a<Integer>(*n)) {
+            return binomial(down_cast<const Integer &>(*n),
+                            down_cast<const Integer &>(*k).as_uint());
```
isuruf: As this is the most important case, move it to the top of if-else

symengine/functions.cpp
```diff
@@ -3270,6 +3270,161 @@ RCP<const Basic> trigamma(const RCP<const Basic> &x)
...
+
+RCP<const Basic> binomial(const RCP<const Basic> &n, const RCP<const Basic> &k)
+{
+    if (is_a_Number(*k) and down_cast<const Number &>(*k).is_zero()) {
+        return one;
+    } else if (is_a_Boolean(*n) or is_a_Boolean(*k) or is_a_Set(*n)
+               or is_a_Set(*k)) {
+        throw SymEngineException(
+            "Boolean and Set objects not allowed in this context.");
+    } else if (eq(*n, *k) and not(is_a<NaN>(*n) or is_a<Infty>(*n))) {
```
isuruf: Move the NaN check to above this, so that you don't have to check here again.

symengine/printer.cpp
```diff
@@ -57,6 +57,21 @@ void StrPrinter::bvisit(const NaN &x)
...
 }
 
+void StrPrinter::bvisit(const Factorial &x)
+{
+    std::ostringstream s;
+    s << "(" << apply(x.get_arg()) << ")!";
+    str_ = s.str();
+}
+
+void StrPrinter::bvisit(const Binomial &x)
```
isuruf: Don't need a special method here. See how `sin` and others are handled

symengine/printer.cpp
```diff
@@ -57,6 +57,21 @@ void StrPrinter::bvisit(const NaN &x)
     str_ = s.str();
 }
 
+void StrPrinter::bvisit(const Factorial &x)
+{
+    std::ostringstream s;
+    s << "(" << apply(x.get_arg()) << ")!";
```
isuruf: `x!` should not be printed as `(x)!`

symengine/functions.cpp
ShikharJ: This is `SymPy`'s output:
```
In [5]: binomial(nan, x)
Out[5]: 
⎛nan⎞
⎜   ⎟
⎝ x ⎠

In [6]: binomial(nan, 3/2)
Out[6]: 
⎛nan⎞
⎜   ⎟
⎝1.5⎠

In [7]: binomial(nan, -3)
Out[7]: 
⎛nan⎞
⎜   ⎟
⎝ -3⎠
```

symengine/printer.cpp
ShikharJ: This was actually done for [these](https://github.com/symengine/symengine/pull/1337/files/7ba71508d9f310d87f9840a87270f133a7d915b5#diff-971819d8abcc9340e783c26a4a53a5cfR569) cases.

isuruf: I mean you should check the precedence of the arg before printing the parenthesis. Take a look at how parenthesis are added. For example, `x**n` vs `(x+y)**n` where there is no parenthesis around `x` in the first case

symengine/functions.cpp
isuruf: If one of the arguments is `NaN`, the value of the function doesn't make sense. Therefore a `NaN` should be returned,

symengine/functions.cpp
```diff
@@ -3270,6 +3270,145 @@ RCP<const Basic> trigamma(const RCP<const Basic> &x)
...
+{
+    if (is_a<Integer>(*n) and is_a<Integer>(*k)) {
+        if (down_cast<const Number &>(*k).is_positive()) {
+            return binomial(down_cast<const Integer &>(*n),
+                            down_cast<const Integer &>(*k).as_uint());
+        } else {
+            return zero;
+        }
+    } else if (is_a_Number(*k) and down_cast<const Number &>(*k).is_zero()) {
+        return one;
```
isuruf: Also return `0` if `k` is a number and `k < 0`.

symengine/functions.cpp
```diff
@@ -3270,6 +3270,145 @@ RCP<const Basic> trigamma(const RCP<const Basic> &x)
...
+}
+
+RCP<const Basic> Binomial::create(const RCP<const Basic> &n,
+                                  const RCP<const Basic> &k) const
+{
+    return binomial(n, k);
+}
+
+RCP<const Basic> binomial(const RCP<const Basic> &n, const RCP<const Basic> &k)
+{
```
isuruf: Here first check that `k` is an integer. If `k` is not an integer, return a `Binomial` object.

sudoWin:
@isuruf Can I take up this issue in a new PR?

I guess only the conflicts need to be corrected.

ShikharJ:
@sudoWin Go for it.

PR #1362, opened 2017-11-12
parser: unary operator fixing

srajangarg:
i really think we need to shift to an autogenerated parser. this is getting out of hands. took me way too much time to read my own code :(

### DIFF
--- a/symengine/parser.cpp
+++ b/symengine/parser.cpp
@@ -16,7 +16,7 @@ public:
     std::map<std::string, int> op_precedence
         = {{")", 0}, {",", 0}, {"|", 1},   {"^", 2},  {"&", 3}, {"==", 4},
            {">", 5}, {"<", 5}, {"<=", 5},  {">=", 5}, {"-", 6}, {"+", 6},
-           {"*", 8}, {"/", 9}, {"**", 10}, {"~", 11}};
+           {"*", 8}, {"/", 9}, {"**", 10}, {"U", 11}, {"~", 12}};
     bool convert_xor_;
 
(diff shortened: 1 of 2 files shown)


srajangarg:
https://github.com/symengine/symengine/issues/1361 @isuruf

certik:
@srajangarg do you want to go ahead and write a generated parser? I've used ANTLR4 before, but I don't know if it would be a good fit here, since it requires a runtime library. One advantage of it is that one just writes a visitor to walk the parse tree, and generate a SymEngine expression from it, and the grammar is kept language independent. We can use Bison, where I don't like that you must hook in the C code in the grammar itself. But in symengine's case, I think the grammar is simple enough, that I think it will be maintainable with Bison. 

There might be other parser generators that we could use.

Ultimately, we'll just have to make sure it's as fast or faster than our current version.

srajangarg:
hi @certik, i've already had a go with flex/bison based parser : https://github.com/symengine/symengine/pull/1211

however this parser sees to be performing slower than the current one. you can check out the discussion on the PR.

certik:
I see, thanks, I'll comment there soon. Yes, as usual, it's a challenge
to make a generated parser as fast as a good handwritten one. That's one
of the reasons why historically production compilers use handwritten
parsers.

srajangarg:
@isuruf can i merge?

isuruf:
Btw, to prevent accidental pushing to upstream I do `git remote set-url upstream --push no_push` so that I can never push to upstream, only pull is allowed.

certik:
Thanks!

PR #1397, opened 2018-02-10
Make ccode print acos(-1) for pi

eeshan9815:
Such cases should be handled at compile-time due to constant folding by compilers
Relevant - #1383


bjodah:
Can you add a test for `jscode` to ensure it still prints `Math.PI`?

bjodah:
Great, thanks!

PR #1416, opened 2018-03-11
Add to CSRMatrix: jacobian, as_vectors & transpose 

bjodah:
Just as in the case of `DenseMatrix`, there was no easy way to access the elements of a `CSRMatrix`.
Also, I couldn't find a function for creating a sparse jacobian matrix, so I went ahead and implemented a static method: `CSRMatrix::jacobian`

Let me know what you think.

### DIFF
--- a/symengine/matrix.h
+++ b/symengine/matrix.h
@@ -279,6 +279,8 @@ class CSRMatrix : public MatrixBase
 public:
     CSRMatrix();
     CSRMatrix(unsigned row, unsigned col);
+    CSRMatrix(unsigned row, unsigned col, const std::vector<unsigned> &p,
+              const std::vector<unsigned> &j, const vec_basic &x);
     CSRMatrix(unsigned row, unsigned col, std::vector<unsigned> &&p,
               std::vector<unsigned> &&j, vec_basic &&x);
 
--- a/symengine/matrix.h
+++ b/symengine/matrix.h
@@ -285,6 +285,9 @@ public:
               std::vector<unsigned> &&j, vec_basic &&x);
     CSRMatrix &operator=(CSRMatrix &&other);
     CSRMatrix(const CSRMatrix &) = default;
+    std::tuple<std::vector<unsigned>, std::vector<unsigned>, vec_basic>
+    as_vectors() const;
+
     bool is_canonical() const;
 
     virtual bool eq(const MatrixBase &other) const;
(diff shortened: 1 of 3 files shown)


bjodah:
For some reason the combination of `BUILD_TYPE=Debug` & `COMPILER=MSVC15` fails (for both Win32 & x64). Tens of thousands of warnings during linking, each looking like this:
```
gmp.lib(assert.obj) : warning LNK4099: PDB 'mpir.pdb' was not found with 'gmp.lib(assert.obj)' or at 'C:\projects\symengine\build\benchmarks\Debug\mpir.pdb'; linking object as if no debug info [C:\projects\symengine\build\benchmarks\add1.vcxproj]
```
and then during the tests it hangs for `test_matrix`:
```
[00:08:42]       Start 24: test_matrix
```
Have I run into a MSVC compiler bug? Or is appveyor using cached objects and failed to identify the need to recompile? The reason for me asking is that I see this line in the log:
```
[00:02:18]   CMake does not need to re-run because C:/projects/symengine/build/symengine/tests/matrix/CMakeFiles/generate.stamp is up-to-date.
```
Not sure if that applies to cached files or to some intermediate stage on a (fresh) checkout.

isuruf:
Appveyor hanging might be due to an error message popping up. You'll have to use rdp to login to the appveyor VM and check

symengine/matrix.h
```diff
@@ -374,6 +383,12 @@ class CSRMatrix : public MatrixBase
         RCP<const Basic> (&bin_op)(const RCP<const Basic> &,
                                    const RCP<const Basic> &));
 
+    std::tuple<std::vector<unsigned>, std::vector<unsigned>, vec_basic>
+    as_vectors() const
+    {
+        return std::make_tuple(p_, j_, x_);
```
isuruf: I'm not sure if `std::make_tuple` will do the correct thing since `p_, j_, x_` are const objects. Maybe this is the reason for the appveyor failure

bjodah: Oh, I didn't think of that. Nice catch. I'll copy construct some locals and std::move them. I'm firing up a RDP build on appveyor in parallel.

bjodah: Same problem still. No pop-up in the RDP server. Will try to reproduce the error locally in a Windows VM.

bjodah:
I was a bit too quick to dismiss the RDP server as not showing a pop up:
screenshot from 2018-03-12 17-41-44

bjodah:
So there was a bug `CSRMatrix::get`, it could read one past end of the vector, but that didn't happen in any of the earlier test cases.

I believe I've rectified it in f80e6e9
Passing `-D_GLIBCXX_DEBUG` to gcc allows it to uncover the same bug.
I will add that macro to some gcc builds on travis in a later PR.

This is ready for review.

PR #1420, opened 2018-03-15
DenseMatrix(vec_basic), CSRMatrix::jacobian(vec_basic, vec_sym)

bjodah:
This is a suggested change to add a constructor to DenseMatrix which accepts vec_basic. I propose that it is interpreted as column vector (I believe it's the convention and e.g. `jacobian` expects column vectors).

I also changed CSRMatrix::jacobian not to strictly require a pair of `DenseMatrix`s, but also accept `vec_basic` and `vec_sym`.

Let me know what you think.


certik:
This is fine with me. @isuruf would you mind reviewing it please?

PR #1427, opened 2018-04-03
WIP: Add initial vscode config files

certik:
I only tested this in Ubuntu 17.10.

TODO:

* [ ] Test on a Mac and Windows (probably the patch in 1869596b9d6287058c30bd9c68d6e46460ec6964 must be applied to the Mac and Windows sections)
* [ ] Write documentation how to use this, mentioning the cmake options (see the git commit message of 1869596b9d6287058c30bd9c68d6e46460ec6964), which extensions to install and so on.
* [ ] Spend some time actually using it, I just started today, and decide later if we want this as part of the official repository. So far it seems to work with this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ cython_test.cpp
 SymEngineConfig.cmake
 SymEngineConfigVersion.cmake
 SymEngineTargets.cmake
+compile_commands.json
 
 # cotire
 symengine/cotire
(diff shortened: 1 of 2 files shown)


certik:
I have been undecided whether the vscode setting should live in the git repository itself, so I just asked at https://github.com/Microsoft/vscode-cpptools/issues/2549.

PR #1477, opened 2018-10-24
Use clang++-7 with Asan & UBsan on Travis

bjodah:
The commits in #1471 was a mess and rebasing was just a mess so adding the diff as a single commit here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,7 @@ matrix:
       install:
         - echo "nothing to install"
       script:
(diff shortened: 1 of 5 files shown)


bjodah:
the address sanitizer correctly caught my contrived buffer-overflow in 8d5a011

bjodah:
once tests pass here I'm happy with this.

bjodah:
Thank you for the review

PR #1497, opened 2019-01-08
New MatchPyCpp package in utilities

Upabjojr:
I think `symengine/utilities` is the optimal location for this package. A Python file is included, which acts as code generator for the SymEngine.

- [ ] ~~move `SymEnginePrinter` to `SymPy`.~~
- [x] optional wildcards testing.
- [ ] ~~sequence-matching wildcard~~ ==> not urgent, open separate issue
- [x] check `var.find(...) == var.end()` for optimizations.
- [x] add more `const ... &` to function parameters.
- [ ] ~~add `typedef map<string, RCP<const Basic>> SubstitutionBasic` ?~~ ==> open separate issue


certik:
@Upabjojr ping me once this is ready for review.

Upabjojr:
Latest updates:

- no more coroutines/generators, they are all expanded to vectors and returned in full.
- pattern matching requiring the commutative property is still not working.

Reasons for discarding the attempts at porting generators to C++:

- Library `coroutine.h` messes up the reference counter of both Teuchos and STL.
- Using `setjmp` gives similar problems.
- Using the block trick is too complicated and is likely to make the code hard for support.
- It looks like _clang_ and _Microsoft Visual C++_ now support an experimental `co_yield` keyword, similar to Python's `yield`. It's still experimental, but hopefully it will get into the C++ standard soon.

asmeurer:
Don't patterns return a combinatorial number of matches in general? For instance matching `a*b*c` against `x*y*z` matches in 3! ways (even more if you allow things like a=1,b=x,c=y*z).  I think you need some way to return one match at a time.

Upabjojr:
Yes, I agree. But I'm getting so much headache implementing coroutines in C++ that I'm giving up for now.

If really needed, we can try multithreading, which is supported by the C++ standard. I fear that even multithreading may cause disturbances to the reference counter.

I just wanna make the commutative matcher work, otherwise this project will never end. Optimizations will be done in the future.

Upabjojr:
The issues with the reference counter are quite serious, sometimes the memory is freed too early, resulting in segmentation faults during the execution.

certik:
I agree, let's get something working, then work on optimizations. Regarding generators in C++, let's do a video call later, so that I can better understand what you are trying to do. In general I would prefer not to use the very latest features, especially if they are not even in the standard yet --- usually there is a way to express the same functionality using older C++. But as you said, we can work on this later, looks like you have a workaround now.

Upabjojr:
> @Upabjojr, do you have a branch with coroutine.h that shows the reference counter issues? If so, I can have a look while you are working on the branch without generators.

Yes, in https://github.com/symengine/symengine/pull/1482

There are various example files:

1. `example_coroutine.cpp`
2. `example_coroutine2.cpp`
3. `example_setjmp.cpp`

In the first two I'm using the `coroutine.h` header, as suggested by you. I get errors about memory not having been freed when using _gdb_.

Example 3 uses _setjmp_, there are also problems.

isuruf:
Thanks @Upabjojr, I do see memory leaks when using `coroutine.h`, but not when using `boost::coroutines`. `coroutine.h` must be doing something wrong here.

Upabjojr:
So maybe we should use the boost library.

isuruf:
Hmm, boost.coroutine is now deprecated in favour of boost.coroutine2, but with boost.coroutine2 there are memory leaks.

isuruf:
Looks like when the stack frame is deleted, the destructors are not called for the C++ objects unless C++20 coroutines are used.

isuruf:
Code generation will have to avoid temporaries or be declared in the `match_root` class if anything other than C++20 coroutines are used.

isuruf:
Pushed a commit to #1482 to make example_coroutine2.cpp work properly. No more memleaks

Upabjojr:
OK thanks.

Upabjojr:
The code generator appears to be working now even in the commutative case. I will try more tests in the next days, but I think the code is now close to an acceptable form.

Upabjojr:
I will temporarily neglect attempts at introducing coroutines as I fear it's going to be a lot of headache. The suggestion by @isuruf can be applied sometime in the future.

symengine/utilities/matchpycpp/.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
isuruf: Can you remove this file? (You can also add this to `.gitignore`)

symengine/utilities/matchpycpp/autogen_tests/test_case001.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <tuple>
+
+generator<tuple<int, Substitution>> match_root(RCP<const Basic> subject)
+{
+    generator<tuple<int, Substitution>> result;
+    Deque subjects;
+    subjects.push_front(subject);
+    Substitution subst0;
+    // State 2194
+    if (subjects.size() >= 1 && subjects[0]->__eq__(*x)) {
```
isuruf: Use `eq(*subjects[0], *x)`

symengine/utilities/matchpycpp/common.h
```diff
@@ -0,0 +1,52 @@
...
+
+const int LEFT = 0;
+const int RIGHT = 1;
+
+typedef vector<Substitution> TEdgeValue;
+
+
+typedef deque<RCP<const Basic>> Deque;
+
+Deque get_deque(RCP<const Basic> expr)
```
isuruf: Use `RCP<const Basic> &expr`

symengine/utilities/matchpycpp/substitution.h
```diff
@@ -0,0 +1,83 @@
...
+        return (x->__hash__()) < (y->__hash__());
+    }
+};
+
+bool operator==(RCP<const Basic> &a, RCP<const Basic> &b)
+{
+    return (a->__hash__()) == (b->__hash__());
+}
+
+typedef multiset<RCP<const Basic>, lessBasic> MultisetOfBasic;
```
isuruf: Use `multiset_basic`

symengine/utilities/matchpycpp/substitution.h
```diff
@@ -0,0 +1,83 @@
...
+        subst[variable_name] = replacement;
+    } else {
+        MultisetOfBasic existing_value = subst.at(variable_name);
+        if (existing_value.size() != replacement.size()) {
+            return 1;
+        }
+        MultisetOfBasic::iterator it1 = existing_value.begin();
+        MultisetOfBasic::iterator it2 = replacement.begin();
+        for (size_t i = 0; i < existing_value.size(); i++) {
+            if ((*it1) != (*it2)) {
```
isuruf: Why is this checking the same value over and over?

symengine/utilities/matchpycpp/substitution.h
```diff
@@ -0,0 +1,83 @@
...
+
+
+struct lessBasic {
+    bool operator()(const RCP<const Basic> &x, const RCP<const Basic> &y) const
+    {
+        return (x->__hash__()) < (y->__hash__());
+    }
+};
+
+bool operator==(RCP<const Basic> &a, RCP<const Basic> &b)
```
isuruf: Where is this needed?

symengine/utilities/matchpycpp/substitution.h
```diff
@@ -0,0 +1,83 @@
...
+
+typedef multiset<RCP<const Basic>, lessBasic> MultisetOfBasic;
+
+typedef map<string, MultisetOfBasic> Substitution;
+
+int try_add_variable(Substitution &subst, const string variable_name,
+                     const MultisetOfBasic &replacement)
+{
+    if (subst.find(variable_name) == subst.end()) {
+        subst[variable_name] = replacement;
```
isuruf: It's better to store the value of `subst.find(variable_name)` and it use it as a hint for insertion. Avoids a second tree traversal

symengine/utilities/matchpycpp/substitution.h
```diff
@@ -0,0 +1,83 @@
...
+    } else {
+        MultisetOfBasic existing_value = subst.at(variable_name);
```
isuruf: Use `const MultisetofBasic &` to avoid copying

symengine/utilities/matchpycpp/substitution.h
```diff
@@ -0,0 +1,83 @@
...
+                return 1;
+            }
+        }
+        return 1;
+    }
+    return 0;
+}
+
+int try_add_variable(Substitution &subst, const string variable_name,
+                     vector<RCP<const Basic>> &replacement)
```
isuruf: `const vector`

certik:
Thanks!

Can you give some example or some documentation how to use this? I would like to play with this.

P.S. You can remove the `.gitkeep` files.

symengine/utilities/matchpycpp/substitution.h
```diff
@@ -0,0 +1,83 @@
...
+{
+    Substitution new_subst; //(*this);
+
+    for (const Substitution &other : others) {
+        for (const pair<string, MultisetOfBasic> &p : other) {
+            int ret = try_add_variable(subst, p.first, p.second);
+            assert(ret == 0);
+        }
+    }
+    return new_subst;
```
isuruf: This is returning an empty object.

symengine/utilities/matchpycpp/.idea/vcs.xml
Upabjojr: I didn't notice I had this file committed.

Upabjojr:
@certik the `README.md` file explains how to use this module.

symengine/utilities/matchpycpp/common.h
```diff
@@ -0,0 +1,52 @@
...
+    Deque d;
+    for (RCP<const Basic> i : expr->get_args()) {
+        d.push_back(i);
+    }
+    return d;
+}
+
+RCP<const Basic> x = symbol("x");
+RCP<const Basic> y = symbol("y");
+RCP<const Basic> z = symbol("z");
```
isuruf: This should be removed and dynamically generated. (The symbols can be created in the `get_match_iter` function as a static local variable

Upabjojr:
The associative property matching is not yet working.

symengine/utilities/matchpycpp/autogen_tests/test_case005.cpp
```diff
@@ -15,6 +15,10 @@
 #include <symengine/utilities/matchpycpp/substitution.h>
 #include <tuple>
 
+RCP<const Basic> w_ = symbol("w_");
+RCP<const Basic> x = symbol("x");
+RCP<const Basic> y = symbol("y");
```
isuruf: These can be moved inside `match_root`
```C++
generator<tuple<int, Substitution>> match_root(RCP<const Basic> subject)
{
  static RCP<const Basic> w_ = symbol("w_");
  static RCP<const Basic> x = symbol("x");
  static RCP<const Basic> y = symbol("y");
```

By declaring them static, they are created the first time the function is called.

symengine/utilities/matchpycpp/autogen_tests/test_case004.cpp
```diff
@@ -0,0 +1,159 @@
...
+    subjects.push_front(subject);
+    Substitution subst0;
+    // State 2207
+    if (subjects.size() >= 1 && is_a<Add>(*subjects[0])) {
+        RCP<const Basic> tmp1 = subjects.front();
+        subjects.pop_front();
+        RCP<const Basic> associative1 = tmp1;
+        string associative_type1 = tmp1->__str__();
+        Deque subjects2 = get_deque(tmp1);
+        CommutativeMatcher2209 *matcher = CommutativeMatcher2209::getInstance();
```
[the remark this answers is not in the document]
Upabjojr: The idea was to implement the singleton pattern, I haven't fixed it yet. Anyways, maybe we don't ned the singleton pattern.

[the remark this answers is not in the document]
isuruf: With the singleton pattern, you lose thread safety in this case

symengine/utilities/matchpycpp/autogen_tests/test_case005.cpp
Upabjojr: They are used both in `match_root` and in `TEST_CASE`, and possibly even in `get_match_iter` of all commutative matchers. Do you want the declaration to be repeated many times?

Alternatively the code can be modified to avoid the creation of symbols. For example, instead of
```c++
eq(*tmp2[0], *y)
```
declare
```c++
tmp2[0]->__str__() == "y"
```

symengine/utilities/matchpycpp/autogen_tests/test_case004.cpp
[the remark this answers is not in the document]
Upabjojr: OK, so no singletons.

symengine/utilities/matchpycpp/autogen_tests/test_case005.cpp
isuruf: Yes. That works. `is_a<Symbol>(*tmp2[0]) and tmp2[0]->__str__() == "y"` is better to avoid expensive `__str__` in the case of large expressions

symengine/utilities/matchpycpp/autogen_tests/CMakeLists.txt
```diff
@@ -1,23 +1,22 @@
...
-include_directories(BEFORE ${symengine_SOURCE_DIR})
-include_directories(BEFORE ${symengine_BINARY_DIR})
-include_directories(BEFORE ${teuchos_SOURCE_DIR})
-include_directories(BEFORE ${teuchos_BINARY_DIR})
-
 include_directories(BEFORE ${catch_SOURCE_DIR})
 
 add_executable(test_case001 test_case001.cpp)
 target_link_libraries(test_case001 symengine catch)
+add_test(test_case001 test_case001)
```
Upabjojr: This file is auto generated. The edits should be done to `generate_tests.py`

Upabjojr:
I have added tests for optional wildcards (which are needed for RUBI). I believe the code is now in a good state and can be prepared for merging (i.e. just need some optimizations and cleanup).

Why are the tests failing? They are passing in my computer.

certik:
It looks like some compiler errors, probably too old compiler on Travis?

isuruf:
Yes, the compiler is old. You are using a C++14 feature. You'll have to do `{std::make_tuple(0, 1), true}` instead of `{{0, 1}, true}` to make it work on pre-C++14 compilers.

symengine/utilities/matchpycpp/substitution.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include <map>
+#include <vector>
+
+using namespace std;
+using namespace SymEngine;
+
+typedef map<string, multiset_basic> SubstitutionMultiset;
+
+int try_add_variable(SubstitutionMultiset &subst, const string variable_name,
```
isuruf: Use `bool` instead of `int`. Makes it easier to read and also convey what you are doing

Upabjojr: It was meant to resemble C's style exception handling. The Python implementation raises an exception.

symengine/utilities/matchpycpp/autogen_tests/test_case001.cpp
```diff
@@ -0,0 +1,59 @@
+/*
+ * This file was automatically generated: DO NOT EDIT.
```
certik: Can we add information here exactly how it was generated and which file should be modified instead?

Upabjojr:
What about merging this PR? There are still some things to do, but the hardest part has been done.

We could suggest porting RUBI's code to SymEngine as part of GSoC after this PR gets merged.

certik:
@Upabjojr do you plan to polish your patches or do you want to merge as is? If "as is", then it would be better if next time you could write better git commit messages than just "updates".

Also, can you please merge or rebase on top of the latest master? To ensure that all tests pass.

Upabjojr:
Do you understand what the test failures are?

certik:
The error here https://travis-ci.org/symengine/symengine/jobs/491286402 seems to be a bug in the code in this PR. Where it's coming from is tough to say. 

@isuruf for cases like these, I think it would be super helpful to get a beautiful stacktrace. The `SymEngine::print_stack_on_segfault();` only prints it on segfault, but I thought there might be a way to hook this machinery up to also get a stacktrace when an exception is not caught.

certik:
Actually maybe it will work, I am testing it here: https://github.com/symengine/symengine/pull/1511

certik:
The stacktrace works!

Here is what I get on my laptop:

https://github.com/symengine/symengine/pull/1511#issuecomment-462172510

@Upabjojr is the stacktrace enough for you to figure out what is going on?

Upabjojr:
> @Upabjojr is the stacktrace enough for you to figure out what is going on?

Hopefully, there was an extra line I forgot to remove at that point in the stacktrace.

certik:
Good, that error is now fixed. That leaves the following error as the last thing to fix:

https://travis-ci.org/symengine/symengine/jobs/491377210#L1654

And in there it has a stacktrace and the error comes from this line:
```c++
        new_match[*left2] = right;
```
And the stacktrace also shows it comes from `operator[]`, so it seems as if `*left2` points to something that is out of bounds for `new_match`? Or something like that.

Looking at your code, I see some problematic parts. This line:
```c++
                        TLeft node2 = get<1>(p2);
```
allocates a stack variable `node2`, assigns something to it (by value I assume?), then you pass a reference to it into `left2`:
```c++
                            left2 = &node2;
```
and then you dereference `left2`, but `node2` is already out of scope, in other words, the pointer is dangling. So this seems to be a bug.

@Upabjojr let me know if you need more help with this.

symengine/utilities/matchpycpp/autogen_tests/test_case001.cpp
certik: This has been added below.

certik:
After that bug is fixed, my other worry is that `cpp_code_generation.py` is not tested. I would test it on a CI by generating the automatic tests directory and compare the result against the files in the git repository and they need to match. That way if we change `cpp_code_generation.py`, the CI will alert us that we have to regenerate the files, and obviously it will test that things actually work.

To run it, by looking at the `README.md`, it seems you have to have some particular version of matchpy installed, and also sympy and symengine. It looks like there are a lot of moving parts, and I would like them to be tested on CI, to also serve as a documentation how to actually get things up and running.

Finally, can you please write some introduction into the `README.md` about what the intended purpose of `matchpycpp` is, what the goal is, what the current status is, and what the plan is, so that if somebody reads it, who doesn't know anything about `matchpy`, can understand what is going on?

In the same manner it would be nice to document some of the classes in the new code, for example the `HopcroftKarp` class, or the `BipartiteGraph` class. I have no idea what those classes are doing, or how they are supposed to be used.

Upabjojr:
> After that bug is fixed, my other worry is that `cpp_code_generation.py` is not tested. I would test it on a CI by generating the automatic tests directory and compare the result against the files in the git repository and they need to match.

It's slightly more complicated, as `cpp_code_generation.py` only generates the decision tree. The files in `autogen_tests` are generated by `generate_tests.py`, which adds the test cases.

One simple solution is to check that the last modification date of the generated tests is newer that the last modification of `cpp_code_generation.py`.

In MatchPy, they test the generated code by compiling it:
https://github.com/HPAC/matchpy/blob/master/tests/test_code_generation.py

What about running `generate_tests.py` before Travis CI compiles the code? Unfortunately it would need some Python dependencies (MatchPy and SymPy).

Otherwise simple code comparison can also work.

Upabjojr:
How do I add a test to call a Python script here?

Should we rather move the Python files to SymPy? `cpp_code_generator.py` was copied from MatchPy and needs to keep the LICENSE.

certik:
@Upabjojr I don't have an opinion on if it should be in sympy or not. If you feel strongly one way or another, go ahead. Otherwise let's get this PR ready.

Let's start by getting all the tests to pass.

Upabjojr:
Hi,

there is a current attempt at porting MatchPy's code generator to SymEngine: https://github.com/symengine/symengine/pull/1497

In the future this could be the start of an attempt at porting RUBI to C++ by transforming the rules into a huge decision tree in a C++ file.

The current main issue is how to handle generators in C++, as of now they solve for all matching combinations before returning the result, which may be inefficient in case the space of combinations is large. There are various possible solutions to handle this problem, so it should not be a big issue.

I have also identified the problem with the MatchPy code generator which has not allowed to generate a meaningful decision tree in GSoC 2018. I have an open PR in MatchPy, just need to add a test there: https://github.com/HPAC/matchpy/pull/50

Generating a decision tree will solve the problem of the loading time (it takes a lot to load all RUBI rules in SymPy).

We should also update the link to MatchPy, as SymPy's iink is incompatible with MatchPy's current master branch.

Upabjojr:
I got to fix the tests. The Python files still untested. Testing them would require installing a whole Python environment with SymPy and MatchPy dependencies. Maybe these files should eventually be moved to SymPy.

Anyway, I guess it's time to prepare for merging.

isuruf:
> Testing them would require installing a whole Python environment with SymPy and MatchPy dependencies

We are using conda to install the C++ dependencies, so it's not hard to install a python environment. Can you create an environment.yml file in matchpygen directory with all the packages you need?

Upabjojr:
Ok I can try.

Upabjojr:
OK, so now the tests run `generate_tests.py` and check for `git diff` not to show any difference. Is that correct?

If that's correct, I believe we can merge this branch.

bin/test_matchpycpp_gen_tests.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
certik: You have to put `set -ex` here. The `-e` is to fail the bashscript when one of the commands fail, and the `-x` is to show the commands, so that one can see in the Travis log what was happening (if something fails).

certik: Fixed in 7d3cbf5a62f3fe9faa97bd83b8f6c1d65c4315bf.

certik:
I am checking that the tests fail in #1515.

certik:
It seems to work now, the tests pass and fail properly. I am +1 to merge this after all tests pass.

Upabjojr:
OK, let's merge so that we can close the NumFOCUS project.

Upabjojr:
@certik is this PR going to be merged?

certik:
@Upabjojr sorry, I was waiting for the tests to pass, but forgot about it. I wish there was a notification for that.

isuruf:
There's https://mergify.io/ where you can set a label `merge-when-green` on a PR and a bot will merge when the CI passes.

asmeurer:
We also considered adding a feature like that to the SymPy bot. The problem is that then the merge commit comes from the bot, instead of the person who merged it. I wish GitHub would just implement this natively, so you could click "merge when the tests pass" and as long as the tests don't fail, new commits aren't pushed, or someone else doesn't cancel it, it will automerge once all the tests pass.

jondo:
FWIW, this auto-merge feature [exists](https://docs.gitlab.com/ee/user/project/merge_requests/merge_when_pipeline_succeeds.html) in GitLab.

isuruf:
Mergify can comment on the PR instead of merging it. Also the SymPy bot could merge with the author's name and email id, but the bot needs to have a way to associate github handle with an email.

asmeurer:
That would require disabling branch protection. Right now no one can push directly to master. The branch protection forces every commit to come from a merged pull request (with passing status).

certik:
@jondo yes, I've been using GitLab for new projects for a few years precisely because of little things like this. But I don't want to move SymPy or SymEngine, as the pain of moving doesn't seem worth the benefits (yet). Plus GitLab has some things that seem a bit worse, like searching for projects, smaller community, etc.

asmeurer:
I've included it in my blog post of things I'd like to see GitHub improve ([number 15](https://www.asmeurer.com/blog/posts/github-cuts/)). I've gotten responses from people at GitHub so I know they've seen it. Let's see how many things they implement. GitHub seems to be doing better with this sort of thing since they got acquired by Microsoft and got a new CEO (at least now they are more. I'm curious how responsive GitLab is to these sorts of requests, relative to GitHub.

certik:
@asmeurer, GitLab has these feature requests as public issues, so you can see the progress there. Anybody can help implementing the features. GitLab itself is usually very slow on things that are not part of their roadmap, but eventually it does seem to come around. Here is an example of a long standing issue in GitLab (something that GitHub had for a long time): https://gitlab.com/gitlab-org/gitlab-ee/issues/9713, and looks like they are getting around to it, but it will be first implemented in the paid version only (which is usually what they do first, and later they move some features from paid to free). Here is their list of issues sorted by popularity in the free and paid versions:

https://gitlab.com/gitlab-org/gitlab-ce/issues?sort=popularity
https://gitlab.com/gitlab-org/gitlab-ee/issues?sort=popularity

That's nice about it, that you can see exactly what they are working on, as well as what the popular requests are and the status of your own requests:

https://gitlab.com/gitlab-org/gitlab-ce/issues?scope=all&state=all&author_username=certik

although sometimes they don't have the manpower to implement everything themselves and rather use 3rd party libraries that don't do things exactly as I would like:

https://gitlab.com/gitlab-org/gitlab-ce/issues/37485

Upabjojr:
Coroutines have been approved for the C++20 standard:
https://www.reddit.com/r/cpp/comments/au0c4x/201902_kona_iso_c_committee_trip_report_c20/

PR #1574, opened 2019-07-29
Add ignore rule for CLion, and ignore rule for cmake-build-xxx

stevenleeS0ht:


stevenleeS0ht:
@certik , could you please give me a review on this PR?

There is a number of people who use CLion as there C/C++ development tool. Add related rule to exclude the setting and cache files from CLion can efficiently avoid unnecessary chaos.

certik:
Thanks!

PR #1575, opened 2019-07-30
WIP: use traditional flex

certik:
This PR changes from `flexcpp` to `flex`.


certik:
I created a simple benchmark:
```c++
    std::string text;
    std::string t0 = "(x + y - sin(x)/(z**2-4) - x**(y**z))";
    text = t0;
    for (int i = 0; i < 5000; i++) {
        text = text + " * " + t0;
    }
    N = 1;
    for (int i = 0; i < N; i++) {
        a = parse(text);
    }
```
Results for optimized build:
* master: 87ms
* this PR: 69ms

certik:
Note: one can make the tokenizer much smaller (in terms of lines of code) by:
```diff
--- a/symengine/parser/tok.l
+++ b/symengine/parser/tok.l
@@ -2,8 +2,6 @@
 %option header-file="tok.h"
 %option noyywrap
 %option nounput
-%option fast
-%option 8bit
 
 %{
 #include "parser.h"
```
The timing seems a bit slower: 72s. Still much faster than our current tokenizer.


Note that this uses the same parser, only the tokenizer changed.

certik:
The code that flex generates is not thread safe (has global variables, etc.). One can turn on being re-entrant. Overall, the generated code seems quite messy. I am not super happy with the user experience.

But it's faster, which is what I was suspecting when I saw the layers and layers of C++ code in flexcpp.

I am now going to try re2c, which I am hoping will be even faster and simpler (nicer code).

certik:
Closing in favor of #1588.

PR #1584, opened 2019-08-01
WIP: Use re2c for tokenizer

certik:
Here are benchmarks on my laptop:

old parser (4f7cd0e74901e4e6cabcfab2cdcbd72c85d455ef): 51ms  
master (4f7cd0e74901e4e6cabcfab2cdcbd72c85d455ef): 40ms  
flex (780f6662a1064b480933c49d88754a26be0c8fb8): 30ms  
re2c (a67249fe336176d5f42888df4c9bd8768b984cf1): 32ms


certik:
@isuruf let me know what you think. Based on this, I would choose re2c. It's much simpler, the generated code is cleaner and shorter than Flex, and it's thread-safe.

certik:
Here is the Flex code: https://github.com/symengine/symengine/pull/1575

certik:
On my other machine, the timings are:

old parser (4f7cd0e74901e4e6cabcfab2cdcbd72c85d455ef): 106ms
master (4f7cd0e74901e4e6cabcfab2cdcbd72c85d455ef): 88ms
flex (780f6662a1064b480933c49d88754a26be0c8fb8): 69ms
re2c (a67249fe336176d5f42888df4c9bd8768b984cf1): 72ms

So it's consistent. The re2c parser is very simpler and much faster than master. So I am going to polish this and prepare a clean PR. The flex parser seems to be a tiny bit faster, but it's a lot messier, and not thread safe, so that can be an option we can contemplate later.

certik:
re2c has optimizations options that I haven't used, it can rewrite switches to if statements or jump tables. With jump tables it's very similar to how flex works I think. I am confused with the new timings. I noticed that sometimes I get only 54ms now. And sometimes around 70ms. I think the CPU runs at higher frequency sometimes. It seems with the jump tables (`-g` option) it is consistently faster than without it.

bjodah:
Perhaps likwid-setFrequencies can help? (a laptop may still throttle if thermally constrained though...)
https://github.com/RRZE-HPC/likwid/wiki/likwid-setFrequencies

certik:
@bjodah nice, have to look at it. The machine I used above is a desktop machine, but it might work.

certik:
I further simplified the re2c tokenizer in #1586 and I am now getting 69ms, the same as flex with the `-b` option. With `-g` I am actually getting 67ms, but  it fails to compile on MSVC, so I'll keep the  `-b` option.

certik:
Closing in favor of #1588.

PR #1612, opened 2019-09-11
LLVMDoubleVisitor init is slow, added example as benchmark

ichumuh:
Hi guys,

similar to diff on expressions with common sub expressions, llvm init is also very slow.
Since you managed to make diff super fast for my usecase, I was wondering if the same would be possible for llvm init?
The benchmark is an example, similar to `diff_cache`.
My output is
```
$ ./benchmarks/llvm_diff_cache 
Time for jacobian with common subexpressions (with cache) : 	     0.108437747
Time for llvm init with common subexpressions             : 	    12.805364401
Time for jacobian without common subexpressions (without cache) : 	     3.684057946
Time for llvm init without common subexpressions                : 	     2.156445590
```
Thanks!

### DIFF
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -85,3 +85,9 @@ target_link_libraries(parsing symengine)
 
 add_executable(diff_cache diff_cache.cpp)
 target_link_libraries(diff_cache symengine)
+
+if (WITH_LLVM)
+    add_executable(llvm_diff_cache llvm_diff_cache.cpp)
+    target_link_libraries(llvm_diff_cache symengine)
+endif()
+
--- a/benchmarks/llvm_diff_cache.cpp
+++ b/benchmarks/llvm_diff_cache.cpp
@@ -59,27 +59,20 @@ double CommonSubexprLLVM()
         << std::setw(15) << std::setprecision(9) << std::fixed
         << std::chrono::duration<double>(t2 - t1).count() << std::endl;
 
-
     t1 = std::chrono::high_resolution_clock::now();
 
     LLVMDoubleVisitor v3;
     bool symbolic_cse = true;
     int opt_level = 3;
-    v3.init(v, {
-        J.get(0,0), 
-        J.get(0,1),
-        J.get(0,2),
-        J.get(0,3),
-        J.get(0,4),
-        J.get(0,5),
-        J.get(0,6),
-        J.get(0,7),
-        J.get(0,8),
-        J.get(0,9)}, symbolic_cse, opt_level);
+    v3.init(v,
+            {J.get(0, 0), J.get(0, 1), J.get(0, 2), J.get(0, 3), J.get(0, 4),
+             J.get(0, 5), J.get(0, 6), J.get(0, 7), J.get(0, 8), J.get(0, 9)},
+            symbolic_cse, opt_level);
     t2 = std::chrono::high_resolution_clock::now();
-    std::cout << "Time for llvm init with common subexpressions (with cache) : \t "
-              << std::setw(15) << std::setprecision(9) << std::fixed
-              << std::chrono::duration<double>(t2 - t1).count() << std::endl;
+    std::cout
+        << "Time for llvm init with common subexpressions (with cache) : \t "
+        << std::setw(15) << std::setprecision(9) << std::fixed
+        << std::chrono::duration<double>(t2 - t1).count() << std::endl;
     return 0;
 }
 
@@ -110,30 +103,23 @@ double NoCommonSubexprLLVM()
     jacobian(M, S, J, false);
     auto t2 = std::chrono::high_resolution_clock::now();
 
-    std::cout
-        << "Time for jacobian without common subexpressions (without cache) : \t "
-        << std::setw(15) << std::setprecision(9) << std::fixed
-        << std::chrono::duration<double>(t2 - t1).count() << std::endl;
-
+    std::cout << "Time for jacobian without common subexpressions (without "
+                 "cache) : \t "
+              << std::setw(15) << std::setprecision(9) << std::fixed
+              << std::chrono::duration<double>(t2 - t1).count() << std::endl;
 
     t1 = std::chrono::high_resolution_clock::now();
 
     LLVMDoubleVisitor v3;
     bool symbolic_cse = true;
     int opt_level = 3;
-    v3.init(v, {
-        J.get(0,0), 
-        J.get(0,1),
-        J.get(0,2),
-        J.get(0,3),
-        J.get(0,4),
-        J.get(0,5),
-        J.get(0,6),
-        J.get(0,7),
-        J.get(0,8),
-        J.get(0,9)}, symbolic_cse, opt_level);
+    v3.init(v,
+            {J.get(0, 0), J.get(0, 1), J.get(0, 2), J.get(0, 3), J.get(0, 4),
+             J.get(0, 5), J.get(0, 6), J.get(0, 7), J.get(0, 8), J.get(0, 9)},
+            symbolic_cse, opt_level);
     t2 = std::chrono::high_resolution_clock::now();
-    std::cout << "Time for llvm init without common subexpressions (without cache) : \t "
+    std::cout << "Time for llvm init without common subexpressions (without "
+                 "cache) : \t "
               << std::setw(15) << std::setprecision(9) << std::fixed
               << std::chrono::duration<double>(t2 - t1).count() << std::endl;
     return 0;
@@ -146,4 +132,3 @@ int main(int argc, char *argv[])
     NoCommonSubexprLLVM();
     return 0;
 }
-
--- a/benchmarks/llvm_diff_cache.cpp
+++ b/benchmarks/llvm_diff_cache.cpp
@@ -1,0 +1,149 @@
+#include <iostream>
+#include <chrono>
+#include <iomanip>
+
+#include <symengine/basic.h>
+#include <symengine/add.h>
+#include <symengine/symbol.h>
+#include <symengine/integer.h>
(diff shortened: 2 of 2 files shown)


certik:
@ichumuh thanks for writing a test case. What is the difference in the llvm init? Is the only difference that in the first case you give it Jacobian that is using a cache, and in the second case it is not using a cache? Or is there another difference.

Why would it be 6x slower? Any ideas?

ichumuh:
I'm only using cache in the 1. test and no cache in the 2. because its the faster way to calculate the jacobian, it doesn't influence llvm init. My goal was to highlight the similarity to https://github.com/symengine/symengine/pull/1592.
There it was also the case that a lot of common sub expressions slowed `diff` down. Maybe its something similar?

certik:
What I don't understand is that just by switching to use a cache to compute the Jacobian it somehow makes the LLVM init slower. It seems the LLVM init sees exactly the same expression in both cases. Although in the first case, some subexpressions are reused (have higher reference count).

We just need to debug this to see what is happening.

ichumuh:
So I'm kind of a noob with callgrind.
I did this to my code (didn't get `--toggle-collect=<function>` to run):
```
[...]
    CALLGRIND_START_INSTRUMENTATION;
    v3.init(v, {
        J.get(0,0), 
        J.get(0,1),
        J.get(0,2),
        J.get(0,3),
        J.get(0,4),
        J.get(0,5),
        J.get(0,6),
        J.get(0,7),
        J.get(0,8),
        J.get(0,9)}, symbolic_cse, opt_level);
    CALLGRIND_STOP_INSTRUMENTATION;
[...]
```
and ran 
```valgrind --tool=callgrind --instr-atstart=no ./benchmarks/llvm_diff_cache```
I didn't know how to interpret the output, but google told me how I can turn it into a png.
I guess this means that ```llvm::MCJIT::finalizeObject()``` takes the most time?
graph2

**Comment by isuruf** [comment]
I usually use `kcachegrind` to interpret the results.

**Comment by isuruf** [comment]
Looking at the png, I don't think there's much we can do. You can try reducing the optimization level, but that would affect the runtime.

**Comment by bjodah** [comment]
We're currently using [essentially](https://llvm.org/doxygen/namespacellvm_1_1CodeGenOpt.html) "-O3" here:
https://github.com/symengine/symengine/blob/4ea4ba72ef45976a1efe20858e9642e9d0483939/symengine/llvm_double.cpp#L294

We should probably use our option "opt_level" to set this in accordance. Perhaps also expose `opt_level` in the python bindings.

But as Isuru points out, it will have an effect on runtime. (If one really wanted, one could probably build a two-tier backend to do some latency hiding by compiling a llvm-callback in one thread and and using a LambdaDoubleVisitor in the main thread until the compilation is done)

**Comment by certik** [comment]
Alternatively we can precompute the Lambda and save the optimized LLVM IR, or perhaps even the compiled object code, so that one can precompute it (that will take some time), but then the second time it can get loaded immediately.

**Comment by isuruf** [comment]
> Alternatively we can precompute the Lambda and save the optimized LLVM IR, or perhaps even the compiled object code, so that one can precompute it (that will take some time), but then the second time it can get loaded immediately.

This is already possible with the `dumps` and `loads` functions in `LLVMVisitor`.

**Comment by ichumuh** [comment]
> You can try reducing the optimization level, but that would affect the runtime.

Reducing the optimization level does not significantly influence the runtime.
`opt_level=2`
```
$ ./benchmarks/llvm_diff_cache 
Time for jacobian with common subexpressions (with cache) : 	     0.110041993
Time for llvm init with common subexpressions (with cache) : 	    12.312735768
```
`opt_level=1`
```
$ ./benchmarks/llvm_diff_cache 
Time for jacobian with common subexpressions (with cache) : 	     0.109979281
Time for llvm init with common subexpressions (with cache) : 	    12.704904674
```
`opt_level=0`
```
$ ./benchmarks/llvm_diff_cache 
Time for jacobian with common subexpressions (with cache) : 	     0.112378036
Time for llvm init with common subexpressions (with cache) : 	    19.536180906
```

> This is already possible with the dumps and loads functions in LLVMVisitor.

That's what I'm currently doing. So it seems like there is not much that can be done. Thanks for looking into it. You can close the PR if you'd like.

**Comment by ichumuh** [comment]
```
$ ./benchmarks/llvm_diff_cache 
opt_level 0
Time for jacobian with common subexpressions (with cache) : 	     0.094360521
Time for llvm init with common subexpressions (with cache) : 	     4.562975447
Time for jacobian without common subexpressions (without cache) : 	     0.000480132
Time for llvm init without common subexpressions (without cache) : 	     0.011323030
opt_level 1
Time for jacobian with common subexpressions (with cache) : 	     0.087834437
Time for llvm init with common subexpressions (with cache) : 	    13.471997591
Time for jacobian without common subexpressions (without cache) : 	     0.000573569
Time for llvm init without common subexpressions (without cache) : 	     0.000693913
opt_level 2
Time for jacobian with common subexpressions (with cache) : 	     0.089672255
Time for llvm init with common subexpressions (with cache) : 	    13.473093269
Time for jacobian without common subexpressions (without cache) : 	     0.000649316
Time for llvm init without common subexpressions (without cache) : 	     0.000696325
opt_level 3
Time for jacobian with common subexpressions (with cache) : 	     0.088931718
Time for llvm init with common subexpressions (with cache) : 	    13.413338506
Time for jacobian without common subexpressions (without cache) : 	     0.000608762
Time for llvm init without common subexpressions (without cache) : 	     0.000846534
```
I was also trying to compare how much time `call` takes, but there doesn't seem to be an implementation for this case? I basically need one that returns a list and not a single double, but that doesn't exist? I wonder why it works in symengine.py.

**Comment by bjodah** [comment]
There are two signatures for call, one takes two pointers where the non-const pointer is the "output array" (you need to allocate for it using e.g. malloc or making a std::vector<double> and passing its `.data()`.

**Comment by ichumuh** [comment]
@bjodah Thanks.
```
$ ./benchmarks/llvm_diff_cache 
opt_level 0
Time for jacobian with common subexpressions: 	     0.108219366
Time for llvm init with common subexpressions: 	     4.474706773
Time for llvm call with common subexpressions: 	     0.000230163
Time for jacobian without common subexpressions: 	     3.747388515
Time for llvm init without common subexpressions: 	     1.289677737
Time for llvm call without common subexpressions: 	     0.000236636
opt_level 1
Time for jacobian with common subexpressions: 	     0.090473821
Time for llvm init with common subexpressions: 	    13.098013865
Time for llvm call with common subexpressions: 	     0.000173625
Time for jacobian without common subexpressions: 	     3.674321327
Time for llvm init without common subexpressions: 	     2.199979134
Time for llvm call without common subexpressions: 	     0.000197330
opt_level 2
Time for jacobian with common subexpressions: 	     0.087502084
Time for llvm init with common subexpressions: 	    13.013300478
Time for llvm call with common subexpressions: 	     0.000143014
Time for jacobian without common subexpressions: 	     3.663375264
Time for llvm init without common subexpressions: 	     2.195645444
Time for llvm call without common subexpressions: 	     0.000290689
opt_level 3
Time for jacobian with common subexpressions: 	     0.087973188
Time for llvm init with common subexpressions: 	    13.013416809
Time for llvm call with common subexpressions: 	     0.000143401
Time for jacobian without common subexpressions: 	     3.704199721
Time for llvm init without common subexpressions: 	     2.204054393
Time for llvm call without common subexpressions: 	     0.000178715
```

**Comment by ichumuh** [comment]
with `LambdaRealDoubleVisitor`:
```
$ ./benchmarks/llvm_diff_cache 
Time for jacobian with common subexpressions: 	     0.112163930
Time for llvm init with common subexpressions: 	     3.560271043
Time for llvm call with common subexpressions: 	     0.002026663
Time for jacobian without common subexpressions: 	     3.660975644
Time for llvm init without common subexpressions: 	     0.336053675
Time for llvm call without common subexpressions: 	     0.000473094
```

**Comment by bjodah** [comment]
Great.

So it looks like opt_level=0 is a win for your use case here? (am I reading you timings correctly?)

I see the timings for call is in the sub millisecond range. You said you did "hundreds" of evaluations for each init? Perhaps time a few a loop of a few hundred calls then? (it would also make the comparison with the LambdaRealDoubleVisitor more fair?)

**Comment by ichumuh** [comment]
```
$ ./benchmarks/llvm_diff_cache 
opt_level 0
Time for llvm init with common subexpressions: 	     4.452341642
Time for 500 llvm calls with common subexpressions: 	     0.056452646
Time for llvm init without common subexpressions: 	     1.291209508
Time for 500 llvm calls without common subexpressions: 	     0.078078626
opt_level 1
Time for llvm init with common subexpressions: 	    12.994182719
Time for 500 llvm calls with common subexpressions: 	     0.043232439
Time for llvm init without common subexpressions: 	     2.167738079
Time for 500 llvm calls without common subexpressions: 	     0.062677866
opt_level 2
Time for llvm init with common subexpressions: 	    12.942163120
Time for 500 llvm calls with common subexpressions: 	     0.043223322
Time for llvm init without common subexpressions: 	     2.175001043
Time for 500 llvm calls without common subexpressions: 	     0.061535840
opt_level 3
Time for llvm init with common subexpressions: 	    12.931606667
Time for 500 llvm calls with common subexpressions: 	     0.045553398
Time for llvm init without common subexpressions: 	     2.229260492
Time for 500 llvm calls without common subexpressions: 	     0.061938380
=================================
Time for lambda init with common subexpressions: 	     3.681221311
Time for 500 lambda calls with common subexpressions: 	     0.547503499
Time for lambda init without common subexpressions: 	     0.331109800
Time for 500 lambda calls without common subexpressions: 	     0.118502574
```

> So it looks like opt_level=0 is a win for your use case here? (am I reading you timings correctly?)

Yes! Exposing `opt_level` in the python wrappers would be great.

PR #1617, opened 2019-09-27
[WIP]Add apart algorithm for polynomials

jmig5776:
This PR tends to add apart algorithm corresponding to polynomials just like as we have in sympy but with better benchmarks.
More information will be added soon.

Please suggest ideas or guidance for it.
CC @certik @isuruf

### DIFF
--- a/symengine/polys/apart.cpp
+++ b/symengine/polys/apart.cpp
@@ -1,0 +1,17 @@
+#include <symengine/basic.h>
+
+using SymEngine::RCP;
+using SymEngine::Basic;
+
+namespace SymEngine
+{
+  RCP<const Basic> apart(const RCP<const Basic> &f)
+  {
+    if (is_a_Atom(*f))
+    {
+      return f;
+    }
+    //work pending here
+    return f;
+  }
+}
--- a/symengine/polys/apart.cpp
+++ b/symengine/polys/apart.cpp
@@ -5,20 +5,18 @@ using SymEngine::RCP;
 
 namespace SymEngine
 {
-RCP<const Basic> apart(const RCP<const Basic> &self, bool full=false)
+RCP<const Basic> apart(const RCP<const Basic> &self, bool full = false)
 {
     umap_basic_num gens = _find_gens_poly(self);
 
-
-    if (gens.size() != 1)
-    {
-      //only considering univariate here
-      return self;
+    if (gens.size() != 1) {
+        // only considering univariate here
+        return self;
     }
     RCP<const Basic> var = gens.begin()->first;
 
     RCP<const UExprPoly> poly = from_basic<UExprPoly>(self, var);
-    std::cout<<poly->get_poly();
+    std::cout << poly->get_poly();
 
     return self;
 }
--- a/symengine/polys/apart.cpp
+++ b/symengine/polys/apart.cpp
@@ -5,13 +5,12 @@ using SymEngine::Basic;
 
 namespace SymEngine
 {
-  RCP<const Basic> apart(const RCP<const Basic> &f)
-  {
-    if (is_a_Atom(*f))
-    {
-      return f;
+RCP<const Basic> apart(const RCP<const Basic> &f)
+{
+    if (is_a_Atom(*f)) {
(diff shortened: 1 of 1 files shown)


jmig5776:
I created this PR because I have a questions regarding it. Please bear me. Can anyone tell me how exactly I can test my code, i am certainly able to test it but i want to know all procedure to test after editing the code so that i should not miss anything.

And can anybody tell me how I should search some already built function just like i found is_A_Atom here.

jmig5776:
@isuruf can you please please advice me here that how should I frame this class and I went through the expand.cpp also.

jmig5776:
Should it also be based on visitor design pattern?

certik:
@jmig5776 thanks for trying to write a prototype. As a prototype, you can use any design pattern you want, the goal is to get something little working. Then we'll work on designing it well.

To get something little working, try to write a function that accepts a SymEngine expression, converts it to a univariate polynomial (there is code to do that in SymEngine already) and then extracts the coefficients and prints them on the screen. Once you have that, please ping me, and we'll go from there. (The next step would be to try to write apart() for that simple case.)

jmig5776:
Okay I am on it.

jmig5776:
ping @certik 
In this the input is basic and it identifies the generators and the polynomial with single generator gets converted to UExprPoly.

jmig5776:
@certik I think we might need cancel function here to extract the common parts and then rationalise and make them Add and then multiplying with the common from cancel. What do you think? What should be the next step according to you?

certik:
Good progress, now you have the `poly`. Now we are at this part of SymPy's apart: https://github.com/sympy/sympy/blob/05f530129287da71600630b6e42f327150943723/sympy/polys/partfrac.py#L128.

Here are the next steps:

1. Figure out how to extract the coefficients from `poly`.
2. Implement `cancel`.
3. Implement `div`.
4. Implement `apart_undetermined_coeffs`

And that's pretty much it. Then it just needs to be assembled back into `Basic`. However, each of these functions will take some time to implement. What you can do is take the simplest of these, perhaps `div` and see if you can implement that, and get the same results as SymPy. That might not be that difficult.

certik:
You can also try to implement it from here directly, to get something working: https://en.wikipedia.org/wiki/Partial_fraction_decomposition#Procedure

You also need `factor_list` (factorizes the polynomial), no matter if you translate the sympy's algorithm or write your own.

So I would start with `factor_list` (and `factor`). To implement `factor` is a lot of work. I suggest we use a library for that. Flint (that SymEngine can optionally link to) can do that, here is an example:

https://github.com/wbhart/flint2/blob/c656ed9bcee6264a83924ea882094c78d6ac000d/fmpz_poly_factor/test/t-factor.c#L47

So let's start by writing a `factor()` function that takes `Basic`, and internally calls Flint to do it, and converts back to give a list of factors (or a `Mul` instance).

certik:
In fact SymEngine already implements `factors()`, here is how to use it: https://github.com/symengine/symengine/blob/d0b1932b4ffaf4b364d4b436a796b456917e079b/symengine/tests/polynomial/test_uintpoly_flint.cpp#L320

certik:
Here is how to do division using Flint: https://github.com/symengine/symengine/blob/d0b1932b4ffaf4b364d4b436a796b456917e079b/symengine/polys/uintpoly_flint.h#L214

So try to use these operations (division and factors) to implement the apart using the algorithm at https://en.wikipedia.org/wiki/Partial_fraction_decomposition#Procedure. I think you just construct a linear matrix using the factors, and solve it (SymEngine can do that) to determine the coefficients, and you are set.

jmig5776:
Okay I am on it . Soon will be sending a PR.

certik:
Thanks for trying to figure it out.

jmig5776:
@certik there is no need to thanks. As a open sourcer I want to contribute to make it fastest but truthfully earlier i didn't know development in c++ so in previous two weeks I learnt all the syntaxes and wrote proposal and read about grobner basis. So I might be slow now but will take up the pace in one or two weeks

PR #1628, opened 2019-12-06
Add license for cmake/cotire.cmake

Marlin-Na:
See https://github.com/symengine/symengine/blob/master/cmake/cotire.cmake


Marlin-Na:
This file is not compiled into library, but as a build time dependency. Not sure if we should include it in `LICENSE`. Any opinion @certik ?

PR #1636, opened 2020-01-27
add support for LLVM 10

lkeegan:
- requires c++14: add WITH_CPP14 option to CMake
  - llvm::make_unique has been removed: alias it to std::make_unique

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,8 +34,13 @@ endif()
 
 # Enable C++11 support in all compilers. SymEngine will not compile unless
 # the C++11 support is enabled.
-if (CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang|Intel")
-    set(CXX11_OPTIONS "-std=c++11")
+set(WITH_CPP14 no CACHE BOOL "Build with C++14 support")
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang|Intel")
+    if(WITH_CPP14)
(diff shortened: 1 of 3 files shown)


certik:
@lkeegan thank you! 

Do you know why the `make_unique` must be defined in the `llvm` namespace? That seems discouraged I think to define new things in other namespaces, isn't it? Is it SymEngine or LLVM that requires it? If you can describe the actual problem, I think there might be a better solution.

lkeegan:
@certik llvm previously used c++11, which doesn't have `std::make_unique`, so they implemented their own `llvm::make_unique`, which you use here:

https://github.com/symengine/symengine/blob/fc05f8d55915c2de956e7797d764eb1116b61711/symengine/llvm_double.cpp#L194

https://github.com/symengine/symengine/blob/fc05f8d55915c2de956e7797d764eb1116b61711/symengine/llvm_double.cpp#L960

From llvm version 10 they are using c++14, which does have `std::make_unique`, so they have removed `llvm::make_unique`.

So the problem is just that `llvm::make_unique` should be replaced with `std::make_unique` when using llvm 10.

certik:
@lkeegan please ping me once tests pass, and we can merge it.

lkeegan:
ping @certik

certik:
Thanks @lkeegan, I really appreciate your PR. Let us know if anything else does not work.

lkeegan:
thanks @certik !

PR #1647, opened 2020-03-13
Mark LambdaDoubleVisitor::call const

bjodah:
I believe this could be marked const?


bjodah:
Never mind, CSE modifies it.

PR #1668, opened 2020-05-21
Return SYMENGINE_NOT_IMPLEMENTED in basic_solve_poly

Marlin-Na:


rikardn (commented):
I can argue for changing this, but also for keeping it. @Marlin-Na why do you think this should be changed?

rikardn:
It could make sense to do this change. I'll merge.

PR #1673, opened 2020-07-08
Add (currently failing) test of copy-constructing LambdaRealDoubleVisitor

bjodah:
Use after free occurs if the new visitor is used after the original visitor was destructed.

xref gh-1674

### DIFF
--- a/symengine/tests/eval/test_lambda_double.cpp
+++ b/symengine/tests/eval/test_lambda_double.cpp
@@ -145,12 +145,18 @@ TEST_CASE("Evaluate double cse", "[lambda_double_cse]")
     r = add(x, add(mul(y, z), pow(mul(y, z), integer(2))));
     s = add(mul(integer(2), x), add(mul(y, z), pow(mul(y, z), integer(2))));
 
-    LambdaRealDoubleVisitor v;
-    v.init({x, y, z}, {r, s}, true);
-
     double d[2];
     double inps[] = {1.5, 2.0, 3.0};
-    v.call(d, inps);
+    LambdaRealDoubleVisitor v2;
+    {
+        LambdaRealDoubleVisitor v;
+        v.init({x, y, z}, {r, s}, true);
+        v.call(d, inps);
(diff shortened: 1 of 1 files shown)


certik:
The test should be fixed.

bjodah:
Yes, I find it quite hard to debug though, using address sanitizer and stepping into the last relevant frame:
```
(lldb) f 6
frame #6: 0x000000000061533e test_lambda_double`SymEngine::LambdaDoubleVisitor<double>::bvisit(this=0x000060800001c240, x=0x00007fffffffb150)::'lambda0'(double const*)::operator()(double const*) const at lambda_double.h:116:44
   113 	        if (it != cse_intermediate_fns_map.end()) {
   114 	            auto index = it->second;
   115 	            result_
-> 116 	                = [=](const T *x) { return cse_intermediate_results[index]; };
   117 	            return;
   118 	        }
   119 	        throw SymEngineException("Symbol not in the symbols vector.");
(lldb) p index
(unsigned long) $1 = 0
(lldb) p cse_intermediate_results
error: <user expression 2>:1:1: use of non-static data member 'cse_intermediate_results' of 'LambdaDoubleVisitor<double>' from nested type ''
cse_intermediate_results
^~~~~~~~~~~~~~~~~~~~~~~~
```
I'm unclear on the rules and lifetime restrictions of capturing lambdas of member variables. Also I'm surprised that I'm not allowed to read the variable in the debugger. But I think this is close to where the issue is at, since a copy-constructed visitor not employing CSE did not trigger any errors when I gave it a try.

lkeegan:
@bjodah you're right this is where the issue is - the lambda is capturing the address of `cse_intermediate_results[index]`, so when `LambdaRealDoubleVisitor` is copy constructed the lambda is copied, the new lambda also points to this address.

One solution would be to ensure that this address remains valid even when the original `LambdaRealDoubleVisitor` , e.g. use a `shared_ptr` for `cse_intermediate_results`

PR #1710, opened 2020-11-16
is_positive_definite for DenseMatrix

rikardn:
Depending on PR #1707.

Implement the same algorithm as in sympy:
1. First "hermitify" (A + A.H) if non hermitian but square
2. Check if any non-positive diagonals. Shortcut to False in that case.
3. Check if diagonally dominant. Shortcut to True if True and only positive diagonals
4. If still undecided use gaussian elimination and check for positive diagonals

Most unit tests were copied from sympy.

### DIFF
--- a/symengine/dense_matrix.cpp
+++ b/symengine/dense_matrix.cpp
@@ -252,6 +252,68 @@ tribool DenseMatrix::is_strictly_diagonally_dominant() const
     return diagdom;
 }
 
+tribool DenseMatrix::shortcut_to_posdef() const
+{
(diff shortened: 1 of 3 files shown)


rikardn:
I have found a bug in this PR so please do not merge yet. I'll update as soon as I have fixed it.

rikardn:
Fixed the problem and added `is_negative_definite`

isuruf (approved):

rikardn:
Thanks @isuruf for reviewing

PR #1726, opened 2021-02-25
Building Doxygen

HaoZeke:
Builds and deploys the API, according to the new standard (in a different PR)


isuruf (commented):
Why do we need `docs/Doxygen/tags/cppreference-doxygen-web.tag.xmldocs/Doxygen/tags/cppreference-doxygen-web.tag.xml` ?

It's a large file

HaoZeke:
@isuruf it basically creates links to cppreference
image
The `std::unordered_map` is [linked to this page](https://en.cppreference.com/w/cpp/container/unordered_map).
Kinda overkill I suppose

HaoZeke:
> Why do we need `docs/Doxygen/tags/cppreference-doxygen-web.tag.xmldocs/Doxygen/tags/cppreference-doxygen-web.tag.xml` ?
> 
> It's a large file

Aaron suggested pulling it in at build time, which will work. We don't need this to be checked in and the generated docs are not any larger without it.

HaoZeke:
@isuruf I think this is ready now; it has the code toggle and also does not keep the large `.xml`
image
image

isuruf:
travis-ci job building docs fail

HaoZeke:
Travis isn't supposed to build the docs anymore, there's a gha for that. Sorry I must have only removed the deploy step. Will update. 

---
Rohit
________________________________

HaoZeke:
@isuruf this should be good to go now.

PR #1728, opened 2021-03-02
API Documentation

HaoZeke:
As per the style guide, after #1726 and #1727 

----
This PR is part of the **Google Season of Docs 2020**


HaoZeke:
@isuruf and @certik; this should be OK now.

certik (approved):
That looks good to me. Some of the code was moved around, but otherwise just new docs were added.

PR #1746, opened 2021-03-19
A unit test for  issue 1745. 

HQS-Peter:
A unit test for testing solve related to issue 1745 where solve returns  a nan.


HQS-Peter:
Strangely the clang format test didn't complain about the formatting of my PR.
I didn't check all of the 24 failing tests above, but it looks like they just fail
as expected as long as issue 1745 it is not fixed.

lkeegan (approved):
lgtm!

PR #1760, opened 2021-03-28
Better symbols for number sets in LaTeX printing

rikardn:
Fix symbols for Reals, Rationals and Integers sets in LaTeX printing. 
I.e. R now becomes 
\mathbb{R}
instead of 
\mathbf{R}


isuruf (approved):

rikardn:
Thanks for reviewing @isuruf. It is strange how codecov thinks there was a reduction in coverage since lines where only changed.

PR #1764, opened 2021-04-03
Add is_complex

rikardn:
Adding `is_complex` test. This is a bit more complete than the previous test functions. It tests for singular points for most trigonometric functions, supports assumptions and `add`, `mul` and `pow`.

### DIFF
--- a/symengine/number.h
+++ b/symengine/number.h
@@ -147,6 +147,7 @@ tribool is_nonpositive(const Basic &b);
 tribool is_negative(const Basic &b);
 tribool is_nonnegative(const Basic &b);
 tribool is_real(const Basic &b, const Assumptions *assumptions = nullptr);
+tribool is_complex(const Basic &b, const Assumptions *assumptions = nullptr);
 
 class NumberWrapper : public Number
 {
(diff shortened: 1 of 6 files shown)


symengine/number.h
```diff
@@ -147,6 +147,7 @@ tribool is_nonpositive(const Basic &b);
 tribool is_negative(const Basic &b);
 tribool is_nonnegative(const Basic &b);
 tribool is_real(const Basic &b, const Assumptions *assumptions = nullptr);
+tribool is_complex(const Basic &b, const Assumptions *assumptions = nullptr);
```
certik, line 150: Where is this method implemented? I didn't see it in this PR, but maybe I missed it. I was just curious.

certik (approved):
I think this looks good. Thanks for working on this.

symengine/number.h
rikardn: It is in `test_visitors.cpp` together with the other "test"-functions related to numbers.

certik: Ah, I see it now. Thanks!

rikardn:
Thanks  @ondrej for reviewing!

PR #1765, opened 2021-04-03
Use assumptions in matrix is_real

rikardn:
As discussed in #1759

### DIFF
--- a/symengine/dense_matrix.cpp
+++ b/symengine/dense_matrix.cpp
@@ -145,12 +145,12 @@ tribool DenseMatrix::is_diagonal() const
     return cur;
 }
 
-tribool DenseMatrix::is_real() const
+tribool DenseMatrix::is_real(const Assumptions *assumptions) const
 {
-    auto A = *this;
+    RealVisitor visitor(assumptions);
     tribool cur = tribool::tritrue;
     for (auto &e : m_) {
-        cur = and_tribool(cur, SymEngine::is_real(*e));
+        cur = and_tribool(cur, visitor.apply(*e));
         if (is_false(cur)) {
             return cur;
         }
--- a/symengine/matrix.h
+++ b/symengine/matrix.h
@@ -133,7 +133,7 @@ public:
     virtual bool is_upper() const;
     virtual tribool is_zero() const;
     virtual tribool is_diagonal() const;
-    virtual tribool is_real() const;
+    virtual tribool is_real(const Assumptions *assumptions = nullptr) const;
     virtual tribool is_symmetric() const;
     virtual tribool is_hermitian() const;
     virtual tribool is_weakly_diagonally_dominant() const;
--- a/symengine/tests/matrix/test_matrix.cpp
+++ b/symengine/tests/matrix/test_matrix.cpp
@@ -8,10 +8,13 @@
 #include <symengine/pow.h>
 #include <symengine/symengine_exception.h>
 #include <symengine/visitor.h>
+#include <symengine/assumptions.h>
 
 using SymEngine::print_stack_on_segfault;
 using SymEngine::RCP;
 using SymEngine::integer;
+using SymEngine::reals;
+using SymEngine::Assumptions;
 using SymEngine::rational;
 using SymEngine::DenseMatrix;
 using SymEngine::Basic;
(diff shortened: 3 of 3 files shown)


certik (approved):
Looks good using the current design.

A design question: why is `B` indeterminate if it is real? Do not we assume that `x` is complex (by default), which would mean that `B` is complex, and thus not real?

lkeegan (approved):
lgtm!

rikardn:
Thanks @ondrej and @lkeegan for reviewing.

If we assume that x is complex it can also be real at the same time hence B is indeterminate. For example if `x = 1` B is real but if `x = i` B is not real.

certik:
So if `x` is complex, and you do `x.is_real()`, is that indeterminate also?

rikardn:
Yes, since all real numbers are complex. I know that internally in symengine a `Complex` cannot be real because of the canonicalization (which is a very clever design!), but in mathematics the number 1 for example is both integer, rational, real and complex at the same time. I hope this makes sense!

certik:
Yes, in mathematics all real numbers are also complex. But there are a few ways that can be encoded in SymEngine, that's why I am asking. These assumptions are mainly for symbols like "x", so we can do anything we want. The current design is the most comprehensive / encodes the most information: if `is_real()` returns true, it means the symbol is definitely real (I think we always have to do this no matter what).

If it returns false, in the current design it means it is definitely not real, and if it returns indeterminate, it can be real or not, but we don't know for sure.

The only alternative design I can think if is to only return true/false, where true is as now, and false is for everything else, i.e. indeterminate and false in the current design would return false in the alternative design.

If these are the only two approaches, then the current approach with indeterminate is better, because it encodes more information, and if somebody wants the alternative design, they can simply get it with `bool is_real_alternative = (x.is_real() == true)`.

So I would say that this is well designed. 

My other unrelated question is: can you assign assumptions to expressions, such as `x+y is positive`?

rikardn:
Thanks for clarifying and sorry for misunderstanding. Yes, I also think this design is the most comprehensive and also very similar (if not the same) to what sympy does with `indeterminate` being represented with `None`. Symengine.py currently translates `indeterminate` to `None`.

As for your other question:
This is currently allowed by the interface to the `Assumptions` class. Any set of symengine statements can be given as assumptions. Your example would be `x + y > 0`. But it also has to be supported by methods on the `Assumptions` class and there is currently only support for checking simple assumption on one symbol.

Currently `Assumptions`:
1. Take set of statements as input at construction time.
2. Cache simple facts like `is_real` or `is_positive` for symbols for fast retrieval.
3. Allow queries for symbols. Like: "do we assume that x is real?" or "do we assume that x is positive?"

So the statement `x + y > 0` would currently be ignored since it cannot lead to any conclusions for `x` or `y` separately. The question is then how can we expand the interface of the `Assumptions` class to allow the information in more complex assumptions being used. Of course we could allow all statements to be reached in their original form. Another option is to let queries on the assumptions be for whole expressions, as the new sympy assumption system does. Currently I have wanted to decouple the assumptions themselves from the questions we ask using the assumptions, i.e. I have wanted the `is_real` function to make use of assumptions to answer the question rather than `is_real` being a method on the assumptions class, which would create a much tighter coupling. Such a design would also need `simplify` and `expand` to be a part of the assumptions system. Of course if this turns out to be a better design it shouldn't be a too big thing to change. So ideas for how to expand the interface to allow making use of more complex statements would be very welcome!

certik:
You are thinking about it in exactly the same way as I am. There are two lessons I learned over the years:

* Decouple all algorithms and data that can be decoupled, ahead of time, and test independently. If we later decide to couple some things for speed or simplicity, we can always do that. But it's very hard to later decouple things.
 
* Decouple data representation and the algorithms that operate on it. (This is a specific subset of the previous point, but very important.)

Two things happen when you do that: it's easier to maintain and develop, but ironically things are also usually faster, because you can concentrate on a given algorithm with clear input and output and you can also modify the data representation as needed to make the algorithm go faster. 

Is this the best possible design in all cases? I don't know, but my experience with these big systems like SymEngine or LFortran is that it delivers maintainable and testable code, as well as excellent speed, faster than other monoliths in practice. And if the speed was not enough, one can later decide to couple for speed.

Applied to the problem at hand, definitely `Assumptions` should store the assumptions, as data. But querying it is a separate problem and thus should be elsewhere, at least for now.

Finally, regarding the API for `x+y>0`, it seems `Assumptions` can already store it, so the job is done there. The question is how we query it --- I would implement the same system as in the new assumptions in SymPy, unless there is a better approach? It's been a while, but how does Mathematica do it? I would allow a similar user experience.

rikardn:
Again, thanks for reviewing, and I'll take a more closer look at the new sympy Assumptions to see if and how it would be possible to make use of more complex assumptions.

PR #1768, opened 2021-04-09
parse->print->parse rountrip tests

lkeegan:
Add parser tests for roundtrip parse -> print -> parse for each existing parser test

Fix failing tests:

- add `True` and `False` to parser constants
- add `!=` operator to tokenizer.re and parser.yy
- modify double round trip parser tests
  - expression stores double in full ~17 digit precision
  - printer prints up to 15 non-zero digits
  - so first parse & print once to get max 15 digits before testing round trip


rikardn (approved):

rikardn:
Great idea to do round trip tests!

PR #1771, opened 2021-04-11
Add primorial function

rikardn:
Given an earlier comment from @isuruf I would like to get the ntheory_funcs.h included from functions.h. I'll have to get back to that.


rikardn:
To fix the errors I need to create the `mp_primorial` function for boost and for flint, but I don't understand where to put the flint function.

lkeegan:
@rikardn would it be ok with you if I force push this branch to rebase on master?

rikardn:
@lkeegan Sure! I got stuck on how the different bignum libraries are handled. They all seem to be handled a bit differently.

rikardn:
Thanks a lot @lkeegan! I have added some unit tests to also pass the coverage test.

PR #1772, opened 2021-04-11
Expand the dependency tree of LLVM libraries.

bjodah:
When trying to link against SymEngine in a project of mine (using static libraries),
I got undefined reference for a number of LLVM related symbols.
llvm-config returned a larger number of link libraries than what
my installed version of SymEngineConfig.cmake gave. I believe this
is the culprit.


lkeegan (approved):
lgtm!

another option could be to use `explicit_map_components_to_libraries`, which expands the dependencies and also removes any unused libs:

https://github.com/llvm/llvm-project/blob/c965e14a12955355ead9ea093989a8fcbf03a8c1/llvm/cmake/modules/LLVM-Config.cmake#L347-L358

bjodah:
Interesting, if that function is preferable we should use it instead (perhaps in CMakeLists.txt too then), I lack the CMake expertise to fully understand how that predicate with `if ( TARGET ... )` interacts with symengine.

lkeegan:
@bjodah looking at that function I don't think that it interacts with symengine at all, I think it is just checking that each llvm library exists before adding it to `out_libs`. I haven't found any documentation for it though.

lkeegan:
@bjodah I would suggest that if this change resolves your issue we just merge this as is: as far as I can see it doesn't change anything for any of the CI builds that use llvm or for me locally

bjodah:
Thanks. Yes, we can improve upon this later as well. (maybe add a test case using static linking to the CI config)

PR #1782, opened 2021-04-28
Add unicode printer

rikardn:
A unicode printer.

### DIFF
--- a/symengine/printers/stringbox.cpp
+++ b/symengine/printers/stringbox.cpp
@@ -1,0 +1,269 @@
+#include <iterator>
+#include <sstream>
+#include <symengine/printers/stringbox.h>
+
+namespace SymEngine
+{
+
+void StringBox::pad_lines(std::size_t new_width)
+{
+    auto diff = new_width - width_;
+    auto half = diff / 2;
+    auto odd = diff % 2;
+    for (std::string &line : lines_) {
+        line.insert(0, std::string(half + odd, ' '));
+        if (half > 0) {
+            line.append(std::string(half, ' '));
+        }
+    }
+}
+
+void StringBox::add_below(StringBox &other)
+{
+    if (other.width_ > width_) {
+        pad_lines(other.width_);
+        width_ = other.width_;
+    } else if (width_ > other.width_) {
+        other.pad_lines(width_);
+        other.width_ = width_;
+    }
+    lines_.insert(lines_.end(), other.lines_.begin(), other.lines_.end());
+}
+
+void StringBox::add_below_unicode_line(StringBox &other)
+{
+    auto new_width = std::max(width_, other.width_);
+    std::string bar;
+    for (unsigned i = 0; i < new_width; i++) {
+        bar.append("\u2015");
+    }
+    StringBox barbox(bar, new_width);
+    add_below(barbox);
+    add_below(other);
+}
+
+void StringBox::add_right(StringBox &other)
+{
+    StringBox *smaller;
+    auto this_size = lines_.size();
+    auto other_size = other.lines_.size();
+    if (other_size > this_size) {
+        smaller = this;
+    } else {
+        smaller = &other;
+    }
+    auto diff
+        = std::max(other_size, this_size) - std::min(other_size, this_size);
+    auto half = diff / 2;
+    auto odd = diff % 2;
+    std::string pad(smaller->width_, ' ');
+    for (unsigned i = 0; i < half; i++) {
+        smaller->lines_.push_back(pad);
+        smaller->lines_.insert(smaller->lines_.begin(), pad);
+    }
+    if (odd == 1) {
+        smaller->lines_.insert(lines_.begin(), pad);
+    }
+    for (unsigned i = 0; i < lines_.size(); i++) {
+        lines_[i].append(other.lines_[i]);
+    }
+    width_ += other.width_;
+}
+
+void StringBox::add_power(StringBox &other)
+{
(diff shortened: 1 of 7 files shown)


rikardn:
Ok, support for the most important classes is in. To keep the size of the PR down I would say this is ready for review. We can extend this in the future.

symengine/printers/unicode.cpp
```diff
@@ -0,0 +1,760 @@
...
+    return box_;
+}
+
+StringBox UnicodePrinter::apply(const Basic &b)
+{
+    b.accept(*this);
+    return box_;
+}
+
+std::vector<std::string> init_unicode_printer_names()
```
lkeegan, line 630: maybe start with the default values from [init_str_printer_names()](https://github.com/symengine/symengine/blob/master/symengine/printers/strprinter.cpp#L1038) then set the ones that differ? reduces duplication & you also support new functions automatically when they are added to the main printer

symengine/printers/unicode.cpp
```diff
@@ -0,0 +1,760 @@
...
+    return names;
+}
+
+const std::vector<std::string> UnicodePrinter::names_
+    = init_unicode_printer_names();
+
+std::vector<unsigned> init_unicode_printer_lengths()
+{
+    std::vector<unsigned> lengths;
+    lengths.assign(TypeID_Count, 0);
```
lkeegan, line 690: is `0` a good default value here? (e.g. if a new function is added at some point & this is not updated, maybe a larger value would be more likely to look reasonable?)

symengine/printers/unicode.h
```diff
@@ -0,0 +1,81 @@
...
+#define SYMENGINE_UNICODE_H
+
+#include <symengine/printers/strprinter.h>
+#include <symengine/printers/stringbox.h>
+
+namespace SymEngine
+{
+
+std::string print_double(double d);
+std::vector<std::string> init_str_printer_names();
```
lkeegan, line 11:
```suggestion
```

symengine/printers/unicode.cpp
lkeegan, line 630:
```suggestion
static std::vector<std::string> init_unicode_printer_names()
```

symengine/printers/stringbox.h
```diff
@@ -0,0 +1,53 @@
...
+class StringBox
+{
+private:
+    std::vector<std::string> lines_;
+    unsigned long width_;
+
+    void pad_lines(unsigned long new_width);
+
+public:
+    StringBox(std::string s)
```
lkeegan, line 16:
```suggestion
    explicit StringBox(std::string s)
```

symengine/printers/unicode.h
```diff
@@ -0,0 +1,81 @@
...
+std::vector<unsigned> init_unicode_printer_lengths();
```
lkeegan, line 12: does this need to be public, or can it be made a static function in unicode.cpp?

lkeegan (commented):
@rikardn looks good to me! I didn't go through everything in detail, but added some minor suggestions/questions.

also, would it make sense to use `std::size_t` throughout instead of a mix of `unsigned` and `unsigned long`? (since it seems these numbers are typically either originally coming from calling `.size()` or being passed to a string constructor that takes a `std::size_t`)

symengine/printers/stringbox.cpp
```diff
@@ -0,0 +1,268 @@
+#include <sstream>
```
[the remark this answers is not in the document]
rikardn: Ok, I'll add this to the same commit.

symengine/printers/stringbox.h
rikardn: :+1:

symengine/tests/printing/test_printing.cpp
```diff
@@ -731,3 +743,330 @@ TEST_CASE("test_latex_printing()", "[latex]")
...
+
+TEST_CASE("test_unicode()", "[unicode]")
+{
+    RCP<const Basic> x = symbol("x");
+    RCP<const Basic> y = symbol("y");
+    std::string s;
+
+    s = unicode(*reals());
+    CHECK(s == "\u211D");
+    std::cout << "Reals:" << s << std::endl;
```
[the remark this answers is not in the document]
rikardn: To keep it simple I'll just remove the printouts.

symengine/printers/unicode.cpp
rikardn: Good idea! I'll change this.

symengine/printers/unicode.cpp
rikardn: Nice catch! Given the suggestion to default to the str printer names I'll initialize the lengths with the `string.length`. Since these strings will never be multi-byte unicode it'll work as a default.

symengine/printers/unicode.h
rikardn: Good catch!

symengine/printers/unicode.cpp
rikardn: Another good catch!

symengine/printers/unicode.h
rikardn: It shouldn't be public. I'll remove the header declaration.

rikardn:
Thanks @lkeegan for your thorough review. I have done all changes you suggested now.

certik (approved):
Looks good to me.

lkeegan (approved):
lgtm!

PR #1796, opened 2021-05-16
Mobius and mertens as symbolic functions

rikardn:
Convert the mobius and mertens functions to become symbolic functions.

Calculations of numeric values can be optimized. In the future I think we should separate the numeric calculations. Currently symbolics and numerics are entangled in `ntheory.cpp`


isuruf (approved):

isuruf (commented):
Looks like tests fail

PR #1799, opened 2021-05-25
Refine

rikardn:
As promised: A refine function that simplifies expressions given assumptions (like refine in mathematica https://reference.wolfram.com/language/ref/Refine.html and sympy https://docs.sympy.org/latest/modules/assumptions/refine.html). A modest start supporting `Abs`, `Sign`, `Floor`, `Ceiling` and `Conjugate`.


rikardn:
Bad rebase. Reopening

PR #1800, opened 2021-05-25
Add refine

rikardn:
As promised: A refine function that simplifies expressions given assumptions (like refine in mathematica https://reference.wolfram.com/language/ref/Refine.html and sympy https://docs.sympy.org/latest/modules/assumptions/refine.html). A modest start supporting `Abs`, `Sign`, `Floor`, `Ceiling` and `Conjugate`.

### DIFF
--- a/symengine/refine.cpp
+++ b/symengine/refine.cpp
@@ -26,8 +26,8 @@ void RefineVisitor::bvisit(const Sign &x)
         result_ = integer(1);
     } else if (is_true(is_negative(*newarg, assumptions_))) {
         result_ = integer(-1);
-        //    } else if (is_true(is_zero(*newarg, assumptions_))) {
-        //        result_ = integer(0);
+    } else if (is_true(is_zero(*newarg, assumptions_))) {
+        result_ = integer(0);
     } else {
         result_ = sign(newarg);
     }
--- a/symengine/refine.cpp
+++ b/symengine/refine.cpp
@@ -1,0 +1,80 @@
+#include <symengine/refine.h>
+
+namespace SymEngine
+{
+
+void RefineVisitor::bvisit(const Abs &x)
+{
+    auto farg = x.get_arg();
+    auto newarg = apply(farg);
+    if (is_true(is_nonnegative(*newarg, assumptions_))) {
+        result_ = newarg;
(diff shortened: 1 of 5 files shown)


symengine/refine.cpp
```diff
@@ -0,0 +1,86 @@
...
+    auto farg = x.get_arg();
+    auto newarg = apply(farg);
+    if (is_true(is_real(*newarg, assumptions_))) {
+        result_ = newarg;
+    } else {
+        result_ = conjugate(newarg);
+    }
+}
+
+void RefineVisitor::bvisit(const UnevaluatedExpr &x)
```
lkeegan, line 73: why does refine evaluate an `UnevaluatedExpr`, is that something that sympy does?

lkeegan (approved):
lgtm!

symengine/refine.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+void RefineVisitor::bvisit(const Sign &x)
+{
+    auto farg = x.get_arg();
+    auto newarg = apply(farg);
+    if (is_true(is_positive(*newarg, assumptions_))) {
+        result_ = integer(1);
+    } else if (is_true(is_negative(*newarg, assumptions_))) {
+        result_ = integer(-1);
+        //    } else if (is_true(is_zero(*newarg, assumptions_))) {
```
lkeegan, line 29: maybe remove these commented-out lines?

symengine/refine.cpp
rikardn: Sympy does it in `simplify` but not in `refine`. I didn't think further than adding it because it was trivial to add. Do you think I should remove it?

symengine/refine.cpp
rikardn: Sorry about these. I didn't have the `is_zero` with assumptions when implementing this. I should uncomment and add some unit tests for this case.

symengine/refine.cpp
lkeegan: I don't know - but I would assume that having the same semantics as sympy where possible would make sense? Maybe this could be factored out into a `doit` visitor?

rikardn: Yes, that makes sense.

rikardn:
Thanks @lkeegan for reviewing!

PR #1816, opened 2021-08-15
Support Add by is_positive

rikardn:
Basic check for all terms positive or all terms negative in `Add` for `is_positive`

### DIFF
--- a/symengine/test_visitors.cpp
+++ b/symengine/test_visitors.cpp
@@ -141,6 +141,51 @@ void PositiveVisitor::bvisit(const Number &x)
     }
 }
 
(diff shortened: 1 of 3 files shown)


lkeegan (approved):
lgtm!

a possible simplification could be to rewrite the if/else if/else block as
```
if(can_be_false and ...){
    can_be_false = false;
}
if(can_be_true and ...){
    can_be_true = false;
}
```
but also looks fine to me as it is

rikardn:
Thanks @lkeegan for reviewing. I actually tried what you propose because it should be faster. Turned out to much harder to get the `else` clause correct.

PR #1817, opened 2021-08-15
Add is_finite and is_infinite

rikardn:
So far the basics is in. We could also consider using the `is_complex` as the same as `is_finite`.

Would be nice to at least have all the `is_` functions that sympy has to be more compatible


lkeegan (approved):
lgtm!

rikardn:
Thanks @lkeegan for the review

PR #1818, opened 2021-08-15
Add is_odd and is_even

rikardn:


lkeegan (approved):
lgtm - a general question about these `is_x` functions: what is the distinction between returning `tribool::indeterminate` vs throwing an exception?
e.g. comparing `is_finite` with `is_even`, it looks like for many inputs the first one would throw while the second one would return `tribool::indeterminate`

rikardn:
Good question @lkeegan. My thinking is that all of these functions assume numeric expressions as input. If encountering something that is not a number, e.g. `Set` or `Logic` this would be like a `TypeError` of the input. Previously I returned `tribool::trifalse` in these cases, but we decided to change that to throw instead. `tribool::indeterminate` is used for cases where we either know that the input is numeric, but we cannot decide the truth value of the test or if we cannot even know if it would be numeric or not. Another option would be to introduce a new subclass `Numeric` or similar where all `Add`, `Mul`, `Number` etc would fall under to distinguish it from `Sets`, which has its own "base" type. This way the `is_*` functions could only allow numeric input at compile time. The downsides would be that it would be a big change and that `Mul` etc cannot be reused for other objects like vectors for example.

Sympy returns `None` (i.e. indeterminate) in these cases which I don't find satisfying.

To summarize our options:

1. throw for clear domain errors or "`TypeError`"
2. return tribool::trifalse in these cases.
3. Change class hierarchy and remove the need for throwing.

rikardn:
There are some functions that haven't been changed from the old "no-throw" system. Perhaps these are confusing. I'll continue to throw for now and we can revisit in the future.

PR #1820, opened 2021-08-15
Remove failing test to workaround #1812

rikardn:
Commented out the failing test to be able to get the CI to pass. It is the `WITH_SANITIZE="undefined"` case.


rikardn:
No longer needed because of #1822

PR #1751, opened 2021-03-21
fix naming of llvm intrinsics

lkeegan:
- rename `llvm.pow.f64.f64` -> `llvm.pow.f64`
- similarly for `pow.*`, `powi.*`, `maxnum.*`, `minnum.*`

PR #1823, opened 2021-08-17
only specify integer type of llvm powi 2nd argument for llvm version > 12

lkeegan:
fix for these warnings when running `./symengine/tests/eval/test_lambda_double` with llvm <= 12:

```
Intrinsic name not mangled correctly for type arguments! Should be: llvm.powi.f32
float (float, i32)* @llvm.powi.f32.i32﻿
```

(#1810 re-introduced this warning which was previously fixed in #1751)

### DIFF
--- a/symengine/llvm_double.cpp
+++ b/symengine/llvm_double.cpp
@@ -495,7 +495,9 @@ llvm::Function *LLVMVisitor::get_powi()
 {
     std::vector<llvm::Type *> arg_type;
     arg_type.push_back(get_float_type(&mod->getContext()));
+#if (LLVM_VERSION_MAJOR > 12)
     arg_type.push_back(llvm::Type::getInt32Ty(mod->getContext()));
+#endif
     return llvm::Intrinsic::getDeclaration(mod, llvm::Intrinsic::powi,
                                            arg_type);
 }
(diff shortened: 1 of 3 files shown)


symengine/llvm_double.cpp
```diff
@@ -495,7 +495,9 @@ llvm::Function *LLVMVisitor::get_powi()
 {
     std::vector<llvm::Type *> arg_type;
     arg_type.push_back(get_float_type(&mod->getContext()));
+#if (LLVM_VERSION_MAJOR > 12)
```
rikardn, line 498: I am not sure I understand this. So for older versions of LLVM only one `arg_type` will be used, but two for versions 13 and forward.

lkeegan: that's right - for older LLVM the second arg is implicitly `int32`, and setting it explicitly appears to be harmless but gives the above warning - while for recent LLVM it needs to be set explicitly otherwise it will segfault (see #1810 for more details)

rikardn: Ok, understood

rikardn (approved):
I approve. I think the codecov issue can be ignored. At least I cannot see what lines codecov believes should be tested.

lkeegan:
@rikardn thanks for the review  - I think codecov is just complaining that overall coverage percentage has gone down (since the #if effectively removed one line that was previously covered in the coverage CI jobs)

PR #1825, opened 2021-08-18
Fix cotire missing flags

lkeegan:
- move cotire to end of CMakeLists.txt after all target properties (e.g. c++14) have been set
- re-enable cotire in ci jobs where it was disabled due to using c++ std > 11
- left it disabled for msan build as it causes other problems there


rikardn (approved):
Looks good.

I read that the functionality in cotire is built into cmake since a while back. Perhaps in the future we would want to use this functionality instead.

PR #1826, opened 2021-08-18
CI improvements

lkeegan:
- bump versions
  - ubuntu -> 20.04
  - llvm -> 12 / 13
  - gcc -> 9 / 10
  - flint-dev -> latest github release
- fixes
  - disable cotire for asan
  - fix teuchos compiler warning about throwing from a destructor
    - copied with minor changes from https://github.com/trilinos/Trilinos/commit/9c3771bd6d74c724c281ae86b33f37bffa78db0b
- speedup
  - add ccache, with unique cache id for each job based on environment variables
  - use -j2 for make everywhere by default
- cleanup
  - remove unused environment variables
  - move coverage ci jobs back into main ci yml
  - remove binutils compilation
  - remove llvm apt install


.github/workflows/coverage1.yml
```diff
@@ -1,27 +0,0 @@
...
-        WITH_BFD: yes
-        WITH_LATEST_GCC: yes
-        WITH_COVERAGE: yes
-        WITH_MPC: yes
-        WITH_LLVM: 8.0
-        USE_GLIBCXX_DEBUG: yes
-        OS: ubuntu-16.04
-        CC: gcc
-        EXTRA_APT_REPOSITORY: 'deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main'
-        EXTRA_APT_PACKAGES: "clang-8 llvm-8-dev binutils-dev g++-8"
```
rikardn, line 27: Has the reason for this being split out to a separate workflow been resolved?

lkeegan:
two comments:
- if cotire is disabled then ccache works better & the builds are faster
  - but as cotire is enabled by default it should be tested in the CI
  - as @rikardn suggested since cotire is no longer maintained maybe using cmake pch instead (for cmake versions >= 3.16) would make sense
- the coverage has changed
  - probably because of the change in gcc/gcov versions
  - from a quick look at the changes, mostly they seem to be improvements
  - e.g. line 449 of mul.cpp
    - was covered in master: https://codecov.io/gh/symengine/symengine/src/master/symengine/mul.cpp#L449
      - although preceeding line is not covered, so this doesn't make much sense
    - no longer covered in this PR: https://codecov.io/gh/symengine/symengine/src/18b1f20dd6f58c956009316c4bb1afb917537e8a/symengine/mul.cpp#L449

.github/workflows/coverage1.yml
lkeegan: yes as far as I know this has been fixed on the codecov side

rikardn (approved):
Nice work! I think we'll have to let the code coverage reduction slip given that it might have been calculated in a worse way before.

Another future improvement is to get rid of the AppVeyor tests. I remember that they were kept because of some issues, but not exactly what the problem was.

lkeegan:
thanks @rikardn for reviewing!

PR #1831, opened 2021-08-28
First version of simplify

rikardn:
Will first do `refine` and then try more advanced simplifications. Started out with supporting `y/csc(x) = y*sin(x)` and similar.

I encountered the issue of having to support both `Pow` and `Mul` for very similar situations. For example `1 / x` will be a `Pow` but `2 / x` will be a `Mul`. Perhaps this would break too much and be a bad idea for other reasons, but in this scenario it would simplify things if also `1 / x`  would be `Mul` with coeff = 1. Then `Mul` would cover all cases with numeric exponents and Pow wouldn't have to cater for these.

### DIFF
--- a/symengine/CMakeLists.txt
+++ b/symengine/CMakeLists.txt
@@ -100,6 +100,7 @@ set(SRC
     test_visitors.cpp
     assumptions.cpp
     refine.cpp
+    simplify.cpp
 )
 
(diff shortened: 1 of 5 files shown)


rikardn:
Any comments or are we ready to merge?

yumingdong:
Why not implement all the functions similar to Sympy's simplify() function? In fact, algebraic simplification plays a crucial role in mathematical computations and analyses.

PR #1844, opened 2021-09-15
sbml parser: add support for zero arg function symbols

lkeegan:
- parse unknown zero-argument functions as function symbols instead of raising exception
- this is now consistent with what is done for functions with arguments


rikardn (approved):

lkeegan:
thanks @rikardn!

PR #1849, opened 2021-09-28
Fixes for "static initialization order fiasco"

cqc-alec:
When symengine is statically linked to a library, especially with link-time optimization, it can result in segfaults due to this issue: https://en.cppreference.com/w/cpp/language/siof

This PR tries to fix the instances I have found. It's possible I've missed some, but at least this is enough to avoid the issue in my use case involving pybind11 (527 valgrind errors from importing the python module reduced to 0).

Typically the fix is some form of the initialize-on-first-use idiom. For the constants, I introduced lambdas for this purpose to avoid having to add `()` to every reference to them in the code base.

### DIFF
--- a/symengine/constants.cpp
+++ b/symengine/constants.cpp
@@ -90,41 +90,4 @@ RCP<const Basic> mC4 = mul(minus_one, C4);
 RCP<const Basic> mC5 = mul(minus_one, C5);
 RCP<const Basic> mC6 = mul(minus_one, C6);
 
-// sin_table[n] represents the value of sin(pi*n/12) for n = 0..23
-RCP<const Basic> sin_table[]
-    = {zero, C0,  C1,  C2,  C3,  C4,  one,       C4,  C3,  C2,  C1,  C0,
-       zero, mC0, mC1, mC2, mC3, mC4, minus_one, mC4, mC3, mC2, mC1, mC0};
-
-umap_basic_basic inverse_cst = {
-    {C3, i3},
-    {mC3, im3},
-    {C2, mul(i2, i2)},
-    {mC2, mul(im2, i2)},
-    {C4, integer(12)},
-    {mC4, integer(-12)},
-    {C5, i5},
-    {mC5, im5},
-    {C6, integer(10)},
-    {mC6, integer(-10)},
-    {div(one, i2), integer(6)},
-    {div(minus_one, i2), integer(-6)},
-};
-
-umap_basic_basic inverse_tct = {
-    {div(one, sq3), mul(i2, i3)},
-    {div(minus_one, sq3), mul(im2, i3)},
-    {sq3, i3},
-    {mul(minus_one, sq3), im3},
-    {add(one, sq2), div(pow(i2, i3), i3)},
-    {mul(minus_one, add(one, sq2)), div(pow(i2, i3), im3)},
-    {sub(sq2, one), pow(i2, i3)},
-    {sub(one, sq2), pow(im2, i3)},
-    {sub(i2, sq3), mul(mul(i2, i2), i3)},
-    {sub(sq3, i2), mul(mul(im2, i2), i3)},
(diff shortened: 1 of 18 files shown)


cqc-alec:
Coverage test failed but this report seems OK: https://app.codecov.io/gh/symengine/symengine/compare/master...fiasco/changes
Can I ignore the failure or should I try to add some test code to make it green?

rikardn:
Interesting! I am trying to understand some things:

1. Will this affect performance? Will the lambda be called each time a constant is used?
2. Why did this only affect static compilation?
3. Could it be solved in another way. I have seen proposals to use `static` in functions for constants. Could some situations use `constexpr`?

lkeegan:
thanks @cqc-alec !

re performance, I found `./benchmarks/parsing_google` was ~2x slower with this pr, but this was just due to copies being made of the static maps, with that fixed (https://github.com/symengine/symengine/pull/1849/commits/f85013974c2dace02d52ffc6e3063a89009c4a66)  the parsing benchmarks are not significantly different to the master branch.

I also looked at a few other benchmarks (but not all), only `/benchmarks/eval_double1` was noticeably slower (~3.8ms for this pr vs ~3.6ms for master)

cqc-alec:
Thanks @lkeegan , yes passing the maps by reference is a good idea. I wonder why `eval_double` got slower.
@rikardn 1. The lambda won't be called each time, just once. 2. Actually I think it would be an issue with dynamic linkage too. 3. Some cases could perhaps be solved with `constexpr` but only for literal types, and many of these are vectors or maps so not sure it could work (at least with pre-C++20).

cqc-alec:
I tried making `init_eval_double` inline -- perhaps this helps a bit with performance. (I see some variation when running the benchmark making it hard to be sure if it has really made a difference, but judging from a few runs on my laptop it is usually < 5ms after compared with usually > 5ms before.)

cqc-alec:
The test failure looks like a transient network issue.

lkeegan:
Here are the relative changes in timings for the `benchmarks/bench_eval_double` benchmark for this pr vs master (negative means faster than master, positive means slower). Looking at these I would say the performance looks fine, none are significantly different from zero. (using [compare.py](https://github.com/google/benchmark/blob/main/docs/tools.md) U test with 20 repetitions)

```
eval_double/2_mean                                      -0.0128
eval_double/8_mean                                      -0.0048
eval_double/64_mean                                     +0.0006
eval_double/512_mean                                    +0.0069
eval_double/4096_mean                                   +0.0031
eval_double/16384_mean                                  +0.0151
eval_double_visitor_pattern/2_mean                      -0.0113
eval_double_visitor_pattern/8_mean                      -0.0200
eval_double_visitor_pattern/64_mean                     -0.0222
eval_double_visitor_pattern/512_mean                    -0.0085
eval_double_visitor_pattern/4096_mean                   -0.0190
eval_double_visitor_pattern/16384_mean                  -0.0409
eval_double_single_dispatch/2_mean                      +0.0065
eval_double_single_dispatch/8_mean                      +0.0037
eval_double_single_dispatch/64_mean                     +0.0001
eval_double_single_dispatch/512_mean                    -0.0015
eval_double_single_dispatch/4096_mean                   -0.0203
eval_double_single_dispatch/16384_mean                  -0.0249
```

rikardn:
Thanks @cqc-alec for the explanation. I have been giving this some thought and I believe your solution is the best we can get for now.

Ideally we would like to have static construction for the constants (i.e. `constexpr`), but it is not straight forward to make the `RCP` be statically constructed. I did some experimentation with smart pointers and I encountered various problems. I tested a nop destructor, but the compiler still does not want to allow `constexpr`  (`‘std::unique_ptr<int, nop>’ has a non-trivial destructor`). Also having a non-standard destructor makes it into a different type than smart pointers with the same destructor so overloaded assignment operators would be needed. To sum up: it could be possible, but significant work would be needed.

rikardn (approved):

isuruf (approved):

PR #1861, opened 2021-10-29
Add logarithm power rule to refine

rikardn:
log(x**y) -> x*log(y) if x > 0 and y real.


certik (approved):
I think that looks great. 

In some sense, the refine is like a compiler optimization pass. If it knows how to optimize, it will do it, otherwise it will ignore it.

rikardn:
Thanks @certik for reviewing. I like the analogy

PR #1876, opened 2022-01-25
Try restoring -O3 optimization for clang builds with boostmp.

cqc-alec:


cqc-alec:
I am not sure if this configuration is actually tested on the CI. I have tried it locally, with clang 12 on ubuntu 20.04 (and boost 1.71.0), and it builds fine.

PR #1880, opened 2022-02-04
Product of n terms in cwrapper

PerformanceCoder:
There is a [https://github.com/symengine/symengine/pull/1829/](vector-based add) but no vector-based product


rikardn (approved):

rikardn:
Thanks for the PR!

PR #1882, opened 2022-02-08
Support of Piecewise in parser

bjodah:
~~This is not working yet. Any feedback on approach/what's missing is of course much appreciated.~~

OK, so tests pass locally. I had to add a token for `Piecewise` to special case the parsing of list-of-pairs. Let me know what you think.


symengine/tests/basic/test_parser.cpp
```diff
@@ -602,6 +604,34 @@ TEST_CASE("Parsing: functions", "[parser]")
...
+        auto cond2 = logical_and({Gt(x, integer(2)), Le(x, integer(5))});
+        auto p = piecewise({{x, cond1},
+                            {y, cond2},
+                            {add(x, y), boolTrue}});
+
+        REQUIRE(eq(*res, *p));
+    }
+    if (1) {
+        s = "Piecewise((x, Contains(x, (-oo, 2])), (y, Contains(x, (2, "
+            "5])), (x + y, Contains(x, (5, 10])))";
```
bjodah, line 623: I think I'll leave `Contains` for a later PR, it fails as is right now.

bjodah (commented):
This is now accepted by the parser:
https://github.com/symengine/symengine/pull/1882/files#diff-ab3976f9753d1b564643de57629491911add9c383c119f7c29705665366a3cdeR928
but the dynamic cast too Boolean fails, I guess the parser is going bottom up and has a partial match on this argument since I added "epair" to the parser. Not sure how to constrain the parser. Add a designated token for Piecewise in the tokenizer?

rikardn (approved):
I think this is looking good

PR #1888, opened 2022-02-21
Support NaN in LLVM- and Lambda visitors

bjodah:
Not sure if we should prefer constructing signaling or quiet NaNs.

If someone has an application where the payload is used, I guess we could support that too at some later point.


rikardn (approved):
lgtm

bjodah:
Thank you for reviews!

PR #1906, opened 2022-04-27
Add new line at end of files

Marlin-Na:
Got NOTEs for the R package checking:
```
* checking line endings in C/C++/Fortran sources/headers ... NOTE
Found the following sources/headers not terminated with a newline:
  src/upstream/cmake/checkgmpxx.cpp
  src/upstream/symengine/polys/basic_conversions.cpp
  src/upstream/symengine/utilities/teuchos/Teuchos_TestForException.cpp
Some compilers warn on such files.
```


rikardn:
Good catch!

PR #1917, opened 2022-07-09
Add is_real to CSRMatrix

rikardn:
And create a virtual method in `Matrix` base class.


bjodah:
Looks good! How ambitious are we with regards to test coverage? A small test wouldn't hurt?

PR #1919, opened 2022-07-14
Use override for overriding member functions in matrix.h

rikardn:
Make code clearer and avoid

```
overrides a member function but is not marked
'override' [-Winconsistent-missing-override]
```

 in clang++

### DIFF
--- a/symengine/matrix.h
+++ b/symengine/matrix.h
@@ -118,15 +118,15 @@ public:
     // and throw an exception if a method is not applicable.
 
     // Get and set elements
-    virtual RCP<const Basic> get(unsigned i, unsigned j) const;
-    virtual void set(unsigned i, unsigned j, const RCP<const Basic> &e);
+    RCP<const Basic> get(unsigned i, unsigned j) const override;
+    void set(unsigned i, unsigned j, const RCP<const Basic> &e) override;
     virtual vec_basic as_vec_basic() const;
 
-    virtual unsigned nrows() const
+    unsigned nrows() const override
     {
         return row_;
     }
-    virtual unsigned ncols() const
+    unsigned ncols() const override
     {
         return col_;
     }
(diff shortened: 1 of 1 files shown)


symengine/matrix.h
```diff
@@ -118,90 +118,88 @@ class DenseMatrix : public MatrixBase
     // and throw an exception if a method is not applicable.
 
     // Get and set elements
-    virtual RCP<const Basic> get(unsigned i, unsigned j) const;
-    virtual void set(unsigned i, unsigned j, const RCP<const Basic> &e);
-    virtual vec_basic as_vec_basic() const;
+    RCP<const Basic> get(unsigned i, unsigned j) const override;
+    void set(unsigned i, unsigned j, const RCP<const Basic> &e) override;
+    vec_basic as_vec_basic() const;
```
bjodah, line 123: does this change `as_vec_basic()` to be a non-virtual method? Could this possibly break existing user code which relies on overriding this method? (I have to admit that I can't come up with an example where one would want to do so)

symengine/matrix.h
```diff
@@ -118,90 +118,88 @@ class DenseMatrix : public MatrixBase
...
+    bool is_upper() const;
+    tribool is_zero() const;
+    tribool is_diagonal() const;
+    tribool is_real(const Assumptions *assumptions = nullptr) const;
+    tribool is_symmetric() const;
+    tribool is_hermitian() const;
+    tribool is_weakly_diagonally_dominant() const;
+    tribool is_strictly_diagonally_dominant() const;
+    tribool is_positive_definite() const;
+    tribool is_negative_definite() const;
```
bjodah, line 145: Same question applies for these methods.

bjodah (commented):
Just a small question regarding virtual/non-virtual methods.

symengine/matrix.h
rikardn: You are right. This might break user code. We should keep this as virtual since it doesn't exist in the base class. Good catch!

symengine/matrix.h
rikardn: Same as above.

As a side note: I now think these would be better implemented using double dispatch as in #1918. This way we decouple the classes from the functions on the classes. Couldn't be easily done before since `Matrix` does not inherit from `Basic` which the proposed `ImmutableDenseMatrix` in the PR does.

PR #1925, opened 2022-08-08
Tribool refactor

rikardn:
* Change `tribool` from plain `enum` to safer `enum class`
* Move `tribool` to its own header `tribool.h`
* Add `or_tribool` and `orwk_tribool` for completeness
* Add tests for `tribool_from_bool`
* Fix bug that was found when moving to `enum class` for tribool


rikardn:
Thanks!

PR #1931, opened 2022-09-22
Add Integral type

AlexanderLasar:
The Integral type is not implemented yet. Expressions where symbolic integrals appear are not supported.

An integrate() method is very complicated and out of the question here, but support for expressions where symbolic integrals appear and support for differentiating them is an important missing feature!

With best regards,
Alex


AlexanderLasar:
Hello rikardn,

thank You for the fast response -- I did not have a look at that PR until just now. 
The code in the PR looks crisp and I hope it performs well when it is being reviewed.

Once the code gets reviewed an accepted, it is not gonna be on the stable branch but the master branch.
I have to build master from source to construct integral expressions and take derivatives on them, right?

Best regards ~

certik:
@erny123 please go ahead and submit a PR.

erny123:
Also, there should be an implementation of unevaluated derivatives and unevaluated integrals. 

I know there's the `Derivative` class but there's no function for evaluating an unevaluated derivative.

PR #1951, opened 2023-02-24
recursively check for missing headers in install in CI

lkeegan:
- add missing headers to cmake install
- see #1950

### DIFF
--- a/bin/test_make_install.py
+++ b/bin/test_make_install.py
@@ -1,5 +1,5 @@
 from __future__ import print_function
-from os.path import join, relpath
+from os.path import join, relpath, sep
 import sys
 from glob import glob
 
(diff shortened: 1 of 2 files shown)


rikardn:
So the idea is that the tests should currently fail with this PR right?

lkeegan:
@rikardn I also added the missing headers to CMakeLists

rikardn:
Ok. Will we also get an error in the other direction? I mean if a header file has been removed but is still in the CMakeLists?

lkeegan:
Yes the `make install` command will fail if you include a file in CMakeLists that doesn't exist, e.g.
```
CMake Error at symengine/cmake_install.cmake:394 (file):
  file INSTALL cannot find "/home/lkeegan/SSC/symengine/symengine/refineQ.h":                                                                                                                                                                       
  No such file or directory.                                                                                                                                                                                                                        
Call Stack (most recent call first):                                                                                                                                                                                                                
  cmake_install.cmake:73 (include) 
```

rikardn (approved):

lkeegan:
@rikardn thanks for reviewing!

PR #1954, opened 2023-03-14
add support for LLVM 16

lkeegan:
- requires C++17
- `getBasicBlockList()` is now private but they provide an `insert()` method
- requires CMake policy `CMP0057` (IN_LIST if() operator)


isuruf (approved):

bjodah:
I believe boost will soon also require C++17 (can't remember if it was 1.82 or 1.83 and exactly what modules will be affected), maybe we should switch the default even? (but keep some test configurations duing CI which explicitly uses -std=c++14)

rikardn:
I can vote for switching the default to C++17. (I'd even vote for C++20 because modules are awesome)

PR #1977, opened 2023-09-25
Fix deprecated function declarations

thilinarmtb:


certik:
@thilinarmtb what warning does this show? I thought the `void` argument is the deprecated way.

thilinarmtb:
@certik : Thanks for approving the changes. I got the following when I included `symengine`
in a project I compile with `-Wall -Wextra -Wpedantic`. I can reproduce it with the following
simple file:
```bash
thili@M2-Pro Misc % cat deprecated.c 
void foo() {}

int main(int argc, char **argv) {
  foo();
  return 0;
}
thili@M2-Pro Misc % gcc deprecated.c -Wall -Wextra -Wpedantic
deprecated.c:1:9: warning: a function declaration without a prototype is deprecated in all versions of C [-Wstrict-prototypes]
void foo() {}
        ^
         void
deprecated.c:3:14: warning: unused parameter 'argc' [-Wunused-parameter]
int main(int argc, char **argv) {
             ^
deprecated.c:3:27: warning: unused parameter 'argv' [-Wunused-parameter]
int main(int argc, char **argv) {
                          ^
3 warnings generated.
```

thilinarmtb:
In C, I think use of `foo(void)` in the parameter list implies empty parameter list
whereas `foo()` declares a function that takes an unspecified number of fixed
parameters.

certik:
Thanks! I see, this only happens in the C wrappers. Thanks for the fix.

PR #1981, opened 2023-09-30
Exposed some query methods

anutosh491:


isuruf (commented):
Why do we need these methods and just 5 of them?
It's a simple one liner `basic_get_type(s) == SYMENGINE_ADD`.

certik:
I agree, I think `basic_get_type(s) == SYMENGINE_ADD` seems good enough to me.

anutosh491:
I think so here we would only need the ` basic_is_Exp` function because I don't we have something like 
`basic_get_type(s) == SYMENGINE_EXP`
The type of `s` would still be `Pow`, we would just need to check it's base against E

certik:
Why couldn't we do `basic_get_type(s) == SYMENGINE_EXP` in LPython?

anutosh491:
Because I don't think we have a defined `SYMENGINE_EXP` class type registered here (https://github.com/symengine/symengine/blob/master/symengine/type_codes.inc)

certik:
EXP I think is represented as just Power with a base E.

PR #1991, opened 2023-11-20
Minor fix for SBML printing of piecewise function

lkeegan:
- If the last piece condition is `true` then don't explicitly print this condition
- This doesn't change symengine behaviour (other than sbml printing)
  - Both forms are equivalent for the symengine sbml parser
- But it improves compatibility of the generated sbml string with the libSBML library
  - libSBML parser returns `NaN` if multiple explicit piece conditions evaluate to `true`


isuruf (approved):

lkeegan:
@isuruf thanks for reviewing

PR #1995, opened 2023-11-29
Use `unique_ptr` for `llvm::LLVMContext` and `llvm::ExecutionEngine`

### LINKED ISSUE #1994: LLVMVisitor assignment constructor after init segfaults
This code segfaults:

```
LLVMDoubleVisitor v1;
v1.init({x, y}, *r);
LLVMDoubleVisitor v2;
v1 = v2;
```

Test case to reproduce:

- test: https://github.com/symengine/symengine/compare/master...lkeegan:symengine:reproduce_llvm_segfault
- stacktrace: https://github.com/lkeegan/symengine/actions/runs/7020980960/job/19101975785

From the stacktrace I think this is a bug in LLVM and not in symengine, but maybe the simplest solution would be for symengine to use a unique_ptr instead of a shared_ptr to store  `llvm::LLVMContext` and `llvm::ExecutionEngine`. This would disable the above problematic copy constructor.

lkeegan:
- this implicity disables the `LLVMVisitor` copy constructors
  - avoids segfault that occured if an initialized `LLVMVisitor` is copy assigned to
  - (segfault itself looks like it may be a bug in LLVM)
- define default constructor/destructor in cpp file to avoid "incomplete type" compilation error
- resolves #1994

### DIFF
--- a/symengine/llvm_double.cpp
+++ b/symengine/llvm_double.cpp
@@ -51,6 +51,9 @@ class IRBuilder : public llvm::IRBuilder<>
 {
 };
 
+LLVMVisitor::LLVMVisitor() = default;
+LLVMVisitor::~LLVMVisitor() = default;
+
 llvm::Value *LLVMVisitor::apply(const Basic &b)
 {
     b.accept(*this);
@@ -126,7 +129,7 @@ void LLVMVisitor::init(const vec_basic &inputs, const vec_basic &outputs,
     llvm::InitializeNativeTarget();
     llvm::InitializeNativeTargetAsmPrinter();
     llvm::InitializeNativeTargetAsmParser();
-    context = std::make_shared<llvm::LLVMContext>();
+    context = make_unique<llvm::LLVMContext>();
     symbols = inputs;
 
     // Create some module to put our function into it.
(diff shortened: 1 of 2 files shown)


rikardn (approved):
This feels as the right thing to do since the visitor has unique ownership of the context.

bjodah:
Looks good!

Should we include the test case you wrote as well? something like this:
```c++
TEST_CASE("LLVMDoubleVisitor assignment constructor after init",
          "[llvm_double]")
{
    RCP<const Basic> x, y, r;
    x = symbol("x");
    y = symbol("y");
    r = add(x, pow(x, y));
    LLVMDoubleVisitor v1;
    v1.init({x, y}, *r);
    if constexpr (std::is_copy_constructible<LLVMDoubleVisitor>::value) {
        LLVMDoubleVisitor v2;
        v1 = v2;
    }
}
```
(untested). But maybe it's overkill. I'm +1 to merging as is as well.

lkeegan:
> Looks good!
> 
> Should we include the test case you wrote as well? something like this:
> 
> ```c++
> TEST_CASE("LLVMDoubleVisitor assignment constructor after init",
>           "[llvm_double]")
> {
>     RCP<const Basic> x, y, r;
>     x = symbol("x");
>     y = symbol("y");
>     r = add(x, pow(x, y));
>     LLVMDoubleVisitor v1;
>     v1.init({x, y}, *r);
>     if constexpr (std::is_copy_constructible<LLVMDoubleVisitor>::value) {
>         LLVMDoubleVisitor v2;
>         v1 = v2;
>     }
> }
> ```
> 
> (untested). But maybe it's overkill. I'm +1 to merging as is as well.

I like the idea, but I think symengine only requires c++11 and this I guess needs c++17?

bjodah:
Ah, yes you're right.

bjodah:
Let's get this in. Merry Christmas.

PR #2017, opened 2024-05-10
XX testing the CI

certik:


anutosh491:
The reason for `Debug-macos-latest-gcc` & `Release-macos-latest-gcc` failing is the following. The mac builds are failing because macos-latest is now an M1 arm64 mac instead of an intel one. A quick short-term fix could be to use macos-13 instead of macos-latest to use an intel mac in the CI.

certik:
Closing in favor of https://github.com/symengine/symengine/pull/2019.

PR #2019, opened 2024-05-19
CI: Use the old Intel based macOS runners

certik:


certik:
It looks like the 4 macOS tests that I changed in this PR are now "required, but not present", and "present, but not required". Then there are 2 Ubuntu tests which are "present, but not required" (due to #2020). Other than that, I think it looks in good shape!

certik:
I updated the settings to require exactly the tests we have.

PR #2023, opened 2024-05-21
Added support for basic series expansion through cwrapper

anutosh491:


anutosh491:
Just one test of the lot fails. Not sure why though .... it just shows that the test was just aborted.
```
54/66 Test #54: test_cwrapper ....................Subprocess aborted***Exception:   0.31 sec
```

Issue #2029, opened 2024-06-01
unicode printer is not complete


jogehl:
If I parse the equation "pow(x, 2) + (x**2 )* 2*x*y + 3*y" and try to use unicode() to get a unicode representation I get a SIGSEGV. 
I use the following code.
```
#include <iostream>
#include <symengine/symbol.h>
#include <symengine/parser.h>
#include <symengine/printers.h>

using namespace SymEngine;
using SymEngine::unicode;

int main() {

    // parse the given string and print the expression
    RCP<const Basic> expr3 = parse("pow(x, 2) + (x**2 )* 2*x*y + 3*y");
    std::cout << "expr3: " << *expr3 << std::endl;


    // print the expression with unicode printer so that it looks nice
    auto s = unicode(*expr3);
    std::cout << s << std::endl;


    return 0;
}

```

rikardn:
Thanks for reporting this. Could the expression be made smaller and still get a segmentation fault?

PR #2031, opened 2024-06-03
Fix unicode printer segfault

lkeegan:
Turns out `x^2 y` is enough to reproduce #2029 which occurred when padding an odd number of characters in `StringBox::add_right`


rikardn (approved):
Thanks for fixing!

PR #2046, opened 2024-09-12
fix 'latest' llvm CI job: add apt packages and use ubuntu 24.04

lkeegan:


MinnieMouse11:
Hi @rikardn, Can you have a look?

rikardn:
LGTM. Why doesn't all tests run?

lkeegan:
> LGTM. Why doesn't all tests run?

I think this is because "required" tests is just a list of test names, but the test name includes all of the build options, so when the CI is updated the names change, but the list of required tests is not updated and now has names that are no longer present in the CI.

I think the solution would be to add a fixed name for each job, then update the list of required jobs (which can only be changed by an admin.)

EinavS1:
@lkeegan it looks like there are still pending checks that do not start

MinnieMouse11:
@lkeegan it looks like there are still pending checks that do not start

oliverlee:
> > LGTM. Why doesn't all tests run?
> 
> I think this is because "required" tests is just a list of test names, but the test name includes all of the build options, so when the CI is updated the names change, but the list of required tests is not updated and now has names that are no longer present in the CI.
> 
> I think the solution would be to add a fixed name for each job, then update the list of required jobs (which can only be changed by an admin.)

You could also use this workflow if you want to define a single stable job name:
https://github.com/re-actors/alls-green

lkeegan:
@oliverlee thanks for the suggestion, looks promising!

lkeegan:
closing in favour of #2049

Issue #1719, opened 2020-12-07
RUBI integrator infrastructure


certik:
@naveensaigit and I brainstormed steps that could be done to get the infrastructure for rule rewriting and decision tree generation working in SymEngine. Our goal is to start with a few rules, but get them working all the way, and then keep adding more rules.

I will be updating the issue description with more details.

certik:
The branch `w` and this file: https://github.com/certik/rubi_generated/blob/79613660c77db95736b1d4eb6fbfb29081230364/run_example.py contains a test for a few simple rules that seem to work.

certik:
@Upabjojr, do you know if the decision tree generated could be modified to output something like:
```python
def generated_rule(b, n, x):
    if (ZeroQ(n+S(1))):      
        return replacement2(b, n, x)
    else:
        return replacement1(b, n, x)
```
instead of the current output, such as in https://github.com/certik/rubi_generated/blob/79613660c77db95736b1d4eb6fbfb29081230364/gen/generated_part000001.py, which seems really complex and probably slower?

Upabjojr:
> @Upabjojr, do you know if the decision tree generated could be modified to output something like:

Do you mean the decision tree generated by MatchPy? The code follows the exact way in which MatchPy operates internally, so it would be quite tough to customize the generated code.

certik:
Yes, the tree generated by the file https://github.com/Upabjojr/rubi_generated/blob/cd35e9e51722b04fb159ada3d5811d62a423e429/decision_tree_rubi_generator.py. I thought you wrote it.

Upabjojr:
It's subclassing _MatchPy_'s `CodeGenerator` object in order to do some customizations of the generated code. 

> ```python
> def generated_rule(b, n, x):
>     if (ZeroQ(n+S(1))):      
>         return replacement2(b, n, x)
>     else:
>         return replacement1(b, n, x)
> ```

Where is this code coming from?

certik:
That's the code that I would like to get generated. If it can be generated, together with the pattern, then it will greatly simplify the decision tree and also probably make it faster.

Upabjojr:
Well, I'm not sure it can be generated like that because it is missing all the logic to match the elements of the expression tree.

Upabjojr:
By the way, I feel that the decision tree is being generated more complicated than what it ought be. Incidentally, I believe it's an issue of replication of objects in the rule definition.

The rules may not be optimal when they are added to MatchPy.

[...]

PR #2068, opened 2024-12-04
Updated Integral

erny123:
Added integral.cpp, integral.h and updated CMakelLists.txt under /symengine/. Added test_integral.cpp and modified CMakeLists.txt under /symengine/tests/. Added directory /Notebooks with a Xeus-Cling Jupyter notebook

This is a basic set up to evlauate integrals. I will further look into adding more integration schemes and adding unevaluated derivatives and integrals.


erny123:
Not sure why the directory I added under `Integrals_and_Derivatives/Notebooks/SymEngineTutorial.ipynb` didn't get added. Can anyone explain that to me?

rikardn:
Was it an empty directory? Empty directories cannot be added in git.

erny123:
> Was it an empty directory? Empty directories cannot be added in git.

It was in there as you can see in this picture: 

image


And here's the git profile: 

image


Which indicates it was pushed correctly?

certik:
Thanks!

I think the basic design is fine, it adds a visitor to evaluate simple integrals. In order to be able to evaluate any integral like sympy, we need to add some of the advanced algorithms there. Or possibly the Rubi algorithm: https://github.com/symengine/symengine/issues/1719.

erny123:
Ya definitely. I am also working on doing unevaluated derivatives and integrals. I actually probably should've started with that. But i'll fork another when I have time.

erny123:
Also, I'm not sure what checks are

bjodah:
Adding an `integrate` method to SymEngine sounds like a great plan! And it does not need to be perfect from day one for it to be useful.

However, I would like the API of such a function not to rely on `try`/`catch` to detect when an integral could not be found. I would use this method opportunistically, and check a very large number of expressions for analytic integrals, and I expect the vast majority of those calls to "fail". Therefor, since exception handling in C++ is potentially much slower, I would rather have the method return a null-(rcp)pointer for when no analytic integral could be found.

erny123:
@bjodah can you link me any  ` null-(rcp)pointer` docs? Sorry, I'm not familiear with this

erny123:
Hi all, added unevaluated integrals along with the printer functions for unevaluated integrals. Anyone care to check?

PR #2077, opened 2025-01-08
Fix boost-cpp version in build script

rikardn:
This hopefully fixes the interpolation issue in the failing CI runs


rikardn:
This fixed the failure.

PR #2104, opened 2025-06-30
Migrate appveyor windows CI to github actions

lkeegan:
- refactor matchpycpp and clang-format into separate jobs
- refactor scripts to simplify logic
- pip install clang-format
- use conda-incubator/setup-miniconda@v3 action to set up conda
- update gmp etc versions
- use windows-latest runners
- use CONDA_PREFIX/Library for install directory on windows & use cygpath to normalise front/back-slashes in git bash on windows
- drop 32-bit windows builds
- make running expand1 part of the test cmake build
- update msys2 jobs to use UCRT64 environment instead of MINGW64
- update FLINT_DEV option to use flint main branch instead of v2.9
- remove appveyor


bjodah (approved):
I think this looks good! I have limited experience with packaging for windows, but as I've understood it ucrt64 is what matters going forward, and mingw is more or less legacy at this point?

PR #2106, opened 2025-07-03
Fix loop conditional in CSRMatrix::csr_has_sorted_indices

bjodah:
Change "a < b - 1" to "a + 1 < b" since 'b' could be zero (and unsigned). Also added some defensive asserts at locations with similar structure.


bjodah:
thank you for the review

Issue #2107, opened 2025-07-07
Failed one testcase in test_integer_class.cpp with boostmp (MSVC 2022)


rujialiu:
First of all, a big thank you to all the maintainers for this awesome library!

I'm using Visual Studio 2022 on Windows 11.  Boostmp is from boost 1.84. Here is the test result:

```
E:\symengine\symengine\tests\basic\test_integer_class.cpp(815): FAILED:
  REQUIRE( mp_get_ui(integer_class("-18446744073709551615")) == 18446744073709551615u )
with expansion:
  4294967295 (0xffffffff)
  ==
  18446744073709551615 (0xffffffffffffffff)

===============================================================================
test cases:   7 |   6 passed | 1 failed
assertions: 455 | 454 passed | 1 failed
```

I'm using boostmp because I need to avoid LGPL dependencies. 

BTW: I'm also thinking about adding another integer class with Fabrice Bellard's libbf which is more lightweight and reasonably fast (and probably the best choice for webassembly), but I'm not sure how hard it is, and whethter it's worth trying.

rujialiu:
I debugged a bit. The reason seems to be the difference in "how many bits in `long`". I changed `mp_get_ui` to use `uint64_t` instead of `unsigned long` like this:

```c++
inline uint64_t mp_get_ui(const integer_class &i)
{
    return mp_abs(i).convert_to<uint64_t>();
}
```

Then that particular test passed, but failed another assert at:

```c++
    REQUIRE(mp_scan1(LONG_MIN) == 63);
```

Because in MSVC 2022 it's defined as (even though I'm building a 64-bit application!):

```c++
#define LONG_MIN    (-2147483647L - 1)
```

I changed it into:

```c++
    REQUIRE(mp_scan1(std::numeric_limits<int64_t>::min()) == 63);
```

And it passed all tests.

So, can I assume all "long" actually means "64-bit integer" and change such tests to use int64/uint64?

rujialiu:
After looking deeper, I saw `unsigned long int mpz_get_ui` in gmp's API, so now I think I probably shouldn't modify `mp_get_ui` and friends, but instead modify the tests that depends on the width of `long`?

lkeegan:
Good point! I agree

PR #2108, opened 2025-07-07
Fixed failing tests when `long` is 32-bit #2107

rujialiu:
PS: The original test contains two copies of

```
REQUIRE(mp_get_ui(integer_class("18446744073709551615"))
        == 18446744073709551615u);
```

I changed the second one to "-18446744073709551615" because it feels like the author's original intention :)


rujialiu:
Fixes #2107

PR #2110, opened 2025-07-12
CI-config for tsan

bjodah:
Currently we have CI jobs for the sanitizers:
- address
- undefined
- memory

This PR adds a config with:
- thread

https://github.com/google/sanitizers/wiki/threadsanitizercppmanual


bjodah:
This is a non-functional change, so I'm going to go ahead and merge this in order to prepare for a PR that aims to fix gh-2111 (we want a tsan run for that PR). I hope that this is ok with everyone.

Issue #2111, opened 2025-07-13
Data race in Dummy creation


bjodah:
This requires a lock:
https://github.com/symengine/symengine/blob/bc858aad303774b12555c1868e21ed04ad67cf98/symengine/symbol.cpp#L43

PR #2112, opened 2025-07-13
Dummy thread safety and serializability

bjodah:
Fixes gh-2111

### DIFF
--- a/symengine/utilities/teuchos/Teuchos_RCPNode.cpp
+++ b/symengine/utilities/teuchos/Teuchos_RCPNode.cpp
@@ -52,6 +52,11 @@
 // Internal implementatation stuff
 //
 
+#if defined(TEUCHOS_DEBUG) && defined(WITH_SYMENGINE_THREAD_SAFE)
+#include <mutex>
+#define USE_MUTEX_TO_PROTECT_NODE_TRACING
+#endif
+
 
 namespace {
 
@@ -122,6 +127,15 @@ rcp_node_list_t*& rcp_node_list()
   return s_rcp_node_list;
 }
 
+#ifdef USE_MUTEX_TO_PROTECT_NODE_TRACING
+std::mutex *& rcp_node_list_mutex()
+{
+  static std::mutex * s_rcp_node_list_mutex = 0;
+  // This construct exists for the same reason as above (rcp_node_list)
+  // We must keep this mutex in place until all static RCP objects have deleted.
+  return s_rcp_node_list_mutex;
+}
+#endif
 
 bool& loc_isTracingActiveRCPNodes()
 {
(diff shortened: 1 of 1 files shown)


symengine/symbol.cpp
```diff
@@ -35,23 +35,49 @@ RCP<const Symbol> Symbol::as_dummy() const
...
+#endif
+                     ))
 {
     SYMENGINE_ASSIGN_TYPEID()
-    count_ += 1;
+#ifdef WITH_SYMENGINE_THREAD_SAFE
+    // this is inefficient: we should not construct Dummy instances using
+    // Dummy() (since we need the thread-safely incremented value *at
+    // construction of Symbol*).
+    dummy_index = std::stoul(get_name().substr(default_Dummy_prefix_len_));
```
bjodah, line 59: I'm not terribly happy with this solution. The crux is that in the constructor:
```c++
Dummy::Dummy() : Symbol(some_name) 
{
    ...
```
we are passing a name to the Symbol class which (rightfully) does not carry and dummy index. So if we want to keep the name of dummy in sync with the index we're in a conundrum, this was the least invasive "fix" I could come up with. If someone knows of another alternative I'm all ears. A builder function works (that's pretty much what `inline RCP<const Dummy> dummy()` is, and there we do the efficient thing...

symengine/symbol.cpp
```diff
@@ -35,23 +35,49 @@ RCP<const Symbol> Symbol::as_dummy() const
...
+#else
     dummy_index = count_;
+    count_ += 1;
+#endif
 }
 
-Dummy::Dummy(const std::string &name) : Symbol("_" + name)
+Dummy::Dummy(const std::string &name) : Symbol(name)
```
bjodah, line 66: this change is needed for a serialization round-trip to be possible.

symengine/utilities/teuchos/Teuchos_RCPNode.cpp
```diff
@@ -52,6 +52,11 @@
 // Internal implementatation stuff
 //
 
+#if defined(TEUCHOS_DEBUG) && defined(WITH_SYMENGINE_THREAD_SAFE)
```
bjodah, line 55: This is backported and modified to use `WITH_SYMENGINE_THREAD_SAFE`, taken from here:
https://github.com/trilinos/Trilinos/blob/293bc60e0456a04fbeaad5505df6a8a08084573f/packages/teuchos/core/src/Teuchos_RCPNode.cpp#L27

The rest of the changes in this file can be found by simply searching for "mutex" throughout the file.

This should hopefully resolve the data race (https://github.com/symengine/symengine/issues/2114#issue-3229206443) without requiring an invasive full update of Teuchos RCP as in #2115

PR #2116, opened 2025-07-15
Dummy thread safety and serializability

### LINKED ISSUE #2111: Data race in Dummy creation

bjodah:
Fixes #2111

Supersedes #2112

### DIFF
--- a/symengine/symbol.cpp
+++ b/symengine/symbol.cpp
@@ -35,23 +35,29 @@ RCP<const Symbol> Symbol::as_dummy() const
     return dummy(name_);
 }
 
+#ifdef WITH_SYMENGINE_THREAD_SAFE
+std::atomic<size_t> Dummy::count_{0};
+#else
 size_t Dummy::count_ = 0;
+#endif
 
-Dummy::Dummy() : Symbol("_Dummy_" + to_string(count_))
+Dummy::Dummy() : Dummy("_Dummy") {}
+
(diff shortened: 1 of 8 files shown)


symengine/symbol.cpp
```diff
@@ -35,23 +35,29 @@ RCP<const Symbol> Symbol::as_dummy() const
...
-    SYMENGINE_ASSIGN_TYPEID()
-    count_ += 1;
-    dummy_index = count_;
 }
 
-hash_t Dummy::__hash__() const
+Dummy::Dummy(const std::string &name, size_t dummy_index)
+    : Symbol(name), dummy_index(dummy_index){SYMENGINE_ASSIGN_TYPEID()}
+
+      hash_t Dummy::__hash__() const
```
bjodah, line 60: clang format insists on indenting this line...

lkeegan (approved):
lgtm! I guess `std::memory_order_relaxed` is enough here since the order of operations doesn't matter at all - we're guaranteed to get a different dummy_index value everytime the Dummy constructor is called which is all we need?

bjodah:
@lkeegan yes, that is my understanding (but I'm by no means an expert in concurrency).

PR #2120, opened 2025-07-29
Add HasBasicVisitor and has_basic(...)

bjodah:
Supersedes gh-2105

### DIFF
--- a/symengine/tests/basic/test_basic.cpp
+++ b/symengine/tests/basic/test_basic.cpp
@@ -963,6 +963,8 @@ TEST_CASE("has_basic: Basic", "[basic]")
     REQUIRE(not has_basic(*r1, *z));
     REQUIRE(not has_basic(*r1, *Nan));
     REQUIRE(has_basic(*r1, *r0));
+    // Matching of subtree's containing associative operators not yet supported:
+    CHECK_THROWS_AS(has_basic(*r1, *r1), NotImplementedError);
 
     r1 = sin(r1);
     REQUIRE(has_basic(*r1, *x));
--- a/symengine/tests/basic/test_basic.cpp
+++ b/symengine/tests/basic/test_basic.cpp
@@ -19,8 +19,11 @@ using SymEngine::EulerGamma;
 using SymEngine::free_symbols;
 using SymEngine::function_symbol;
 using SymEngine::FunctionSymbol;
+using SymEngine::has_basic;
 using SymEngine::has_symbol;
 using SymEngine::I;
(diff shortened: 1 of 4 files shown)


symengine/visitor.h
```diff
@@ -90,6 +90,30 @@ void preorder_traversal_stop(const Basic &b, StopVisitor &v);
...
 void preorder_traversal_local_stop(const Basic &b, LocalStopVisitor &v);
 
+class HasBasicVisitor : public BaseVisitor<HasBasicVisitor, StopVisitor>
+{
+protected:
+    Ptr<const Basic> looking_for_;
+    bool has_;
+
+public:
+    HasBasicVisitor(Ptr<const Basic> x) : looking_for_(x) {}
```
isuruf, line 100: We should raise a NotImplementedError when `x` is an associative op like `Mul, Add, Or` since we don't check for subexpressions.

bjodah: But if we want to check for subexpressions?

symengine/tests/basic/test_basic.cpp
```diff
@@ -945,6 +949,38 @@ TEST_CASE("Complex: Basic", "[basic]")
...
+    x = symbol("x");
+    y = symbol("y");
+    z = symbol("z");
+    r0 = pow(y, integer(2));
+    r1 = add(x, r0);
+    REQUIRE(has_basic(*r1, *x));
+    REQUIRE(has_basic(*r1, *y));
+    REQUIRE(not has_basic(*r1, *z));
+    REQUIRE(not has_basic(*r1, *Nan));
+    REQUIRE(has_basic(*r1, *r0));
```
bjodah, line 965: so this tests matching of sub-trees

symengine/visitor.h
bjodah: alright then, I'll try to filter those then. I'm not quite sure what the best approach would be, `is_a_Atom` is not quite right since e.g. Infty has non-empty `get_args` vector. So I guess I should spell out our type codes. What about `Pow`? It doesn't contain arbitrary number of arguments, so it's then ok to match against it?

isuruf (approved):

symengine/visitor.h
bjodah: Yeah, but bug or not, that's how `Infty` is implemented (it returns `_direction` which gives the sign e.g. +Inf vs. -Inf). I guess a changing that should be a different PR, and I'm not sure what backcomp issues may arise (my gut feeling is that it might be a few...)

Issue #2114, opened 2025-07-14
Vendored Teuchos needs updating


bjodah:
The current version of Teuchos reference counted pointer is susceptible to data races.

<details>
<summary>Here's an except from CI logs in gh-2112</summary>

```
WARNING: ThreadSanitizer: data race (pid=7511)
  Read of size 4 at 0x5599e710dd0c by thread T2:
    #0 Teuchos::RCPNodeTracer::addNewRCPNode(Teuchos::RCPNode*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&) /home/runner/work/symengine/symengine/symengine/utilities/teuchos/Teuchos_RCPNode.cpp:402:34 (test_functions+0x99aa9e) (BuildId: 022bf799b30a1221e854603e4f261e242ae36eb8)
    #1 Teuchos::RCPNodeHandle::RCPNodeHandle<SymEngine::Dummy const>(Teuchos::RCPNode*, SymEngine::Dummy const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, bool, Teuchos::ERCPStrength) /home/runner/work/symengine/symengine/symengine/utilities/teuchos/Teuchos_RCPNode.hpp:745:9 (test_functions+0x37824f) (BuildId: 022bf799b30a1221e854603e4f261e242ae36eb8)
    #2 Teuchos::RCP<SymEngine::Dummy const>::RCP(SymEngine::Dummy const*, bool) /home/runner/work/symengine/symengine/symengine/utilities/teuchos/Teuchos_RCP.hpp:211:15 (test_functions+0x3777f3) (BuildId: 022bf799b30a1221e854603e4f261e242ae36eb8)
    #3 Teuchos::RCP<SymEngine::Dummy const> Teuchos::rcp<SymEngine::Dummy const>(SymEngine::Dummy const*, bool) /home/runner/work/symengine/symengine/symengine/utilities/teuchos/Teuchos_RCP.hpp:574:10 (test_functions+0x377489) (BuildId: 022bf799b30a1221e854603e4f261e242ae36eb8)
[...]

PR #2128, opened 2025-10-13
CI: use symengine_rcp for thread safe tests

bjodah:
The version of Teuchos::RCP we are vendoring is not thread safe in its debug configuration. That has been fixed upstream, but updating Teuchos is non-trivial. xref: gh-2114

To avoid [false positives](https://github.com/symengine/symengine/pull/2127#issuecomment-3389435439) in tests involving multi-threading, I suggest we use our own RCP pointer type for those configurations.


bjodah:
Since there have been no objections, and I feel confident as to why we see the failure with TeuchosRCP, I'm going ahead and merging this PR.

PR #2130, opened 2026-01-21
CI: Bump macos image

jschueller:
Also ubuntu fails with WITH_LLVM=LATEST (22)


jschueller:
> Since we need to leave the `macos-13` runner because of github retiring it this is necessary. I don't understand the coverage going down, but I think we can ignore that. Since macos-14 is now ARM should we also use the `macos-15-intel` runner (if possible)? Should we test on more than one version of macos?

probably, but that could be done in another PR, I just want the CI to go green

Issue #2131, opened 2026-01-26
Fix unicode() and latex() and the function expansion of Derivative


Xingye-Dujing:
### Description:
In the function `UnicodePrinter::bvisit(const Add &x)`, **when `x.get_coef()` is zero and the first term’s multiplier is -1**, the expression is printed as the negative of the original via unicode(*expr). For example, the expression `-sin(x)+cos(x)` incorrectly outputs as `sin(x)-cos(x)`.
### Proposed Fix:
Just take this situation into consideration.
Modify the relevant section of code from:
```
if (not first) {
    if (minus) {
        StringBox op(" - ");
        box.add_right(op);
        box.add_right(t);
        minus = false;
    } else {
        StringBox op(" + ");
        box.add_right(op);
        box.add_right(t);
    }
} else {
    box.add_right(t);
    first = false;
}
```
To:
```
if (not first) {
	if (minus) {
		StringBox op(" - ");
		box.add_right(op);
		box.add_right(t);
		minus = false;
	}
	else {
		StringBox op(" + ");
		box.add_right(op);
		box.add_right(t);
	}
}
else {
	if (minus) {
		StringBox op("- ");
		box.add_right(op);
		box.add_right(t);
		minus = false;
	}
	else box.add_right(t);

	first = false;
}
```
Of course, it can also be changed to
```
if (minus) {
    StringBox op(" - ");
    box.add_right(op);
    box.add_right(t);
    first=false;
    minus = false;
}
else if (not first){
    StringBox op(" + ");
    box.add_right(op);
    box.add_right(t);
}
else {... };
```
**But this would result in first being set to false every time minus becomes true. That is to say, they will do unnecessary work.** So I recommend the first way of writing, although it is rather complicated.

PR #2132, opened 2026-01-27
Fix latex() and unicode(), and seek the advice of Derivative

Xingye-Dujing:
#2131 
In the function UnicodePrinter::bvisit(const Add &x), when x.get_coef() is zero and the first term’s multiplier is -1, the expression is printed as the negative of the original via unicode(*expr).


Xingye-Dujing:
# Description

Sometimes, the `unicode()` function mistakenly add two '-'.
```
unicode:2 - -3.y
str: 2 - 3*y
```

# Solution

**Method 1**: 

Use `abs` to handle the coef (add the '-' separately at the end).

```
else {
    t = parenthesizeLT(abs(p.second), PrecedenceEnum::Mul);                   # abs(p.second)
    auto op = print_mul();
    t.add_right(op);
    auto rhs = parenthesizeLT(p.first, PrecedenceEnum::Mul);
    t.add_right(rhs);
}
```

**Method 2**: 

Just add '+' as an extra, except for coef is -1 where no '-' is added. **Use the negative coef ' - ' as the minus sign**. The drawback of doing this is that there will be **no spaces on either side of the '-'.**

See the submitted code

Xingye-Dujing:
### Description

When the numerator is 1, the num will mistakenly add an extra multiplication sign.

0f9d353aa9a7d3c112a5532c51716211

### Solution

Just delete `box1.add_right(mulbox);` (See the submitted code)

symengine/printers/latex.cpp
```diff
@@ -193,7 +193,13 @@ void LatexPrinter::bvisit(const Derivative &x)
             s << "\\partial " << apply(*prev) << "^" << count << " ";
         }
     }
-    s << "} " << apply(x.get_arg());
+
+    const RCP<const Basic> arg = x.get_arg();
+    std::string arg_str = apply(arg);
+    if (is_a<Add>(*arg) || is_a<Mul>(*arg)) {
+        arg_str = "\\left(" + arg_str + "\\right)";
+    }
```
isuruf:
```suggestion
    arg_str = parenthesizeLT(arg_str, PrecedenceEnum::Pow);
```

isuruf:
Fixed now

PR #2137, opened 2026-02-12
llvm: Fix build with 22.x

jschueller:
Also bumps llvm>=5 requirement to cleanup code (CI still tests 5.x on macos)

/cc @bjodah


bjodah:
Great! Thank you.

PR #2138, opened 2026-03-15
Add cudacode printer

lkeegan:
- with float and double versions
  - `cudacode` is like `ccode` but for cuda using double precision
  - `cudacode_float` is the same using float precision
- add optional cuda runtime test
  - requires cuda and an nvidia gpu to run
  - compiles and runs the generated cuda code
  - checks it numerically matches lambda_double within the expected precision
- add some missing logical operators to code printing
  - so that code printing supports everything that llvm_double supports
- fix bug in logical xor implementation in lambda_double
  - add tests to compare with llvm_double


rikardn (approved):
Interesting and LGTM!

bjodah (approved):
Indeed!

certik:
Nice! Are you running symbolic computation on a GPU, or is this numerical computation that only started as symbolic?

lkeegan:
> Nice! Are you running symbolic computation on a GPU, or is this numerical computation that only started as symbolic?

The latter - this PR just generates cuda code (as a string) from the symbolic expression to numerically evaluate it, the same as the existing c code printer does.

isuruf (commented):
Some questions
1. Should we add the data type (float, double) as an input instead of different printers?
2. Should we add the data type to CCode, etc?
3. Should the mathematical functions add `f` at the end? i.e. `sinf(2.0)` instead of `sin(2.0)`?

lkeegan:
> Some questions
> 
> 1. Should we add the data type (float, double) as an input instead of different printers?
> 2. Should we add the data type to CCode, etc?
> 3. Should the mathematical functions add `f` at the end? i.e. `sinf(2.0)` instead of `sin(2.0)`?

All sound good to me, thanks! Done in e62d3aa72ac27b66a0106a1c43c200ff4093a630

symengine/printers.h
```diff
@@ -19,7 +21,10 @@ std::string latex(const Basic &x);
 std::string latex(const DenseMatrix &x, const unsigned max_rows = 20,
                   const unsigned max_cols = 12);
 
-std::string ccode(const Basic &x);
+std::string ccode(const Basic &x, CodePrinterPrecision precision
+                                  = CodePrinterPrecision::Double);
```
isuruf, line 27: This changes the ABI. Let's have a `ccode` and a `ccode_precision` maybe

bjodah: While we are at it, instead of passing just `precision`, can we introduce some class e.g. CodePrinterSettings, and have the signature take a pointer maybe (either smart or raw), instead? Then precision would be a field on the Settings struct. If we are careful about how we implement its interface, we could even enable adding of new fields in the future without breaking backcomp ABI. Or am I over-complicating things?

isuruf: Yeah, that sounds good.

lkeegan: Done in 0fed64992a7e4ecc8f9ee1ffd2907b73e195f9dd
I assume that the ABI here is `ccode` etc and the C wrapper, but not `CodePrinter` etc?
If the `CodePrinter` classes are also part of the ABI that we don't want to change then it would get a bit more complicated.

bjodah: Looks nice! Your solution is new to me, but as long as we monotonically add fields to the struct this should be backwards compatible I think?

isuruf: From ABI I meant the ABI of the C Wrapper. Everything else (i.e. C++ interface) can change without an issue. I don't like passing around pointers in the C++ layer. Let me get back to you shortly.

isuruf:
@lkeegan I pushed a commit to use the same pattern as the rest of cwrapper. What do you think?

lkeegan:
> @lkeegan I pushed a commit to use the same pattern as the rest of cwrapper. What do you think?

@isuruf thanks - lgtm!

PR #2139, opened 2026-04-15
Add metalcode printer

lkeegan:
- equivalent of cudacoda printer using metal for mac gpus
- key differences versus cudacode:
  - supported precision in metal is float or half
    - default precision for metalcode is float
    - unlike cuda metal does not support double precision
  - some intrinsics are not supported by metal
    - erf/erfc
    - tgamma/lgamma
- add optional metal runtime tests
   - requires a mac and metal-cpp headers to compile
   - compiles and runs the generated metal code on the gpu
   - checks it numerically matches lambda_double within the expected precision
   - add this to the mac CI jobs
- bump github actions versions in CI

### DIFF
--- a/symengine/printers/codegen.cpp
+++ b/symengine/printers/codegen.cpp
@@ -8,9 +8,18 @@ namespace SymEngine
 namespace
 {
 
-std::string print_float_literal(double d)
+const char *print_precision_suffix(CodePrinterPrecision precision)
 {
-    return print_double(d) + "f";
+    switch (precision) {
+        case CodePrinterPrecision::Double:
+            return "";
+        case CodePrinterPrecision::Float:
+            return "f";
+        case CodePrinterPrecision::Half:
+            return "h";
+        default:
+            throw SymEngineException("Unknown code printer precision");
+    }
 }
 
 CodePrinterPrecision
@@ -19,6 +28,7 @@ normalize_code_printer_precision(CodePrinterPrecision precision)
     switch (precision) {
         case CodePrinterPrecision::Double:
         case CodePrinterPrecision::Float:
+        case CodePrinterPrecision::Half:
             return precision;
         default:
             throw SymEngineException("Unknown code printer precision");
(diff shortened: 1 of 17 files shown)


certik:
I'll let others review the details, but metal printer is good to have, I recently added Metal offloading of do concurrent into LFortran and it works super well.

symengine/printers/codegen.cpp
```diff
@@ -495,6 +506,135 @@ void CudaCodePrinter::bvisit(const Infty &x)
...
+    const auto &args = x.get_args();
+    switch (args.size()) {
+        case 0:
+        case 1:
+            throw SymEngineException("Impossible");
+        case 2:
+            s << "fmin(" << apply(args[0]) << ", " << apply(args[1]) << ")";
+            break;
+        default: {
+            vec_basic inner_args(args.begin() + 1, args.end());
```
bjodah: Not sure if this is worth doing, but formally stack depth will grow linearly with number of arguments for this form, whereas a midpoint split would give us log(n) scaling of stack depth. I do realize that this is quite silly since any use involving a large number of values would naturally not use codegen with explicit names, but rather a loop to perform the reduction. I'll let you decide if this is worth spending time on or not.

bjodah (approved):
Looks good!

symengine/printers/codegen.cpp
lkeegan: good point - changed to doing this iteratively instead of recursively: https://github.com/symengine/symengine/pull/2139/commits/5de45867025480e3333a15272b7e663ad04b8756

bjodah: Ah, nice, that looks good, that was actually a point I didn't even consider, I was thinking about the code generated:
`f(f(a,b), f(c,d))` vs. `f(a, f(b, f(c, d)))`.

lkeegan: Oh I see! How about this: https://github.com/symengine/symengine/pull/2139/commits/4c247b052a73da6d1fdc7bd61f00cf688a8f95c9

bjodah: Beautiful! (and recursion is fine since depth scales as log(N))

lkeegan: @bjodah thanks for reviewing!

PR #2140, opened 2026-04-18
 Explicit symbol visibility

jschueller:
This generalizes the use of SYMENGINE_EXPORT visibility macro to export all symbols for all windows compilers instead of using CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS in order to allow shared builds with mingw.


jschueller:
@isuruf hello, I got to a more suitable solution where the same dllimport/dllexport mecanism is used for both msvc and mingw compilers, with a few changes about polys and GaloisField for which I'm not sure (due to subtle template export differences between the two compilers)

jschueller:
maybe someone else could review this, maybe @rikardn @lkeegan ?

lkeegan:
@jschueller thanks for the PR - I pushed one small bugfix ([4851205](https://github.com/symengine/symengine/pull/2140/commits/4851205f52a5742773b24c1ad6d07e8f80630fd7)) and extended our tests to reproduce the bug ([05ddd7d](https://github.com/symengine/symengine/pull/2140/commits/05ddd7d0d7b1fd2ac7ff741e2d8a560c637169ee)).
Also restored sparse iteration for UPoly with the dense fallback only used for GaloisField ([83b57bc](https://github.com/symengine/symengine/pull/2140/commits/83b57bc95131eda04a62174114bfe5e38a9f994f))
Otherwise lgtm!

jschueller:
great, thanks!

symengine/expression.h
```diff
@@ -267,6 +269,12 @@ inline Expression pow(const Expression &base, const Expression &exp)
     return pow(base.get_basic(), exp.get_basic());
 }
 
+inline void mp_pow_ui(Expression &res, const Expression &base,
```
isuruf, line 272: `mp_pow_ui` seems like a weird name for `Expression` class. `pow` name seems enough.

symengine/basic-inl.h
```diff
@@ -121,10 +121,18 @@ std::string to_string(const T &value)
...
-} constantInitializer; // static initializer for every translation unit
+};
+
+inline bool _symengine_init_constants()
+{
+    static ConstantInitializer ci;
+    return true;
+}
+
+static bool _symengine_constants_initialized = _symengine_init_constants();
```
isuruf, line 135: Can you explain why this is needed?

jschueller: indeed, thats tricky, as we need constants to be explicitly exported now, this guarantees the initialization order of the constants storage: the _symengine_constants_initialized triggers the instanciation of the ConstantInitializer across all TUs which in turns allocates the constants (see also constants.cpp)

symengine/expression.h
jschueller: this is needed in usymenginepoly, else I get an error:
D:\a\symengine\symengine\symengine\polys\usymenginepoly.h(14,5): error C2665: 'SymEngine::mp_pow_ui': no overloaded function could convert all the argument types [D:\a\symengine\symengine\build\symengine\symengine.vcxproj]

bjodah:
I think it would be great to get this in, you all have more experience dealing with symbol hiding on these platforms than me so I'll leave the technicalities to you. One question though: would it be feasible to add a (or replace an existent) CI job which exercises these changes? (and guards against regressions)

jschueller:
ok, done, there was already a msvc job with shared libs, and I modified one mingw job to build shared libs too

jschueller:
@isuruf your changes to set default visibility to hidden exposed a few more places where SYMENGINE_EXPORT was missing which is fixed now, but it also exposed some issues on the ci with the new default BUILD_SHARED_LIBS=ON

as this is an independent issue, lets discuss this in #2152 (you can inspect the ci failures there)

isuruf:
@jschueller, let's add my changes back in with the default for `BUILD_SHARED_LIBS=OFF`?

jschueller:
done, its enabled by default except when piranha is used because of its template instanciations (like hash_set iterators) creating ODR violations leading to crashing the tests

isuruf:
Thanks for the PR

jschueller:
great news!

PR #2141, opened 2026-04-24
Fix bug in llvm visitor logical_not implementation

lkeegan:
- not operation was applied directly to float input, and output was not assigned to result
- added a failing test for the issue: https://github.com/lkeegan/symengine/actions/runs/24895479444/job/72898970327#step:8:1815
- fix with same pattern as for other logical operations
  - convert input to bool
  - apply not
  - convert to float


bjodah (approved):
Looks good!

lkeegan:
@bjodah thanks for reviewing!

PR #2142, opened 2026-04-27
Add codegen support for inverse trigonometric functions

lkeegan:
- add lowering pass for reciprocal trig/hyperbolic functions and their inverse variants
- use it before all code printers and llvm codegen passes
- previously for an expression like `acoth(x)`:
  - code printers would generate invalid C/cuda/metal/js code, e.g. `acoth(x)`
  - llvm visitors would raise `NotImplemented`
- now:
  - code printers and llvm visitors generate equivalent valid code, e.g. `atanh(1/x)`


symengine/codegen_lowering.h
```diff
@@ -0,0 +1,14 @@
...
+#define SYMENGINE_CODEGEN_LOWERING_H
+
+#include <symengine/functions.h>
+
+namespace SymEngine
+{
+
+RCP<const Basic> lower_codegen_expr(const RCP<const Basic> &x);
+RCP<const Basic> lower_codegen_expr(const Basic &x);
+
```
isuruf: Can we move this to codegen.h file? No need to have this in a separate file and increase compilation time.

lkeegan: sure: 05294bf26a857833c700a9e502367570132eb607

symengine/printers/codegen.cpp
```diff
@@ -35,13 +38,104 @@ normalize_code_printer_precision(CodePrinterPrecision precision)
...
+                div(one, down_cast<const OneArgFunction &>(x).get_arg()));
+        case SYMENGINE_ACOTH:
+            return atanh(
+                div(one, down_cast<const OneArgFunction &>(x).get_arg()));
+        default:
+            return x.rcp_from_this();
+    }
+}
+
+class LowerCodegenExpr : public BaseVisitor<LowerCodegenExpr, TransformVisitor>
```
isuruf: Can we do something like 
```suggestion
class CodegenBaseVisitor : public BaseVisitor<CodegenBaseVisitor, StrPrinter>
```

and then have all codegen visitors inherit from this class?

lkeegan: If only the codegen visitors needed this lowering then yes, but LLVMVisitor also uses it, and I don't see how it could inherit from this class?

bjodah: Yes, I agree with @lkeegan , since he needs RCP<Basic> (to pass into cse routine) in the LLVM visitor, he can't make use of a StrPrinter visitor here.

bjodah (approved):
LGTM!

isuruf:
@lkeegan @bjodah what do you think of my last commit?

lkeegan:
> @lkeegan @bjodah what do you think of my last commit?

@isuruf thanks this looks good to me! I pushed a couple of minor fixes

symengine/visitor.h
```diff
@@ -360,6 +360,63 @@ class CountOpsVisitor : public BaseVisitor<CountOpsVisitor>
...
 };
 
+template <class Derived, class Base = Visitor>
+class RewriteTrigVisitor : public BaseVisitor<Derived, Base>
+{
+public:
+    using BaseVisitor<Derived, Base>::BaseVisitor;
+    using BaseVisitor<Derived, Base>::visit;
+
+    void visit(const Cot &x)
```
bjodah: Nitpick/taste question: the `visit` methods are virtual right? How do we feel about appending `override` in the header for readability (I personally find it helpful, but I guess it also adds to noise)?

symengine/llvm_double.h
```diff
@@ -24,7 +24,7 @@ namespace SymEngine
 
 class IRBuilder;
 
-class LLVMVisitor : public BaseVisitor<LLVMVisitor>
+class LLVMVisitor : public RewriteTrigVisitor<LLVMVisitor>
```
bjodah, line 27: I think this makes sense as I foresee most people wanting to rewrite those trig functions. But it got me curious, what's our "user story" when it comes to composability of "rewriting" in our printers? Further subclassing I suppose?

bjodah (approved):
Looks good to me. I added two comments which are more out of interest than anything else.

symengine/visitor.h
lkeegan: I agree: 7aaae5c

isuruf (approved):

lkeegan:
thanks for the reviews!

PR #2144, opened 2026-05-28
Fix serialization of Rational/Complex

lkeegan:
- keep a reference to all objects being serialized during serialization
  - some methods return temporary objects, e.g. `Rational::get_num`
  - if a reference to these objects is not kept, their address can be re-used by another temporary object
  - serialization checks for objects already serialized based on raw address
  - so if two different temporary objects have the same address, the second one will be serialized as the first one
  - e.g. `(1/6)*x**3 + (1/600)*x**15` -> `(1/600)*x**3 + (1/600)*x**15`
- add test cases for rational & complex expressions which (often) fail without this change
  - this reproduces https://github.com/symengine/symengine.py/issues/514
  - they only fail if an address is reused, but this seems to happen most of the time for these expressions (at least in CI)

### DIFF
--- a/symengine/serialize-cereal.h
+++ b/symengine/serialize-cereal.h
@@ -43,6 +43,12 @@ public:
         if (not first_seen) {
             return;
         }
+        // Keep a reference to every serialized object for the lifetime of the
+        // archive, as some are temporary objects (e.g. Rational::get_num).
+        // Without this a temp object can be freed and its address
+        // then re-used by another temp object, which is then incorrectly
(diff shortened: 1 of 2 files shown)


rikardn:
Good catch! And LGTM

rikardn (approved):

bjodah (approved):
Yes, really nice catch!

lkeegan:
@bjodah @rikardn thanks for reviewing!

PR #2145, opened 2026-06-01
Fix (harmless) ODR warnings in symengine.r

lkeegan:
- remove #ifdef type switch in cwrapper: basic_struct is now always CRCPBasic_C
  - previously it was CRCPBasic in the cwrapper and CRCPBasic_C everywhere else
- add basic_rcp helper to cast CRCPBasic_C to CRCPBasic
  - without the #ifdef type switch we now need to explicitly do this cast in cwrapper
  - `s->m` -> `basic_rcp(s)`

### DIFF
--- a/symengine/cwrapper.cpp
+++ b/symengine/cwrapper.cpp
@@ -147,6 +147,19 @@ static_assert(std::alignment_of<CRCPBasic>::value
                   == std::alignment_of<CRCPBasic_C>::value,
               "Alignment of 'basic' is not correct");
 
+extern "C++" {
+static inline SymEngine::RCP<const SymEngine::Basic> &basic_rcp(basic_struct *s)
+{
+    return reinterpret_cast<CRCPBasic *>(s)->m;
+}
+
+static inline const SymEngine::RCP<const SymEngine::Basic> &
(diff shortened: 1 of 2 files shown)


lkeegan:
This would resolve https://github.com/symengine/symengine.R/issues/131 but it means replacing `s->m` with `basic_rcp(s)` everywhere in the cwrapper, @isuruf what do you think?

symengine/cwrapper.cpp
```diff
@@ -174,72 +187,72 @@ const char *symengine_version()
 
 void basic_const_set(basic s, const char *c)
 {
-    s->m = SymEngine::constant(std::string(c));
+    basic_rcp(s) = SymEngine::constant(std::string(c));
```
isuruf, line 190: Overall the PR looks good. One minor nit. Should we do something like


```suggestion
    basic_set_rcp(s, SymEngine::constant(std::string(c)));
```

isuruf (approved):

symengine/cwrapper.cpp
isuruf: nvm, you'll need a `basic_get_rcp` too

PR #2147, opened 2026-06-03
Add Poly serialization support

lkeegan:
- extend serialization tests
- fix signature of load_basic for URatPoly
- add load/read_basic for other Poly types
- add /bigobj flag to msvc builds in cmake
  - this PR adds more template instantiations in dense_matrix.cpp
  - for debug builds the intermediate object file can now be larger than the default limit
  - this flag has no effect on the resulting compiled binaries/libraries


bjodah (approved):
This looks good to me!

PR #2150, opened 2026-06-16
Add parser support for truncate

mcaccin:
Add native parsing of the `truncate(...)` syntax to all parsers.

`parse(SymEngine::truncate(x)->__str__())` now returns `SymEngine::Truncate` as one would expect from round-tripping.
Previously, parsing `"truncate(x)"` was resulting in an anonymous `FunctionSymbol("truncate", arg)`

### DIFF
--- a/symengine/parser/parser.cpp
+++ b/symengine/parser/parser.cpp
@@ -102,6 +102,7 @@ init_parser_single_arg_functions()
             {"dirichlet_eta", dirichlet_eta},
             {"floor", floor},
             {"ceiling", ceiling},
+            {"truncate", truncate},
             {"ln", (single_arg_func)log},
             {"log", (single_arg_func)log},
             {"zeta", (single_arg_func)zeta},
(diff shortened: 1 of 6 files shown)


YesserLab (approved):

bjodah:
I think this would need some tests exercising observable effects via e.g. the Lambda visitors, otherwise users are better served by leveraging `FunctionWrapper` (which should allow parsing and printing of any domain specific function).

mcaccin:
> I think this would need some tests exercising observable effects via e.g. the Lambda visitors, otherwise users are better served by leveraging `FunctionWrapper` (which should allow parsing and printing of any domain specific function).

Neither floor nor ceiling have such test coverage either, would you like me to add them for all three? It would be kind of awkward otherwise.

bjodah:
Sure, if you don't mind, that would be great!

isuruf:
Thanks

mcaccin:
Thank you for the patience, $day_job has been a bit insane lately and you must've thought I dropped the ball... apologies about that

PR #2154, opened 2026-08-21
CMake: disable hidden visibility in static builds

mcaccin:
WITH_VISIBILITY_HIDDEN applies to static libraries despite being documented for shared libraries. This hides SymEngine template instantiations required by downstream serialization-hook overloads. As a result, symengine.py cannot apply PySymbol's save_basic/load_basic overrides, and pickle round-trips discard Python Symbol subclasses.

Enable the option only when BUILD_SHARED_LIBS is enabled. Shared builds retain hidden visibility.


isuruf:
Thanks

PR #2155, opened 2026-08-26
ComplexMPC: Fix stub export

jschueller:
When MPC is not available, the ComplexMPC stub class incorrectly had SYMENGINE_EXPORT (__declspec(dllexport) on MSVC). This forced the linker to expect all declared member functions to be exported from the DLL. The accept() methods declared via IMPLEMENT_TYPEID were never defined (type_codes.inc guards them with HAVE_SYMENGINE_MPC), causing unresolved external symbol errors on MSVC with BUILD_SHARED_LIBS.


isuruf:
Thanks

PR #2156, opened 2026-08-26
Fix Unicode printer dropping leading minus on first term

### LINKED ISSUE #2131: Fix unicode() and latex() and the function expansion of Derivative

jschueller:
When the first term of an Add expression had a negative coefficient, the minus sign was silently dropped.
For example, -sin(x)+cos(x) printed as sin(x)+cos(x).

Closes #2131

### DIFF
--- a/symengine/printers/unicode.cpp
+++ b/symengine/printers/unicode.cpp
@@ -437,7 +437,14 @@ void UnicodePrinter::bvisit(const Add &x)
                 box.add_right(t);
             }
         } else {
-            box.add_right(t);
+            if (minus) {
+                StringBox op("- ");
+                box.add_right(op);
+                box.add_right(t);
+                minus = false;
+            } else {
+                box.add_right(t);
+            }
             first = false;
         }
     }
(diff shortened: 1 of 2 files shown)


rikardn (approved):
Good catch and LGTM!

PR #2159, opened 2026-08-26
Fix Eq() for division with negated denominators

### LINKED ISSUE #2044: Equality checking of division
Hi.

I encountered an error when I tried to compare if two symbolic values are equal. They are expressed as $\frac{-x}{1-(y-z)}$ and $\frac{x}{(y-z)-1}$, which should be equal. But in the following code, the Boolean variable `is_eq` is `false`.

```
auto x = SymEngine::symbol("x");
auto y = SymEngine::symbol("y");
auto z = SymEngine::symbol("z");
auto is_eq = SymEngine::eq(
    *SymEngine::div(
        SymEngine::mul(SymEngine::minus_one, x),
        SymEngine::sub(SymEngine::one, SymEngine::sub(y, z))
    ),
    *SymEngine::div(x, SymEngine::sub(SymEngine::sub(y, z), SymEngine::one))
);
```

Any suggestion? Is it a bug? Thank you.

jschueller:
Absorb -1 coefficient into Add denominator with odd exponent in Mul::from_dict so that (-x)/(1-(y-z)) and x/((y-z)-1) canonicalize to the same form.

Closes #2044

### DIFF
--- a/symengine/logic.cpp
+++ b/symengine/logic.cpp
@@ -1,7 +1,42 @@
 #include <symengine/logic.h>
+#include <symengine/mul.h>
+#include <symengine/add.h>
 
 namespace SymEngine
 {
+
+// Helper: if `expr` is a Mul with coef=-1 and an Add factor with an odd
+// integer exponent, absorb -1 into that Add (negating it) and return the
+// equivalent expression with coef=+1.  This lets Eq() recognise that
+// e.g.  -x/(1-(y-z))  ==  x/((y-z)-1).
+static RCP<const Basic> normalize_mul_sign(const RCP<const Basic> &expr)
+{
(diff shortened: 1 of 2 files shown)


isuruf:
`Eq` is structural inequality. We don't usually make these simplification like sympy does to avoid performance penalties. There was a discussion in sympy about automatically distributing `const * Add` that described the pitfalls.

jschueller:
should we close #2044 then ?

rikardn:
I agree with @isuruf's conclusion not to add this as an automatic simplification. But it would be a shame to let this code go to waste. I think it would fit well in the `refine` function.

jschueller:
see #2169

PR #2160, opened 2026-08-26
Fix is_nth_residue returning false positives for negative values

jschueller:
Normalize `a` modulo `mod` using `mp_fdiv_r` before passing it to `_is_nthroot_mod_prime_power` in `is_nth_residue`. This ensures negative values are properly reduced to their canonical residue class, so equivalent residues produce the same result.

Closes #2136


rikardn (approved):
Thanks! I think this should also make it faster.

PR #2161, opened 2026-08-26
sets: fix Intersection::contains and Union::contains

jschueller:
Intersection::contains had two bugs:
- Returned true if ANY component set contained the element (union semantics) instead of requiring ALL component sets (intersection semantics)
- Threw NotImplementedError when a sub-set returned a Contains expression

Union::contains also threw NotImplementedError in the same case.

Both now return a symbolic Contains expression when containment cannot be decided, instead of throwing.

### DIFF
--- a/symengine/sets.cpp
+++ b/symengine/sets.cpp
@@ -1265,13 +1265,18 @@ RCP<const Set> Union::set_complement(const RCP<const Set> &o) const
 
 RCP<const Boolean> Union::contains(const RCP<const Basic> &o) const
 {
+    bool has_contains = false;
     for (auto &a : container_) {
         auto contain = a->contains(o);
         if (eq(*contain, *boolTrue)) {
             return boolean(true);
         }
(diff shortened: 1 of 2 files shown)


symengine/sets.cpp
```diff
@@ -1270,8 +1270,12 @@ RCP<const Boolean> Union::contains(const RCP<const Basic> &o) const
...
             return boolean(true);
         }
-        if (is_a<Contains>(*contain))
-            throw NotImplementedError("Not implemented");
+    }
+    for (auto &a : container_) {
+        auto contain = a->contains(o);
+        if (is_a<Contains>(*contain)) {
+            return make_rcp<Contains>(o, rcp_from_this_cast<const Set>());
+        }
```
isuruf: Instead of another pass, we can remember whether there was a `Contains`.

isuruf (approved):

isuruf:
Thanks

PR #2164, opened 2026-08-26
Fix negative number parenthesization in Mul printing

### LINKED ISSUE #2091: Substitution of negative numbers
Hello,
I have a problem with result of substitution. Could you help me how to do it properly ?

I have simple formula "0.95 + 0.05 * alpha_h_u" and I want to substitute alpha_h_u = -0.5. After the substitution I get "0.95 + 0.05 -0.5". Should not be  the negative number at least in brackets ? It has also consequence to conversion to latex when no \cdot is placed and result is same as subtraction of two numbers.

Thank you for your help in advance

jschueller:
When negative numbers appear as factors in a Mul expression's dict, they were printed without parentheses (e.g. `0.05*-0.5` instead of `0.05*(-0.5)`). In LaTeX this produced ambiguous output like `0.05 -0.5` that looked like subtraction.

Closes #2091

### DIFF
--- a/symengine/printers/strprinter.cpp
+++ b/symengine/printers/strprinter.cpp
@@ -596,7 +596,7 @@ void StrPrinter::bvisit(const Mul &x)
             and down_cast<const Number &>(*p.second).is_negative()
             and neq(*(p.first), *E)) {
             if (eq(*(p.second), *minus_one)) {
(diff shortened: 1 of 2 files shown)


isuruf:
https://github.com/symengine/symengine/issues/2091 was closed

PR #2165, opened 2026-08-27
Add log2, log10, log1p, expm1 and cbrt functions

jschueller:
the idea is to catch up operators available in exprtk (and also available in Python math module)


rikardn (commented):
I can see the benefit of having these as separate classes. They are part of the IEEE floating point standard and having them separated in symengine would make it possible to evaluate standard rounding floating point arithmetic. But, I am still not sure that we want them since they add quite a bit of complexity. We could have them as functions that simply creates them from already available classes. What would be your reason for wanting them? For example `log2` is simply a scaled version of `log`.

jschueller:
yes, it was mainly for convenience to avoid tricky string expression replacement in the formulas

isuruf:
> avoid tricky string expression replacement in the formulas

what do you mean by this?

jschueller:
if I want to compute the analytical diff of an expression containing log2 from a string, I have to replace log2(...) in the string to log(...) and add the scale factor, then pass the replaced string to symengine parser

isuruf:
We can make `functionify` a virtual function and then a subclass can override it.

jschueller:
sounds like a plan

jschueller:
see #2167

PR #2166, opened 2026-08-27
series: Replace Poly /= Coeff with Poly = Poly / Coeff

jschueller:
fmpq_poly_wrapper (FLINT polynomial type) does not define operator/=, which causes error C2676 on MSVC. GCC silently decomposes r /= a into r = r / a, but MSVC requires the compound assignment operator to exist.


isuruf:
Thanks

PR #2167, opened 2026-08-27
parser: Make functionify virtual

jschueller:
This allows subclassing Parser to override functionify and handle custom functions (e.g. log2 expanding to log(x)/log(2)) without modifying the core codebase.


isuruf:
Thanks

PR #2168, opened 2026-08-27
CI: Set cancel-in-progress flag

jschueller:
allows to automatically previous running jobs from the same branch


rikardn:
I think these are good ideas. On the master branch we might not want to cancel. Something like this:

```
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
```

PR #2169, opened 2026-08-27
refine: division with negated denominators and int/float comparison

jschueller:


symengine/tests/basic/test_refine.cpp
```diff
@@ -200,4 +214,26 @@ TEST_CASE("Test refine", "[refine]")
...
+    // #1340: numeric equality of int/float expressions via refine
+    // refine of sub(2*x*y, 2.0*x*y) should simplify to 0
+    REQUIRE(eq(*refine(sub(mul(mul(integer(2), x), y),
+                           mul(mul(real_double(2.0), x), y)),
+                       nullptr),
+               *zero));
+    // refine of sub(x*y, 1.0*x*y) should simplify to 0
+    REQUIRE(
+        eq(*refine(sub(mul(x, y), mul(real_double(1.0), mul(x, y))), nullptr),
+           *zero));
```
rikardn: Doesn't this already auto-simplify to 0? At least it does when I test in python:

```python
x, y = symengine.symbols("x y")
symengine.Integer(1)*x*y - symengine.RealDouble(1.0)*x*y
```

symengine/refine.cpp
```diff
@@ -203,6 +205,71 @@ void RefineVisitor::bvisit(const Interval &x)
     result_ = x.rcp_from_this();
 }
 
+void RefineVisitor::bvisit(const Mul &x)
+{
+    // Apply to children first
+    vec_basic newargs;
+    for (const auto &a : x.get_args()) {
+        newargs.push_back(apply(a));
+    }
```
rikardn, line 214: This has the added benefit of refining each part of a mul separately. Should be able to handle |x|*|y| where x>0 and y>0 for example.

jschueller: indeed; I added a testcase

PR #2171, opened 2026-08-28
Try to reintroduce some MSVC jobs into github actions

### LINKED ISSUE #2163: No MSVC builds in CI
With https://github.com/symengine/symengine/pull/2104 there are only mingw builds and no MSVC builds.
MSVC build errors now show up only in symengine.py's MSVC builds.

cc @lkeegan @bjodah

bjodah:
Fixes #2163

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -238,6 +238,7 @@ jobs:
 
           - ARCH: x64
             OS: windows-latest
+            CC: cl
             BUILD_TYPE: Release
             WITH_MPC: yes
             WITH_MPFR: yes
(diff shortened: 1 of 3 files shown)


bjodah:
Looks like the build is working:
https://github.com/symengine/symengine/actions/runs/33190496039/job/98914438635?pr=2171#step:9:531

Not sure what "sparsity pattern" is wanted for the combination of build options / compiler flags. My suggestion here is quite arbitrary.

isuruf:
I'd like one with these conditions https://github.com/symengine/symengine-wheels/blob/conda/recipe/bld.bat#L10-L23

PR #2173, opened 2026-09-12
Allow non-numeric FunctionWrapper in evalf

alexfikl:
This allows `evalf` from symengine to work similarly to sympy in the symbolic domain: the expression just stays as is. It's up to the `FunctionWrapper` to raise an exception if it can't be evaluated.

For some context, this comes from https://github.com/inducer/sumpy/pull/306 and also requires a bit of code in `symengine.py` to work correctly for that.

xref: https://github.com/inducer/sumpy/pull/306
cc: @inducer


symengine/eval.cpp
```diff
@@ -78,7 +79,13 @@ class EvalVisitor : public BaseVisitor<EvalVisitor, TransformVisitor>
     }
     void bvisit(const FunctionWrapper &x)
     {
-        result_ = x.eval(bits)->rcp_from_this();
+        try {
+            result_ = x.eval(bits)->rcp_from_this();
+        } catch (const NotImplementedError &) {
```
alexfikl, line 84: What would be the right exception to raise here? `NotImplementedError` feels wrong because it will never be implemented and something like `SymEngineError` feels too generic.

alexfikl:
The tests failures from https://github.com/symengine/symengine/actions/runs/34707021267?pr=2173 should be fixed now. From what I can tell, the issue was that I was using a newer gcc that defaults to c++20, while the CI was using g++11 that defaults to c++17.

Might want to force a `CMAKE_CXX_STANDARD` or something in the tests too?

